PR #2, opened 2025-09-26
initial kubebuilder init

dmolik:
New PR created with autoupdate and helm plugins enabled

```
go mod init valkey.io/valkey-operator
kubebuilder init --domain valkey.io --plugins go/v4,autoupdate/v1-alpha,helm/v1-alpha  --project-version 3 --owner 'Valkey Contributors' --project-name valkey-operator
```


PROJECT
```diff
@@ -0,0 +1,15 @@
...
+cliVersion: 4.9.0
+domain: valkey.io
+layout:
+- go.kubebuilder.io/v4
+- autoupdate.kubebuilder.io/v1-alpha
+- helm.kubebuilder.io/v1-alpha
+plugins:
+  autoupdate.kubebuilder.io/v1-alpha: {}
+projectName: valkey-operator
+repo: valkey.io/valkey-operator
```
bjosv, line 14: I don't know if the information about the repo is used, but is there a need to run kubebuilder with `--repo github.com/valkey-io/valkey-operator`?

dmolik: It's my understanding that ``Repo`` is primarily used for Go's import path IE:
image
Thus this PR's import path is `valkey.io/valkey-operator` which is really more marketing than anything. In the event someone wanted to import the operator a couple of headers would need to be added to the website to enable redirect. Local testing, release cutting, and images should not be effected.

bjosv: Ah, yes, I now see that kubebuilder replaces the module name..

The path looks cleaner as it is.
Maybe we will need some forwarding in the future to be able to `go get` or `go install` tools or kubectrl addons from the repo, we can bring it up when the time comes.
```
go get valkey.io/valkey-operator 
go: unrecognized import path "valkey.io/valkey-operator": reading https://valkey.io/valkey-operator?go-get=1: 404 Not Found
```

PR #5, opened 2025-10-05
feat(operator): flush out v1alpha1 CRDs

dmolik:
Signed-off-by: Dan Molik <dan@hyperspike.io>


api/v1alpha1/valkeycluster_types.go
```diff
@@ -0,0 +1,121 @@
...
+type ValkeyClusterSpec struct {
+
+	// Override the default Valkey image
+	Image string `json:"image,omitempty"`
+
+	// Override the default Valkey exporter image
+	ExporterImage string `json:"exporterImage,omitempty"`
+
+	// The number of shards groups. Each shard group contains one primary and N replicas.
+	// +kubebuilder:default:=3
```
bjosv: Optionally we could add `// +kubebuilder:validation:Minimum:=1` here,
and `// +kubebuilder:validation:Minimum:=0` for replicas, to avoid any user weirdness

api/v1alpha1/valkeycluster_types.go
```diff
@@ -0,0 +1,121 @@
...
+	Shards int32 `json:"shards,omitempty"`
+
+	// The number of replicas for each shard group.
+	// +kubebuilder:default:=0
+	Replicas int32 `json:"replicas,omitempty"`
+
+	// Override resource requirements for each pod
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
```
bjosv, line 46: This isn't a ptr compared to the Affinity option that also is optional.
Is this to be handled differently then checking it for nil?

api/v1alpha1/valkeycluster_types.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// The number of shards groups. Each shard group contains one primary and N replicas.
+	// +kubebuilder:default:=3
+	Shards int32 `json:"shards,omitempty"`
+
+	// The number of replicas for each shard group.
+	// +kubebuilder:default:=0
+	Replicas int32 `json:"replicas,omitempty"`
+
+	// Override resource requirements for each pod
```
bjosv: Resources field is in the container spec I think, would something like this work?
I guess the exporter container can get a fixed value.
```suggestion
	// Override resource requirements for the Valkey container in each pod
```

api/v1alpha1/valkeycluster_types.go
dmolik: Sounds good

api/v1alpha1/valkeycluster_types.go
dmolik: No I'll clean it up

dmolik: There seems to be an issue with Deepcopy, the linter throws an error if we try to make resource requirements a pointer
https://github.com/valkey-io/valkey-operator/actions/runs/18779683212/job/53582458472?pr=5#step:4:34

bjosv (approved):
Nice!

api/v1alpha1/valkeycluster_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// +kubebuilder:validation:Minimum=0
+	Replicas int32 `json:"replicas,omitempty"`
+
+	// Override resource requirements for the Valkey container in each pod
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Override resource requirements for the Valkey exporter container in each pod
+	// +optional
+	ExporterResources corev1.ResourceRequirements `json:"exporterResources,omitempty"`
```
jdheyburn, line 50: Could we have this (along with `ExporterImage`) nested under `Exporter`? Such that the yaml is templated as:

```yaml
exporter:
  image: foo
  resources:
    # ... resources go here
```

jdheyburn: But I am fine fixing it in a later version

dmolik: I'm cool with this, we're early enough on that I want to get this merged, and I'll fix it in the next PR. No new CRDs needed.

PR #6, opened 2025-10-10
Add initial developer guide documentation

bjosv:
Add a howto for running the operator locally.

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -5,22 +5,24 @@
 The kubebuilder scaffolding gives a build target `make run` which runs the operator process locally, but towards a K8s cluster.
 Since neither Pod IPs are routable, nor Pod FQDNs are resolvable outside the cluster, any attempt by the operator to connect to a Valkey pod will fail.
 
-Here is the procedure to make it work.
+Here is a procedure to make it work, but you might need to adapt depending on your setup.
 
-**Prerequisites** (i.e. you might need to adapt depending on your setup)
-* Linux and a distro using `systemd-resolved` for DNS (like Ubuntu >= 22.04, Fedora >= 36).<br />
-* K8s cluster via `minikube`, with the [Docker driver](https://minikube.sigs.k8s.io/docs/drivers/docker/) (default on Linux).<br />
-* The domain name for your cluster is `cluster.local`.<br />
+### Prerequisites
 
-**Steps**
-1. Start the K8s cluster and install the operator CRD.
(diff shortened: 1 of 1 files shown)


docs/developer-guide.md
```diff
@@ -0,0 +1,76 @@
...
+
+The kubebuilder scaffolding gives a build target `make run` which runs the operator process locally, but towards a K8s cluster.
+Since neither Pod IPs are routable, nor Pod FQDNs are resolvable outside the cluster, any attempt by the operator to connect to a Valkey pod will fail.
+
+Here is the procedure to make it work.
+
+**Prerequisites** (i.e. you might need to adapt depending on your setup)
+* Linux and a distro using `systemd-resolved` for DNS (like Ubuntu >= 22.04, Fedora >= 36).<br />
+* K8s cluster via `minikube`, with the [Docker driver](https://minikube.sigs.k8s.io/docs/drivers/docker/) (default on Linux).<br />
+* The domain name for your cluster is `cluster.local`.<br />
```
zuiderkwast: A level 3 heading is structurally/semanically more correct.

The `<br>` after the bullet isn't necessary. (It's clutter when reading it in markdown source code format, which people do.)

```suggestion
### Prerequisites

You might need to adapt depending on your setup

* Linux and a distro using `systemd-resolved` for DNS (like Ubuntu >= 22.04, Fedora >= 36).
* K8s cluster via `minikube`, with the [Docker driver](https://minikube.sigs.k8s.io/docs/drivers/docker/) (default on Linux).
* The domain name for your cluster is `cluster.local`.
```

docs/developer-guide.md
```diff
@@ -0,0 +1,76 @@
...
+
+**Steps**
+1. Start the K8s cluster and install the operator CRD.
+
+```bash
+minikube start
+make install
+```
```
zuiderkwast: The content under `1.` should all be part of `1.`, right? Indent it the same as "Start" (in `1. Start...`) to

````suggestion
### Steps
1. Start the K8s cluster and install the operator CRD.

   ```bash
   minikube start
   make install
   ```
````

Alternatively, change the numbered list items into headings instead. Then the body doesn't need to be indented.

````suggestion
### Steps
#### 1. Start the K8s cluster and install the operator CRD.

```bash
minikube start
make install
```
````

docs/developer-guide.md
bjosv: Thanks, looks better.

PR #10, opened 2025-10-31
Feature - Build out initial Deployments

dmolik:

PR #24, opened 2025-12-02
Add rudimentary reconciliation using Deployment

bjosv:
This is somewhat a copy and continuation of #10 Build out initial Deployments.

Reconciles a CR to create the simplest Service, ConfigMap and multiple Deployments, where each Deployment provides a single Pod with a Valkey node for the cluster.

- Includes initial types to "scrape" the current cluster state, which is used to determine what actions to perform.
- Includes readiness and liveness checks that allows us to use official Valkey containers.
- Includes a e2e-test that will verify the Valkey instances are created and accessible by valkey-cli from within the K8s. The cluster is initiated, and slots assigned, and we expect a "cluster_state:ok" from nodes.
- The reconciliation can handle when pods are killed; the recreated Pod (with a new Valkey instance) is added to the cluster, and the killed instance is forgotten by the cluster. This have been tested by randomly killing a pod each 10s without any lost data (minimal dataset).

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -67,7 +67,7 @@ var scripts embed.FS
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.4/pkg/reconcile
 func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := logf.FromContext(ctx)
-	log.V(1).Info("Reconcile...")
+	log.V(1).Info("reconcile...")
 
 	cluster := &valkeyiov1alpha1.ValkeyCluster{}
 	if err := r.Get(ctx, req.NamespacedName, cluster); err != nil {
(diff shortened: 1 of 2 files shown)


sandeepk97:
> * The reconciliation can recreate killed Pods/Valkey instances, and have been tested by randomly killing a pod each 10s without any lost data (minimal dataset)

i have 1 question on this point. when we kill pod, doesnt deployment controller automatically recreate pods? how is reconciliation responsible for recreating pods?

internal/controller/valkeycluster_controller.go
```diff
@@ -33,31 +49,302 @@ type ValkeyClusterReconciler struct {
...
+		}
+	}
+	return nil
+}
+
+// Create Valkey instances, one Deployment and Pod each
+func (r *ValkeyClusterReconciler) upsertDeployments(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+	existing := &appsv1.DeploymentList{}
+	if err := r.List(ctx, existing, client.InNamespace(cluster.Namespace)); err != nil {
+		return err
```
sandeepkunusoth, line 212: here are we listing deployments in whole namespace. i think we need to add  filter by this valkey cluster name label?

`client.MatchingLabels(labels))`

internal/controller/valkeycluster_controller.go
```diff
@@ -33,31 +49,302 @@ type ValkeyClusterReconciler struct {
...
+	}
+
+	replicas := int(cluster.Spec.Shards * (1 + cluster.Spec.Replicas))
+
+	// Create missing deployments
+	for i := len(existing.Items); i < replicas; i++ {
+		deployment := createClusterDeployment(cluster)
+		if err := controllerutil.SetControllerReference(cluster, deployment, r.Scheme); err != nil {
+			return err
```
sandeepkunusoth, line 221: do we need replicas for initial PR?

internal/controller/valkeycluster_controller.go
```diff
@@ -33,31 +49,302 @@ type ValkeyClusterReconciler struct {
...
+				continue
+			}
+
+			log.V(1).Info("Add a new replica", "primary address", primary.Address, "primary Id", primary.Id, "replica address", node.Address)
+
+			if err := node.Client.Do(ctx, node.Client.B().ClusterReplicate().NodeId(primary.Id).Build()).Error(); err != nil {
+				log.Error(err, "Command failed: CLUSTER REPLICATE", "nodeId", primary.Id)
+				return err
+			}
+			return nil
```
sandeepkunusoth, line 317: here cant we add all shards replicas sequentially one time in same reconcilation loop.

if we add return nil, it will add 1 replica per reconciliation.

internal/controller/valkeycluster_controller.go
bjosv: Well spotted, yes I will add that.
Maybe we should add a test with 2 CRs in the same namespace in the near future.

internal/controller/valkeycluster_controller.go
bjosv: Not sure, maybe not, but it didn't add much additional code to make sure we have a resilient cluster.

internal/controller/valkeycluster_controller.go
bjosv: Here in `addValkeyNode()` we only handle a single node that is not yet member of the cluster.
We either add it as a primary or as a replica depending on the current state.
My thinking was that we start off by only performing a single action in each reconcile-loop, i.e. we take decision based on the current cluster state and then make a change, and in the next iteration we see the outcome of that action.
I think it's simpler to start with and can be optimized, wdyt? Would it be a problem?

Issue #26, opened 2025-12-18
(feat): Emit conditions via ValkeyClusterStatus


jdheyburn:
# Implement Status Conditions for ValkeyCluster

## Summary

The `ValkeyCluster` CRD defines a `Status.Conditions` field following Kubernetes API conventions, but the controller currently does not set or update any conditions. This makes it difficult for users and tools to observe the cluster's health and state.

## Current State

The API types include condition support:

https://github.com/valkey-io/valkey-operator/blob/main/api/v1alpha1/valkeycluster_types.go#L82-L94

However, the controller's `Reconcile` method never updates the status, leaving users without visibility into:

- Whether the cluster is ready to serve traffic
- If the cluster is progressing toward the desired state
- When the cluster is degraded but still operational
- What's happening during reconciliation

## Proposed Solution

Implement standard Kubernetes conditions following the [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties):

#### 1. **Ready** (or **Available**)

- **Type:** `Ready`
- **True when:** All shards exist with correct number of nodes, all slots assigned, no pending operations
- **False when:** Initial creation, topology changes needed, missing shards/replicas
- **Unknown when:** Unable to determine cluster state

#### 2. **Progressing**

- **Type:** `Progressing`
- **True when:** Adding nodes, creating deployments, forgetting stale nodes, any reconciliation in progress
- **False when:** Cluster matches desired state and is stable

#### 3. **Degraded**

- **Type:** `Degraded`
- **True when:** Primary lost, failing nodes detected, slot assignment issues, but cluster still functional
- **False when:** All components healthy

#### 4. **ClusterFormed**

- **Type:** `ClusterFormed`
- **True when:** Valkey cluster topology is correctly established (all shards and replicas present)
- **False when:** Pending nodes or missing shards/replicas

#### 5. **SlotsAssigned**

- **Type:** `SlotsAssigned`
- **True when:** All 16384 hash slots are assigned to primaries
- **False when:** Unassigned slots remain

### Key Changes Needed

1. **Import status condition helpers:**

```go
import "k8s.io/apimachinery/pkg/api/meta"
```

2. **Update conditions throughout reconciliation:**

```go
// At start of reconcile
setProgressingCondition(cluster, "Reconciling", "Reconciling ValkeyCluster")

// When infrastructure errors occur
[...]

PR #33, opened 2025-12-19
(feat): Status Conditions for ValkeyCluster

sandeepkunusoth:
This PR enhances ValkeyCluster status (#26) by updating status metadata and status.conditions on every reconciliation. It also updates E2E tests to cover: (Success) provisioning a ValkeyCluster and asserting it reaches Ready with the expected condition, and (Negative) simulating shard loss by deleting a shard deployment to force Degraded state and verifying the Degraded/Ready conditions are set as expected.

### DIFF
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -149,7 +149,7 @@ const (
 	ReasonMissingShards     = "MissingShards"
 	ReasonMissingReplicas   = "MissingReplicas"
 	ReasonReconcileComplete = "ReconcileComplete"
-	ReasonTopologyComplete     = "TopologyComplete"
+	ReasonTopologyComplete  = "TopologyComplete"
 	ReasonAllSlotsAssigned  = "AllSlotsAssigned"
 	ReasonSlotsUnassigned   = "SlotsUnassigned"
 	ReasonPrimaryLost       = "PrimaryLost"
@@ -194,4 +194,3 @@ type ValkeyClusterList struct {
 func init() {
 	SchemeBuilder.Register(&ValkeyCluster{}, &ValkeyClusterList{})
 }
-


test/e2e/e2e_test.go
```diff
@@ -81,7 +81,11 @@ var _ = Describe("Manager", Ordered, func() {
 	// and deleting the namespace.
 	AfterAll(func() {
 		By("cleaning up the curl pod for metrics")
-		cmd := exec.Command("kubectl", "delete", "pod", "curl-metrics", "-n", namespace)
+		cmd := exec.Command("kubectl", "delete", "pod", "curl-metrics", "-n", namespace, "--ignore-not-found=true")
+		_, _ = utils.Run(cmd)
+
+		By("deleting the metrics ClusterRoleBinding")
```
bjosv, line 89: Ah interesting. This is something that is missing in kubebuilders scaffolding I guess.

internal/controller/status.go
```diff
@@ -0,0 +1,41 @@
+package controller
```
bjosv, line 16: We should add the copyright banner here

api/v1alpha1/valkeycluster_types.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
bjosv: Even if its a .go file all other kubebuilder-generated files have space here instead tab.
I think we can undo this change.

hack/load-images-kind.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# script to pull and load images into kind cluster in local development environment
```
bjosv: When do we need this script? The e2e tests worked fine for me without doing anything to download the curl image, but maybe I have something that differs to you. The script is fine, I just want to learn how others run tests.

There is the function `LoadImageToKindClusterWithName` that is used to download the valkey-operator image, maybe we could let the test do something similar for these images?

internal/controller/status_test.go
```diff
@@ -0,0 +1,119 @@
+package controller
```
bjosv, line 16: Copyright banner

hack/load-images-kind.sh
sandeepkunusoth: when i run make test-e2e. i think this is needed everytime we recreate new kind cluster

these images are failing with imagepullbackoff always 

i think this load script is needed one time when we create new kind cluster or we need to load everytime onto kind

currently we are loading only valkey operator image in e2e-tests. i can update this to load all images


UPDATE:

i checked with kind team this issue maybe only be happening with me in my local as i am using rancher desktop. will remove this script for now.
https://github.com/rancher-sandbox/rancher-desktop/issues/9192

test/e2e/e2e_test.go
sandeepkunusoth: yes this is needed.  I get the error during creation of ClusterRoleBinding that it already exists. whenever i run make test-e2e repeatedly on same kind cluster. it doesn't automatically delete this after tests.

test/e2e/e2e_test.go
```diff
@@ -417,6 +494,146 @@ var _ = Describe("Manager", Ordered, func() {
...
+						Shards      int    `json:"shards"`
+					} `json:"status"`
+				}
+
+				err = json.Unmarshal([]byte(output), &cr)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(cr.Status.State).To(Equal("Ready"))
+				g.Expect(cr.Status.ReadyShards).To(Equal(3))
+			}
+			Eventually(verifyClusterReady, 3*time.Minute, 5*time.Second).Should(Succeed())
```
bjosv: Is the extra 5s needed here?
```suggestion
			Eventually(verifyClusterReady, 3*time.Minute).Should(Succeed())
```

test/e2e/e2e_test.go
```diff
@@ -417,6 +494,146 @@ var _ = Describe("Manager", Ordered, func() {
...
+
+				// Ready condition should be False
+				for _, c := range cr.Status.Conditions {
+					if c.Type == "Ready" {
+						g.Expect(c.Status).To(Equal("False"), "Ready condition should be False when degraded")
+						break
+					}
+				}
+			}
+			Eventually(verifyDegradedState, 2*time.Minute, 5*time.Second).Should(Succeed())
```
bjosv: Is the extra 5s needed here?
if not; if I understand it correctly the default timeout is 2 minutes (`SetDefaultEventuallyTimeout`)

```suggestion
			Eventually(verifyDegradedState).Should(Succeed())
```

test/e2e/e2e_test.go
```diff
@@ -417,6 +494,146 @@ var _ = Describe("Manager", Ordered, func() {
...
+			cmd = exec.Command("kubectl", "delete", "valkeycluster", degradedClusterName, "--wait=false")
+			_, _ = utils.Run(cmd)
+
+			By("waiting for cluster to be deleted")
+			verifyClusterDeleted := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "valkeycluster", degradedClusterName)
+				_, err := utils.Run(cmd)
+				g.Expect(err).To(HaveOccurred(), "Cluster should be deleted")
+			}
+			Eventually(verifyClusterDeleted, 2*time.Minute).Should(Succeed())
```
bjosv:
```suggestion
			Eventually(verifyClusterDeleted).Should(Succeed())
```

test/e2e/e2e_test.go
```diff
@@ -417,6 +494,146 @@ var _ = Describe("Manager", Ordered, func() {
...
+					fmt.Sprintf("Expected NodeAddFailed reason, but got: %s", cr.Status.Reason))
+
+				// Verify the Degraded condition is present and True
+				degradedFound := false
+				for _, c := range cr.Status.Conditions {
+					if c.Type == "Degraded" && c.Status == "True" {
+						degradedFound = true
+						g.Expect(c.Reason).To(Equal("NodeAddFailed"),
+							fmt.Sprintf("Degraded condition should have NodeAddFailed reason, got: %s", c.Reason))
+						fmt.Fprintf(GinkgoWriter, "✓ Cluster is in Degraded state with reason: %s\n", c.Reason)
```
bjosv: oh, I didn't add any writing to the GinkgoWriter in the other test, whats the criteria for adding this?
Maybe I should have added that as well?

test/e2e/e2e_test.go
```diff
@@ -417,6 +494,146 @@ var _ = Describe("Manager", Ordered, func() {
...
+				// Ready condition should be False
+				for _, c := range cr.Status.Conditions {
+					if c.Type == "Ready" {
+						g.Expect(c.Status).To(Equal("False"), "Ready condition should be False when degraded")
+						break
+					}
+				}
+			}
+			Eventually(verifyDegradedState, 2*time.Minute, 5*time.Second).Should(Succeed())
+
```
bjosv, line 585: The cluster will return to Ready/ClusterHealthy after the previous step where we delete the Deployment, right? The operator at least should try to add the Deployment again..

test/e2e/e2e_test.go
```diff
@@ -417,6 +494,146 @@ var _ = Describe("Manager", Ordered, func() {
...
+			_, err = utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), "Failed to delete deployment")
+
+			By("waiting for the cluster to enter degraded state with NodeAddFailed")
+			verifyDegradedState := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "valkeycluster", degradedClusterName, "-o", "json")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				var cr struct {
```
bjosv: Since this is duplicated in each test we might need to declare it somewhere common.
Or maybe its more common to have it in each test?

docs/status-conditions.md
```diff
@@ -0,0 +1,372 @@
...
+- `MissingShards` – waiting for shards
+- `MissingReplicas` – waiting for replicas
+
+#### `SlotsAssigned`
+Indicates whether all **16384** hash slots are assigned to primaries.
+
+| Status | Meaning |
+|---|---|
+| `True` | All slots are assigned. |
+| `False` | Some slots remain unassigned. |
```
bjosv: Should we add `(or absent)` in these conditions that are not set in the start of the reconcile function? Maybe that is to detailed? I don't know how it's usually documented.

Then I'm not sure if `SlotsUnassigned` is ever be set in this PR, maybe this conditions should be set at start of the reconcile function?

internal/controller/valkeycluster_controller.go
```diff
@@ -113,15 +138,48 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 		}
 	}
 
+	// Cluster is healthy - set all positive conditions
+	setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonClusterHealthy, "Cluster is healthy", metav1.ConditionTrue)
+	setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonReconcileComplete, "No changes needed", metav1.ConditionFalse)
+	meta.RemoveStatusCondition(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionDegraded)
+	setCondition(cluster, valkeyiov1alpha1.ConditionClusterFormed, valkeyiov1alpha1.ReasonClusterFormed, "Cluster is formed", metav1.ConditionTrue)
+
+	// Check if all slots are assigned
```
bjosv: Hum, all slots should have been assigned when coming here.
We have also already set ReasonClusterHealthy in the lines above, and all slots needs to have been assigned when we have a healthy cluster.

We might need to move the failure-check, maybe we can check it and set SlotsUnassigned to right after `getValkeyClusterState()`?

test/e2e/e2e_test.go
sandeepkunusoth: Currently i dont think operator is automatically trying to add deployment back. need to check this. i think this ca be part of seperate PR.

bjosv: We can add that test in another PR (the operator will add a new Deployment)

internal/controller/valkeycluster_controller.go
sandeepkunusoth: moved

test/e2e/e2e_test.go
sandeepkunusoth: added test case for assert ready after deletion of deployment

docs/status-conditions.md
sandeepkunusoth: didnt understand what u meant by absent here. but moved slots conditions logic as per other comment

internal/controller/valkeycluster_controller.go
```diff
@@ -115,6 +107,23 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		}
+	}
+	// If we have all shards but slots aren't assigned yet, wait for slot assignment
+	if !allSlotsAssigned && len(state.Shards) == int(cluster.Spec.Shards) {
+		log.V(1).Info("slots not assigned yet, requeue..")
+		setCondition(cluster, valkeyiov1alpha1.ConditionSlotsAssigned, valkeyiov1alpha1.SlotsUnassigned, "Waiting for slots to be assigned", metav1.ConditionFalse)
+		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonReconciling, "Waiting for slots assignment", metav1.ConditionFalse)
+		setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonReconciling, "Assigning slots", metav1.ConditionTrue)
+		_ = r.updateStatus(ctx, cluster, state)
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
```
bjosv: Isn't there a risk we get stuck in a loop since we return here before performing any actions on the valkey cluster?
In the steps after this point we will forget stale nodes, add nodes and assign slots,
but here we set the condition that we are waiting for slots to be assigned.
We might have a situation where we never go past this point in the reconciliation loop, and therefor never assign any slots.
Do we need to return here?
In the `updateStatus` function we seem to refetch the current status everytime we update the status anyway.

test/e2e/e2e_test.go
```diff
@@ -567,59 +518,77 @@ spec:
...
-						break
-					}
+				readyCond := utils.FindCondition(cr.Status.Conditions, valkeyiov1alpha1.ConditionReady)
+				if readyCond != nil {
+					g.Expect(readyCond.Status).To(Equal(metav1.ConditionFalse), "Ready condition should be False when degraded")
+				}
+			}
+			Eventually(verifyDegradedState).Should(Succeed())
+
+			By("waiting for the operator to recreate the deployment and recover the cluster")
```
bjosv, line 540: Great!

api/v1alpha1/valkeycluster_types.go
```diff
@@ -177,6 +177,7 @@ type ValkeyCluster struct {
 	Spec ValkeyClusterSpec `json:"spec"`
 
 	// status defines the observed state of ValkeyCluster
+	// +kubebuilder:default:={state: "Initializing", readyShards:0}
```
bjosv, line 180: Ah, this was a clever way to get rid of code.
Is the `+optional` still needed when we have a default?

docs/status-conditions.md
bjosv: In the `Degraded` section you added:
`| False (or absent) | Cluster is not degraded. |`
which I guess is in the case where there is no `Degraded` condition.
I was thinking if we should do the same here, but you have moved it instead which works.

bjosv (commented):
Just a last question about the return after setting `slots not assigned` status

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: Status field should be optional. As this will not be present when user creates CR. i tried removing optional. but i get error during creation of CR that CR is invalid. not sure why its not setting default here.

internal/controller/valkeycluster_controller.go
sandeepkunusoth: yes removed this

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,36 +90,81 @@ type ExporterSpec struct {
...
+	ReasonDeploymentError   = "DeploymentError"
+	ReasonPodListError      = "PodListError"
+	ReasonAddingNodes       = "AddingNodes"
+	ReasonNodeAddFailed     = "NodeAddFailed"
+	ReasonMissingShards     = "MissingShards"
+	ReasonMissingReplicas   = "MissingReplicas"
+	ReasonReconcileComplete = "ReconcileComplete"
+	ReasonClusterFormed     = "TopologyComplete"
+	AllSlotsAssigned        = "AllSlotsAssigned"
+	SlotsUnassigned         = "SlotsUnassigned"
```
bjosv: I think we should prefix these two as well
```suggestion
	ReasonAllSlotsAssigned        = "AllSlotsAssigned"
	ReasonSlotsUnassigned         = "SlotsUnassigned"
```

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,36 +90,81 @@ type ExporterSpec struct {
...
+	ReasonServiceError      = "ServiceError"
+	ReasonConfigMapError    = "ConfigMapError"
+	ReasonDeploymentError   = "DeploymentError"
+	ReasonPodListError      = "PodListError"
+	ReasonAddingNodes       = "AddingNodes"
+	ReasonNodeAddFailed     = "NodeAddFailed"
+	ReasonMissingShards     = "MissingShards"
+	ReasonMissingReplicas   = "MissingReplicas"
+	ReasonReconcileComplete = "ReconcileComplete"
+	ReasonClusterFormed     = "TopologyComplete"
```
bjosv: This was changed in the last fixup. I think it's good if the reason type name matches the string output.
```suggestion
	ReasonTopologyComplete     = "TopologyComplete"
```

bjosv (commented):
Found two comments but then I'm happy.

bjosv (approved):
LGTM!

Issue #27, opened 2025-12-18
(feat): EventRecorder implementation


jdheyburn:
## Summary

Add Kubernetes Event recording to the ValkeyCluster controller to provide visibility into operator actions and debugging information. Events will appear in `kubectl describe` and `kubectl get events`.

### 1. Add EventRecorder to Controller

**File:** `internal/controller/valkeycluster_controller.go`

```go
import (
    "k8s.io/client-go/tools/record"
)

type ValkeyClusterReconciler struct {
    client.Client
    Scheme   *runtime.Scheme
    Recorder record.EventRecorder  // ADD THIS
}
```

### 2. Add RBAC Marker

**File:** `internal/controller/valkeycluster_controller.go` (around line 55-61)

```go
// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```

### 3. Wire Up in main.go

**File:** `cmd/main.go` (around line 183-189)

```go
if err := (&controller.ValkeyClusterReconciler{
    Client:   mgr.GetClient(),
    Scheme:   mgr.GetScheme(),
    Recorder: mgr.GetEventRecorderFor("valkeycluster-controller"), // ADD THIS
}).SetupWithManager(mgr); err != nil {
    setupLog.Error(err, "unable to create controller", "controller", "ValkeyCluster")
    os.Exit(1)
}
```

### 4. Emit Events

Add events at key reconciliation points using:

```go
// Normal events
r.Recorder.Event(cluster, corev1.EventTypeNormal, "Reason", "Message")
r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "Reason", "Format %s", arg)

// Warning events
r.Recorder.Event(cluster, corev1.EventTypeWarning, "Reason", "Error message")
```

## Suggested Events to Implement

> jdheyburn: Feel free to add as many or as little from the list

| Location | Reason | Type | Message |
|----------|--------|------|---------|
| **Infrastructure** |
| Line 152-160 (upsertService) | `ServiceCreated` | Normal | "Created headless Service" |
| | `ServiceUpdateFailed` | Warning | "Failed to update Service: {error}" |
| Line 193-201 (upsertConfigMap) | `ConfigMapCreated` | Normal | "Created ConfigMap with configuration" |
| | `ConfigMapCreationFailed` | Warning | "Failed to create ConfigMap: {error}" |
| Line 218-226 (upsertDeployments) | `DeploymentCreated` | Normal | "Created deployment X of Y" |
[...]

PR #37, opened 2025-12-27
(feat): EventRecorder implementation

sandeepkunusoth:
This PR Implements https://github.com/valkey-io/valkey-operator/issues/27 Kubernetes Event recording for the ValkeyCluster controller to improve operator visibility and debugging. Adds an EventRecorder to the reconciler and wires it via the manager, updates RBAC to allow creating/patching Events, emits Normal/Warning events at key reconciliation points (e.g., node add/join failures/success), and documents event semantics in docs/status-conditions.md.

Added e2e tests for some events. facing issues with validation for all events as events are rate limited https://github.com/kubernetes/kubernetes/issues/136061#issue-3787699874

### DIFF
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -256,7 +256,8 @@ func GetValkeyClusterStatus(name string) (*valkeyiov1alpha1.ValkeyCluster, error
 
 // GetEvents fetches and categorizes Kubernetes events for a given resource.
 func GetEvents(resourceName string) (map[string]bool, map[string]bool, error) {
-	cmd := exec.Command("kubectl", "get", "events", "--field-selector", fmt.Sprintf("involvedObject.name=%s", resourceName), "-o", "json")
+	cmd := exec.Command("kubectl", "get", "events", "--field-selector",
+		fmt.Sprintf("involvedObject.name=%s", resourceName), "-o", "json")
 	output, err := Run(cmd)
 	if err != nil {
 		return nil, nil, err


internal/controller/valkeycluster_controller.go
```diff
@@ -203,11 +212,14 @@ func (r *ValkeyClusterReconciler) upsertService(ctx context.Context, cluster *va
 	if err := r.Create(ctx, svc); err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			if err := r.Update(ctx, svc); err != nil {
+				r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "ServiceUpdateFailed", "Failed to update Service: %v", err)
 				return err
 			}
 		} else {
 			return err
```
bjosv, line 219: We have a `ConfigMapCreationFailed`, but we don't have a `ServiceCreationFailed` and emit it here.
Should we have one, or is this event implicit in some other way?

bjosv: Hum, or maybe we have too many events already..
https://book.kubebuilder.io/reference/raising-events#events-should-be-raised-in-certain-circumstances-only

When testing on `minikube` all events are noted, but on `kind` many are lost/rate-limited, and I guess that why the test fails intermittently. The status conditions is a better indicator for a healthy cluster.

I guess we need to investigate what level of detail we really need here.

sandeepkunusoth: yeah, i will add new issue for this. we need to agree on what all events should be removed.

Issue #32, opened 2025-12-18
(feat): Support Affinity Configuration for ValkeyCluster


jdheyburn:
## Summary

Apply `spec.affinity` from ValkeyCluster to pods to control pod placement using affinity and anti-affinity rules.

## Current State

- ✅ `spec.affinity` field exists in API (line 50-53 in `api/v1alpha1/valkeycluster_types.go`)
- ✅ Comment notes "Some basic anti-affinity rules will be applied by default"
- ❌ Not applied to pod spec in deployment

## Implementation

**File:** `internal/controller/deployment.go`

Update the pod spec in `createClusterDeployment()`:

```go
Spec: corev1.PodSpec{
    Containers: containers,
    Affinity:   cluster.Spec.Affinity,  // ADD THIS
    Volumes:    volumes,
}
```

## Acceptance Criteria

- [ ] `spec.affinity` applied to pod spec

## Notes

The API comment mentions "Some basic anti-affinity rules will be applied by default" - this implementation applies user-specified affinity. Default anti-affinity rules will be added in a separate enhancement.

## References

- [Kubernetes Affinity and Anti-Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity)

riccardoforzan:
Hey there,
I've tried to tackle this feature as starter contribution. 

I would like to point out that the devcontainer bundled with the project cannot be used "as is", because the default image for golang:1.24 is now based on debian trixie and moby has been removed. 

This is the error I got while building the devcontainer:
```txt
0.168 (!) The 'moby' option is not supported on Debian 'trixie' because 'moby-cli' and related system packages have been removed from that distribution. 
0.168 (!) To continue, either set the feature option '"mob [2026-01-04T13:20:29.860Z] y": false' or use a different base image (for example: 'debian:bookworm' or 'ubuntu-24.04'). 
0.168 ERROR: Feature "Docker (Docker-in-Docker)" (ghcr.io/devcontainers/features/docker-in-docker) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/docker-in-docker for help troubleshooting this error.
```

A simple fix woud be to pin the base image to `golang:1.24-bookworm`.

To streamline the onboarding procedure maybe you can consider using only one local kuberenets solution: currently kind is used in the makefile and minikube is suggested in the contributing guide.

PR #38, opened 2026-01-04
(feat): Support Affinity Configuration for ValkeyCluster (#32)

riccardoforzan:
Fixes https://github.com/valkey-io/valkey-operator/issues/32

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module valkey.io/valkey-operator
 go 1.24.6
 
 require (
+	github.com/google/go-cmp v0.7.0
 	github.com/onsi/ginkgo/v2 v2.22.1
 	github.com/onsi/gomega v1.36.2
 	github.com/stretchr/testify v1.10.0
@@ -38,7 +39,6 @@ require (
 	github.com/google/btree v1.1.3 // indirect
 	github.com/google/cel-go v0.26.0 // indirect
 	github.com/google/gnostic-models v0.7.0 // indirect
-	github.com/google/go-cmp v0.7.0 // indirect
 	github.com/google/pprof v0.0.0-20241210010833-40e02aabc2ad // indirect
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.26.3 // indirect


internal/controller/deployment.go
```diff
@@ -108,6 +108,7 @@ func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.De
 							},
 						},
 					},
+					Affinity: cluster.Spec.Affinity,
 					Volumes: []corev1.Volume{
```
sandeepkunusoth, line 122: Can u add some e2e tests. Also we may need to update sample https://github.com/valkey-io/valkey-operator/blob/main/config/samples/v1alpha1_valkeycluster.yaml

riccardoforzan: Ok, how should I approach this? By looking at https://github.com/valkey-io/valkey-operator/blob/be52c44841cfc7ef84d3c26b879306b576ad3c7d/test/e2e/e2e_test.go#L337 I would edit the default ValkeyCluster (as you mentioned above) and then add (or rework) the validating condition in `verifyPodStatus`. Also, I was thinking about testing with an affinity condition here to avoid tinkering with the number of nodes required in the E2E test setup

sandeepkunusoth: i think this is good for now as you have already validated deployment spec in unit tests. maybe we can add e2e test for this in seperate PR covering node placement

sandeepkunusoth (commented):
LGTM

internal/controller/deployment.go
jdheyburn: @sandeepkunusoth Do we need a separate issue to track that?

sandeepkunusoth (approved):

internal/controller/deployment.go
sandeepkunusoth: i am not sure how setting this affinity on deployment alone will help here. will this help in scheduling valkey master on node 1 and its respective replica on a different node?  i think we need to add e2e test case later with different nodes on kind. yes it can be added as issue or update this exisiting PR add todo comment maybe.

jdheyburn: @sandeepkunusoth The goal with this issue is to support affinity block from what the user provided. For advanced scheduling such as placing a master and replica on different hosts, that is to come in another issue. I think we needed some time to discuss what those rules should be.

internal/controller/deployment_test.go
```diff
@@ -19,6 +19,8 @@ package controller
 import (
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
```
bjosv, line 22: I believe there is a need to run `go mod tidy` and include that update in this PR.
Also, do we need to require `cmp` or can it replaced by an existing dependency already used in our tests?

riccardoforzan: I'll run `go mod tidy` and include the changes in the PR.
Regarding `cmp`, if there is an existing dependency already used for deep comparisons I would be more than happy to follow your suggestion

bjosv: I believe `assert.Equal` should work for this test, but `cmp` is fine with me.
Seems commonly used in other projects, and it was an indirect dependency already.

PR #42, opened 2026-01-07
(feat): Add support for auto-mounting extra configs

utdrmac:
This PR adds support for auto-discovery of a user-created Secret, or ConfigMap, that is named `$clusterName-conf`, which is then mounted into each pod under `/config/valkey.conf.d/`

The default configmap provided by the operator was modified to include the above directory for additional configs.

This operator design pattern follows a similar one found in Percona's database Operators (MySQL, Mongo, PG) where users can provide additional database configs by creating configMaps with a specific naming convention.

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -50,7 +51,8 @@ const (
 // ValkeyClusterReconciler reconciles a ValkeyCluster object
 type ValkeyClusterReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
 }
 
 //go:embed scripts/*
@@ -64,6 +66,7 @@ var scripts embed.FS
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
 // +kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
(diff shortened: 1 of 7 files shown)


internal/controller/volumes.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// Add volumes to spec
+	cluster.Spec.Volumes = append(
+		cluster.Spec.Volumes,
+		scriptsVolume,
+		defaultConfigVolume,
+		userConfigVolume,
+	)
+
```
sandeepkunusoth: this mutation of cluster.spec.Volumes should be done carefully, as this will in turn invoke cluster reconciliation.
There can be possibility of having duplicate volume definitions in this spec. as this is appending on every reconciliation.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -72,6 +72,10 @@ type ValkeyClusterSpec struct {
 	// Metrics exporter options
 	// +optional
 	Exporter ExporterSpec `json:"exporter,omitempty"`
+
+	// Attached Volumes
+	// +optional
+	Volumes []corev1.Volume `json:"volumes,omitempty"`
```
sandeepkunusoth: Do we want users to pass operator-managed volumes via cluster.Spec.Volumes? Instead, can we expose something like spec.customConfig.configMapName / secretName and have the operator build/mount the volume internally.

I’m not sure how common the “mount an existing ConfigMap” use case is, and since the operator doesn’t manage that ConfigMap’s lifecycle.

test/e2e/e2e_test.go
```diff
@@ -438,14 +451,22 @@ var _ = Describe("Manager", Ordered, func() {
...
 			Eventually(verifyConfigMapRemoved).Should(Succeed())
 
+			By("validating that the user ConfigMap remains")
+			verifyConfigMapRemains := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "configmap", valkeyClusterName + "-conf")
+				_, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+			}
+			Eventually(verifyConfigMapRemains).Should(Succeed())
+
```
sandeepkunusoth, line 583: i think we can add here test to validate that valkey-server is having updated config. verify config via CONFIG GET.  

just to verify if custom config from valkey.conf.d is being used or valkey.conf in case same config is present in both.

api/v1alpha1/valkeycluster_types.go
utdrmac: Thinking back on this, what about data persistence? If someone wants to enable snapshots, or AOF, then they should be able to mount a PV/PVC. I would imagine this will be an eventual feature of the operator.

internal/controller/volumes.go
```diff
@@ -0,0 +1,131 @@
...
+	configMapName := cluster.Name + "-conf"
+	userConfigFilter := types.NamespacedName{
+		Namespace: cluster.Namespace,
+		Name:      configMapName,
+	}
+
+	// First, look for a Secret named "$clusterName-conf"
+	err := c.Get(ctx, userConfigFilter, &corev1.Secret{})
+	if err == nil {
+		log.V(1).Info("found user-created secret config")
```
sandeepkunusoth, line 83: I think we should pass "$clusterName-conf" from the input spec so users can optionally provide any existing ConfigMap if needed. Also, getUserConfigVolume / userConfig should only be used when this new custom field is set in the spec; otherwise, the default behavior should remain not to use any existing resource. 

we need to agree on spec changes with others. Please update spec design or feel free to start discussion. not sure if this feature is priority for others.

internal/controller/valkeycluster_controller.go
```diff
@@ -267,14 +267,15 @@ func (r *ValkeyClusterReconciler) upsertDeployments(ctx context.Context, cluster
 
 	expected := int(cluster.Spec.Shards * (1 + cluster.Spec.Replicas))
 
-	// Attach script volume, and config volumes to cluster
-	if err := attachVolumesToCluster(ctx, r.Client, cluster); err != nil {
+	// Get script volume, default config, and user config volumes
+	configVolumes, err := getConfigVolumes(ctx, r.Client, cluster)
```
sandeepkunusoth, line 287: can we move this getConfigVolumes call to inside createDeployment. this way we dont need to pass it as param?

internal/controller/valkeycluster_controller.go
```diff
@@ -74,7 +75,7 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	log.V(1).Info("reconcile...")
 
 	cluster := &valkeyiov1alpha1.ValkeyCluster{}
-	if err := r.Get(ctx, req.NamespacedName, cluster); err != nil {
+	if err := r.Client.Get(ctx, req.NamespacedName, cluster); err != nil {
 		return ctrl.Result{}, client.IgnoreNotFound(err)
```
sandeepkunusoth, line 82: i think we need to switch back to r.Get( here just to stay consistent with c r.List(, r.Create( and rest of the project. If we want to include this, it should be done for all maybe in other PR.

internal/controller/valkeycluster_controller.go
utdrmac: Sure, I was trying to keep createDeployment() clean, and not have a chain of nested func calls.

Issue #43, opened 2026-01-08
ValkeyCluster Intermittently getting struck in Degraded after deployment deletion E2E Test


sandeepkunusoth:
seeing an intermittent E2E test failure where a ValkeyCluster remains stuck in Degraded state during a E2E test run.

Reconcile emits repeated warning events:

Warning NodeForgetFailed ... Failed to forget node: Can't forget my master!

E2E job: [GitHub](https://github.com/valkey-io/valkey-operator/actions/runs/20806446577/job/59761668293)

What happened

During the failing run, the cluster becomes degraded and does not recover within the test timeout. When i checked events there is this warning.
Warning  NodeForgetFailed  59s (x6 over 67s)  valkeycluster-controller  Failed to forget node: Can't forget my master!

I suspect this is triggered when the E2E deletes the primary/master shard Deployment (or Pod) and the controller then attempts CLUSTER FORGET in a way that causes Valkey to reject it with ERR Can't forget my master!. I tried to delete master shard outside this e2e test but was unable to reproduce it manually.

Expected behavior

After the test action (deleting a primary/master deployment/pod or whatever the E2E does), the operator should converge the cluster back to healthy.

E2E test should should pass always on multiple runs. need to fix controller.

PR #44, opened 2026-01-09
(feat): Extra Valkey config

utdrmac:
This feature request implements the ability for users to supply their own configuration for Valkey as an inline stanza when deploying a new cluster. The user configuration is appended to the required base configuration, which is created as a configMap. Since the operator manages the configMap, any changes to the config are reconciled. Additionally, users cannot override the required cluster settings.


config/samples/v1alpha1_valkeycluster.yaml
```diff
@@ -5,3 +5,6 @@ metadata:
 spec:
   shards: 3
   replicas: 1
+  valkeyConfig: |
+    maxmemory 50mb
+    maxmemory-policy allkeys-lfu
```
sandeepkunusoth: Hi, 

instead of passing whole config file inline. i think it will be good to have customConfig in form of 
map[string]string (i feel this option would be better to validate later if we want to.)
```
spec:
  customConfig:
    maxmemory: "50mb"
    maxmemory-policy: "allkeys-lfu"
```

or []string (like “maxmemory 50mb”)

```
spec:
  customConfig:
    - "maxmemory 50mb"
    - "maxmemory-policy allkeys-lfu"
```
 reference: https://github.com/spotahome/redis-operator/blob/632aa3da88ee2a28ccb9fc4f571ddcb556dbd713/api/redisfailover/v1/types.go#L45

internal/controller/valkeycluster_controller.go
```diff
@@ -235,19 +236,26 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
+	serverConfig := `# Base operator config
+cluster-enabled yes
+protected-mode no
+cluster-node-timeout 2000
+`
+
+	if cluster.Spec.ValkeyConfig != "" {
+		serverConfig += "\n# Extra config\n" + cluster.Spec.ValkeyConfig
+	}
+
```
sandeepkunusoth, line 521: i think  this approach will work for creating Valkey cluster. but for updating config of exisitng cluster. ConfigMap updates wont trigger rollout restart of previous deployments. something we need to consider later maybe.

internal/controller/valkeycluster_controller.go
```diff
@@ -224,8 +224,9 @@ func (r *ValkeyClusterReconciler) upsertService(ctx context.Context, cluster *va
 	return nil
 }
 
-// Create or update a basic valkey.conf
-func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+// Create or update a default valkey.conf
+// If additional config is provided, append to the default map
+func (r *ValkeyClusterReconciler) upsertConfigMaps(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
 	readiness, err := scripts.ReadFile("scripts/readiness-check.sh")
```
sandeepkunusoth: why did we change to upsertConfigMaps? is this creating 2 configmaps here?

config/samples/v1alpha1_valkeycluster.yaml
utdrmac: I was following the pattern established in the [valkey-helm chart](https://github.com/valkey-io/valkey-helm/blob/main/valkey/values.yaml#L160) which permits whole config inline. Similar to [mysql cluster operator](https://github.com/percona/percona-xtradb-cluster-operator/blob/main/pkg/apis/pxc/v1/pxc_types.go#L577). But if you'd rather the map[string]string, I can do that.

internal/controller/valkeycluster_controller.go
utdrmac: right, the current behavior of the operator does not trigger rolling restarts. agree to be implemented in a future PR.

internal/controller/valkeycluster_controller.go
utdrmac: initially it was creating 2, but then i realized i could just append the config to existing. i'll revert.

config/samples/v1alpha1_valkeycluster.yaml
sandeepkunusoth: its just that if we use inline we will need to update deployment using hash if configmap changes deployment wont get updated with new configmap.   https://github.com/percona/percona-xtradb-cluster-operator/blob/97b510d489c33a6715e31be05dac6339b7babf00/pkg/controller/pxc/upgrade.go#L91

api/v1alpha1/valkeycluster_types.go
```diff
@@ -74,7 +74,7 @@ type ValkeyClusterSpec struct {
 	Exporter ExporterSpec `json:"exporter,omitempty"`
 
 	// Additional valkey configuration
-	ValkeyConfig string `json:"valkeyConfig,omitempty"`
+	ValkeyConfig map[string]string `json:"valkeyConfig,omitempty"`
 }
```
sandeepkunusoth, line 116: can we add TODO here updating config of exisitng CR will not be propagated today. we need to fix this later.

utdrmac: done

sandeepkunusoth (approved):
LGTM

sandeepkunusoth (dismissed):
lint errors and e2e errors

api/v1alpha1/valkeycluster_types.go
```diff
@@ -72,6 +72,10 @@ type ValkeyClusterSpec struct {
 	// Metrics exporter options
 	// +optional
 	Exporter ExporterSpec `json:"exporter,omitempty"`
+
+	// Additional valkey configuration
+	// TODO Updating the config of an existing CR currently does not trigger cluster restart
```
jdheyburn: Would you mind creating an issue of what is needed to remove this TODO please?

api/v1alpha1/valkeycluster_types.go
```diff
@@ -72,6 +72,10 @@ type ValkeyClusterSpec struct {
...
+	ValkeyConfig map[string]string `json:"valkeyConfig,omitempty"`
```
jdheyburn: Not pressing for this PR, but would it make sense to simply name this `config`? Or have it nested under `valkey`?

```yaml
kind: ValkeyCluster
spec:
  valkey:
    config:
      # ...
```

I'm thinking about future compatibility for when we extend the operator to other use cases, such as `ValkeySentinel`.

internal/controller/valkeycluster_controller.go
jdheyburn: Could we raise a GitHub discussion (idea) for that please?

internal/controller/valkeycluster_controller.go
```diff
@@ -235,19 +236,29 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
+protected-mode no
+cluster-node-timeout 2000
+`
+
+	if len(cluster.Spec.ValkeyConfig) > 0 {
+		serverConfig += "\n# Extra config\n"
+		for k, v := range cluster.Spec.ValkeyConfig {
+			serverConfig += k + " " + v + "\n"
+		}
+	}
```
jdheyburn: What would happen if the user set a config for `cluster-node-timeout`? This would template out `cluster-node-timeout` twice, and I assume the config defined would supersede this default.

Similarly if a user specified `cluster-enabled no`. I know that won't be a supported model. Do we want to ignore user provided configs that overwrite these defaullts?

internal/controller/valkeycluster_controller.go
```diff
@@ -235,19 +236,29 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
+
 	cm := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      cluster.Name,
+			Name:      cluster.Name + "-config",
```
jdheyburn: Are we able to set `cluster.Name + "-config"` to a variable so that it can be reused in deployment.go?

jdheyburn (commented):
Minor comments, keen to hear opinions, otherwise LGTM

api/v1alpha1/valkeycluster_types.go
utdrmac: https://github.com/valkey-io/valkey-operator/issues/50

internal/controller/valkeycluster_controller.go
utdrmac: For `cluster-enabled`, I would say prevent override because that's pretty much the purpose of the operator. But for `protected-mode` (and others), we should allow the user to override this at their risk (ie: they must provide an ACL for the default user, or add a cluster manager user). I would imagine that most production environments do NOT want to operate in an unprotected-by-default state. This is something we can fix later on with proper user creation/handling.

api/v1alpha1/valkeycluster_types.go
utdrmac: I moved this to Spec.ValkeySpec.Config. For future stuff, ValkeySpec can be other things specific to Valkey. Could also add ValkeySentinelSpec in the future.

internal/controller/valkeycluster_controller.go
utdrmac: Added a check for protected config params. Right now it's only `cluster-enable` but could be easily adjusted for others.

internal/controller/valkeycluster_controller.go
utdrmac: Done

utdrmac:
e2e still fails due to #43

api/v1alpha1/valkeycluster_types.go
```diff
@@ -70,12 +78,22 @@ type ValkeyClusterSpec struct {
...
 	// Metrics exporter options
+	// +kubebuilder:default:={enabled:true}
 	// +optional
 	Exporter ExporterSpec `json:"exporter,omitempty"`
 
-	// Additional valkey configuration
+	// Options, or config which are specific to Valkey server
+	// +optional
+	ValkeySpec ValkeySpec `json:"valkey,omitempty"`
+}
```
sandeepkunusoth: i think we need to change naming here ValkeySpec. why did we additional level here?

sandeepkunusoth:
> e2e still fails due to #43

no this is different issue than 43. its failing because of events limitation. maybe as u have added new emit event other events are not coming.   https://github.com/valkey-io/valkey-operator/actions/runs/20936033780/job/60159472854?pr=44. 
image

check this PR description https://github.com/valkey-io/valkey-operator/pull/37. maybe you can update events check accordingly. currently we are not asserting for all events due to events rate limit.

api/v1alpha1/valkeycluster_types.go
utdrmac: @jdheyburn requested the nested config in a previous comment

bjosv (dismissed):
LGTM

api/v1alpha1/valkeycluster_types.go
jdheyburn: Do you think it makes sense @sandeepkunusoth ? The rationale was that ValkeySpec can be reused across other Valkey depoyments (standalone, sentinel, etc).

sandeepkunusoth: Yes. I was just confused we have valkeyclusterspec and valkeyspec.

sandeepkunusoth: Yes. I was just confused we have valkeyclusterspec and valkeyspec.

utdrmac: @jdheyburn From your earlier comment, I assumed you were thinking along the lines of Spec.ValkeySpec (configs/params for standalone, async, cluster), and a separate Spec.ValkeySentinelSpec (for configs/params, acl specific to sentinel)

internal/controller/valkeycluster_controller.go
```diff
@@ -237,19 +240,36 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
+		for k, v := range cluster.Spec.ValkeySpec.Configuration {
+
+			if slices.Contains(valkeyiov1alpha1.NonUserOverrideConfigParameters, k) {
+				log.Error(nil, "Prohibited valkey server config", "parameter", k)
+				r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "ConfigMapUpdateFailed", "Prohibited config: %v", k)
+				continue
+			}
+
+			serverConfig += k + " " + v + "\n"
+		}
```
sandeepkunusoth, line 515: nit:  is it better to do sorting for config here for serverconfig? order of keys change can be different while iterating over it each time https://go.dev/blog/maps#:~:text=When%20iterating%20over,in%20key%20order%3A. this may unneccessarily update configmap across reconciles?

api/v1alpha1/valkeycluster_types.go
jdheyburn: I'm actually not entirely confident my suggestion this is the path we should go down, but we can leave this here for now and revisit it in future.

Apologies, I'm working on fleshing out the design proposal a bit more this week, and have decided against adding any dependencies this early on. ValkeyCluster should be independent for the time being. We can change our mind while we're still in Alpha after 0.1.0 release. Thank you!

internal/controller/valkeycluster_controller.go
utdrmac: done

internal/controller/config.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Need to create new configMap
+	if needCreateConfigMap {
+		if err := r.Create(ctx, serverConfigMap); err != nil {
+			log.Error(err, "Failed to create server configMap")
+			r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "ConfigMapCreationFailed", "Failed to create server configMap: %v", err)
+			return err
+		} else {
+			r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "ConfigMapCreated", "Created server configMap")
+			return nil
```
sandeepkunusoth: the unit test was failing previously because event description was changed fyi 	
		r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "ConfigMapCreated", "Created ConfigMap with configuration")

to 
		r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "ConfigMapCreated", "Created server configMap")

utdrmac: Yes. It seemed excessive to have 3 checks for the same single line, especially when the "comment" can be somewhat arbitrary. Thus, I opted to remove the 3rd check. The e2e still checks for Normal event, and "ConfigMapCreated"

utdrmac (commented):
Everything is resolved. @sandeepkunusoth github says you've requested changes but I can't find anything unresolved when looking at the 'Files changed' tab.

internal/controller/config.go
```diff
@@ -0,0 +1,177 @@
...
+	if err := r.Update(ctx, serverConfigMap); err != nil {
+		log.Error(err, "Failed to update server configMap")
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "ConfigMapUpdateFailed", "Failed to update server configMap: %v", err)
+		return err
+	}
+
+	r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "ConfigMapUpdated", "Synchronized server configMap")
+
+	// All is good. Server configMap will be auto-mounted in the deployment
+	return nil
```
sandeepkunusoth, line 219: i dont think updated configmap will be auto mounted in the existing deployments during update. Did u test this

internal/controller/config.go
```diff
@@ -0,0 +1,177 @@
...
+	// Register ownership of the configMap
+	if err := controllerutil.SetControllerReference(cluster, serverConfigMap, r.Scheme); err != nil {
+		log.Error(err, "Failed to grab ownership of server configMap")
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "ConfigMapCreationFailed", "Failed to grab ownership of server configMap: %v", err)
+		return err
+	}
+
+	// Calculate hash of existing config parameters
+	origServerConfigHash := fmt.Sprintf("%x", sha256.Sum256([]byte(serverConfigMap.Data[configFileKey])))
+
```
sandeepkunusoth, line 182: ere we are checking only hash of configFileKey. what happens if we change readiness-check.sh and "liveness-check.sh" configs later in operator i think this will not be updated in existing clusters.

i feel this file has become too much complicated(i mean lot of things are happening in upsertConfigMap function)

utdrmac: Those shell scripts are not hashed because those scripts are only in the "internal" configMap. Any changes to those scripts would require a version bump of the operator. When the user deploys the new operator, the configMap will get updated with the new script contents, and then kubelet will update the mounted scripts inside each pod. Thus, it's not necessary to check them.

> i feel this file has become too much complicated

Seems pretty straightforward: build the default config, append user config, search for/fetch existing configMap, calculate/compare hashes, check if needs creating/updating.

sandeepkunusoth: i don't think kubelet will automatically update pods with updated configmap there are some custom controllers out there to do this like https://github.com/stakater/Reloader. but whenever eadiness-check.sh/liveliess check is updated within operator its possible that operator may reconcile and restart pods with updated configmap which i dont think its happening today.

internal/controller/config.go
```diff
@@ -0,0 +1,177 @@
...
+	configFileKey = "valkey.conf"
+)
+
+//go:embed scripts/*
+var scripts embed.FS
+
+// Create or update a default valkey.conf
+// If additional config is provided, append to the default map
+func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+
```
sandeepkunusoth, line 117: can we rename this file to configmap.go

internal/controller/config.go
sandeepkunusoth: in upsertConfigMap, when the ConfigMap does not exist, you create it with:
readiness-check.sh
liveness-check.sh
valkey.conf

But when the ConfigMap does exist, your update path only sets:

serverConfigMap.Data[configFileKey] = serverConfig

So if we ever change those scripts in a later operator version, existing clusters won’t get the new script contents (because we never re-set those keys on update). and operator doesnt detech changes in .sh contents

utdrmac: Ahh, sorry. I see what you mean now. If the scripts will only be updated in conjunction with operator updates, shouldn't we handle this via version checking? If the operator version on the cluster CR label is less than the operator version running, update the scripts? I'm not a fan of calculating the hashes of these two static scripts every 30s when they won't change unless the operator upgrades.

internal/controller/config.go
bjosv: It seem to update the configfile in the pods when I test it on ~~kind~~ minikube now, but it takes some time before the file is getting updated.

But, I now see that we are not using `SubPath` in the VolumeMounts on containers..
We probably should (?) since both `/scripts` and `/configs` contains all files from the configmap.
The problem then will be that the volume mount will not receive updates when the ConfigMap changes though..

internal/controller/config.go
bjosv: Good point

internal/controller/config.go
```diff
@@ -0,0 +1,177 @@
...
+		return err
+	}
+
+	// Calculate hash of existing config parameters
+	origServerConfigHash := fmt.Sprintf("%x", sha256.Sum256([]byte(serverConfigMap.Data[configFileKey])))
+
+	// Calculate hash of new config parameters
+	newServerConfigHash := fmt.Sprintf("%x", sha256.Sum256([]byte(serverConfig)))
+
+	// Was the configMap changed? This is an invalid action, as users should modify
```
bjosv: This is a bit hard to follow with a lot of flags (maybe I need more coffee..)
It possibly simpler to just create the configmap at the start of this function, including its annotation, and then return thereafter.
This part would only handle the cases when the configmap already exists in some form.
I think that would remove the need of `needCreateConfigMap` and `origConfigModified`

internal/controller/config.go
bjosv: Updating the configmap is one thing, but the Valkey instances will not use these updates anyway in the current form. Is that what you where thinking on @sandeepkunusoth ?
Maybe the update part in this PR could be moved to its own PR, which could include the pod restart/config reload depending on how we decide.

internal/controller/utils.go
```diff
@@ -43,3 +44,44 @@ func labels(cluster *valkeyv1.ValkeyCluster) map[string]string {
...
+	}
+	return annotations[key] == value
+}
+
+// This function takes a K8S object reference (eg: pod, secret, configmap, etc),
+// and a key, and value to add to, or replace an existing, annotation within the object.
+// Returns true if the annotation was added, or updated
+func upsertAnnotation(o metav1.Object, key string, val string) bool {
+
+	updated := false
```
bjosv: We could skip this flag and just return true/false as well

utdrmac: Fixed

internal/controller/config.go
```diff
@@ -0,0 +1,188 @@
...
+// If additional config is provided, append to the default map
+func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+
+	log := logf.FromContext(ctx)
+
+	// Hash writer for embedded scripts, and configuration parameters
+	hashConfigContents := sha256.New()
+
+	// Embed readiness check script
+	readiness, err := scripts.ReadFile("scripts/readiness-check.sh")
```
jdheyburn, line 121: I don't know if this is nitpicking, but does it make sense to have scripts in a configmap named `mycluster-config`? That to me indicates that the ConfigMap contains only configurations.

I would argue to keep `-config` out of the resource name and just keep it to the name of the cluster.

internal/controller/config.go
```diff
@@ -0,0 +1,188 @@
...
+
+			if slices.Contains(valkeyiov1alpha1.NonUserOverrideConfigParameters, k) {
+				log.Error(nil, "Prohibited valkey server config", "parameter", k)
+				r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "ConfigMapUpdateFailed", "UpsertConfigMap", "Prohibited config: %v", k)
+				continue
+			}
+
+			serverConfig += k + " " + specConfig[k] + "\n"
+		}
+	}
```
jdheyburn: Could this be extracted into its own function? Something like:

```go
if len(cluster.Spec.ValkeySpec.Configuration) > 0 {
    appendSpecConfig(&serverConfig, cluster.Spec.ValkeySpec.Configuration)
}
```

internal/controller/config.go
jdheyburn: Agree that we should omit live applying changes to the config for another PR

internal/controller/config.go
utdrmac: Is your suggestion to move the scripts to their own configMap? I have no objection with that.

internal/controller/config.go
utdrmac: I guess it could. What's the benefit though of having the same code but in another function? Doesn't that introduce additional code jumps?

jdheyburn: The benefit is readability - ideally we want to keep isolated units of work behind functions that are named well so that when reading the entirety of `upsertConfigMap` it is easy to follow along and not get bogged down on the low-level implementation of building the extra config :)

internal/controller/config.go
jdheyburn: I think it's fine to leave as is. We can refactor later on

config/crd/bases/valkey.io_valkeynodes.yaml
```diff
@@ -2639,10 +2639,6 @@ spec:
                       More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
                     type: object
                 type: object
```
utdrmac, line 2641: @jdheyburn FYI on this change. I renamed it to "better align"(?) with the other changes since the configmap has both config and the scripts.

internal/controller/valkeycluster_controller.go
```diff
@@ -524,16 +480,16 @@ func buildClusterValkeyNode(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex
...
-			UsersACLSecretName:   getInternalSecretName(cluster.Name),
+			Image:              cluster.Spec.Image,
+			WorkloadType:       cluster.Spec.WorkloadType,
+			Resources:          cluster.Spec.Resources,
+			NodeSelector:       cluster.Spec.NodeSelector,
+			Affinity:           cluster.Spec.Affinity,
+			Tolerations:        cluster.Spec.Tolerations,
+			Exporter:           cluster.Spec.Exporter,
+			Containers:         cluster.Spec.Containers,
+			UsersConfigMapName: getConfigMapName(cluster.Name),
```
jdheyburn: I wonder if `UsersConfigMapName` implies it is related to Users, in a similar way that  `UsersACLSecretName` does. Or does Users mean the human deploying the operator?

internal/controller/config.go
```diff
@@ -0,0 +1,224 @@
...
+	configHashKey  = "valkey.io/config-hash"
+	configFileKey  = "valkey.conf"
+
+	readinessScriptKey = "readiness-check.sh"
+	livenessScriptKey  = "liveness-check.sh"
+
+	// This hash should be updated whenever the contents of either script changes, which would
+	// coincide with operator version bump.
+	// $ cat internal/controller/scripts/{liveness-check.sh,readiness-check.sh} | sha256sum
+	scriptsHash = "8531132f52ac311772dfcb45c107c34ab05e719a0df644cc332512277b564346"
```
jdheyburn: Is there not a way we can have this calculated for us?

internal/controller/config.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// Average-ish length of Valkey parameter + value
+	averageParameterLength = 20
+)
+
+//go:embed scripts/*
+var scripts embed.FS
+
+func getConfigMapName(clusterName string) string {
+	return clusterName + "-config"
```
jdheyburn: Can we prepend `valkey-` to this? I'm thinking in case there is a conflict and we can minimise that.

internal/controller/config.go
```diff
@@ -0,0 +1,224 @@
...
+
+		writeConfigLine(&configBuilder, param, specConfig[param])
+	}
+
+	return configBuilder.String()
+}
+
+// Create or update a default valkey.conf
+// If additional config is provided, append to the default map
+func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
```
jdheyburn, line 116: I've started refactoring other upserts to use `controllerutil.CreateOrUpdate`, could the same be done here for consistency? It might help reduce some lines too.

internal/controller/config.go
```diff
@@ -0,0 +1,224 @@
...
+
+// Return a base config of parameters that users shouldn't be able to override
+func getBaseConfig() string {
+	return `# Base operator config
+cluster-enabled yes
+protected-mode no
+cluster-node-timeout 2000
+aclfile /config/users/users.acl
+`
+}
```
jdheyburn, line 79: Could this be in a `map[string]string`? Then we could merge this easily with what is provided in `cluster.Spec.Config`, with these keys taking priority.

It would mean we don't need to specify `NonUserOverrideConfigParameters` either.

internal/controller/config.go
```diff
@@ -0,0 +1,224 @@
...
+		if slices.Contains(valkeyiov1alpha1.NonUserOverrideConfigParameters, param) {
+			log.Error(nil, "Prohibited valkey server config", "parameter", param)
+			continue
+		}
+
+		writeConfigLine(&configBuilder, param, specConfig[param])
+	}
+
+	return configBuilder.String()
+}
```
jdheyburn, line 112: If the merged config values are a `map[string][string]`, would that simplify the string builder because we would be able to do the below (untested code)?

```go
for key, val := range mergedConfig {
  configBuilder.WriteString(fmt.Sprintf("%s %s\n", key, val))
}
```

internal/controller/config.go
```diff
@@ -0,0 +1,224 @@
...
+	// Exit early if nothing
+	if len(specConfig) == 0 {
+		return ""
+	}
+
+	log := logf.FromContext(ctx)
+
+	// Build the config
+	var configBuilder strings.Builder
+	configBuilder.Grow(len(specConfig) * averageParameterLength)
```
jdheyburn: I wonder if the use of Grow here is prematurely optimising, would it simplify the logic if we removed it?

internal/controller/valkeycluster_controller.go
utdrmac: Users, as the human deploying the operator. There are some config params that the operator sets, and others that the users would set.

internal/controller/config.go
utdrmac: The goal was to not have the operator (re)calculating sha256 on static file contents every 30s. As for an "outside" solution, I could probably do something similar to how version numbers are done, by linking the value to the variable during compile time?

internal/controller/config.go
utdrmac: Yup. I'll push that shortly.

internal/controller/config.go
utdrmac: Probably. I will look into it.

internal/controller/config.go
utdrmac: Good point.

internal/controller/config.go
utdrmac: Most likely; Will check.

internal/controller/config.go
utdrmac: It would simplify the logic by 1 line. I read that this pre-allocation was a simple memory optimization if the size of the eventual string was somewhat known.

internal/controller/valkeycluster_controller.go
jdheyburn: My interpretation was that `Users` here meant for ACLs. If it is ambiguous maybe it should be removed?

internal/controller/config.go
jdheyburn: How do other operators do this? Could we perform a faster check before the sha256? 

```
has_changed = false
if len(current) != len(desired):
  has_changed = true
elif sha256(current) != sha256(desired):
  has_changed = true
```

internal/controller/config.go
jdheyburn: I wonder if this would help with the script-hash that is being set in the PR today.

internal/controller/config.go
utdrmac: I can speak for 3 of Percona's operators (mysql, pxc, mongo): The liveness, readiness, and health check scripts are part of the operator docker image. The image is used as init-container, where a separate entrypoint runs to copy the scripts into the running pod. The init container finishes, then the main container, using the same image with default entrypoint, launches as a regular container and the scripts are there to be used.
This has some benefits: 1) no checksum calculations, 2) nothing to 'embed', and 3) upgrades to the script become automatic new release/version of the operator.
Once we decide on if we will be making an operator-specific image (ie: for including modules), perhaps we could use a similar model/process?

jdheyburn: That could work, but until we decide on that - could we instead calculate the sha256sum on operator boot? Given that the scripts would be coupled with the operator version, we would then be able to avoid calculating the sha on every reconcile. How does that sound?

utdrmac: Sounds doable!

utdrmac:
Just an update, that I'm working on the minor changes requested above, but I'm getting golang crashes when building the docker image. `make build` works just fine, but `make docker-build` is not. I'm investigating what has recently changed, or if it's something on my end. EDIT: fixed. erase all / fresh install (turn it off then back on)

jdheyburn:
I've deployed this locally, and works perfectly! @utdrmac there are some merge conflicts from the latest PR I merged - which was @hieu2102's PR adding operator and exporter user. If you clear these up then I think we can get this approved and merged.

internal/controller/valkeynode_resources.go
```diff
@@ -51,7 +51,7 @@ func valkeyNodeLabels(node *valkeyiov1alpha1.ValkeyNode) map[string]string {
 
 // buildValkeyNodeConfigMap builds a ConfigMap containing the embedded liveness
 // and readiness probe scripts, plus an empty valkey.conf.
-// The ConfigMap is named after valkeyNodeResourceName(node).
+// The ConfigMap is named via config.go:getConfigMapName(node).
```
bjosv, line 54: nit: remove this line or change to GetServerConfigMapName,  getConfigMapName was maybe used in a previous iteration.

bjosv (commented):
LGTM, just added a nit for the record.
Just a rebase needed, then open comments can be addressed in separate PR.

utdrmac:
@sandeepkunusoth Please check my merge with your latest TLS, as the config is moved into a separate file.

internal/controller/config.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "ConfigMapUpdated", "UpsertConfigMap", "Synchronized server configMap")
+
+	// All is good. configMap was updated with new contents.
+	return nil
+}
+
+// getTLSConfig returns the TLS configuration for a ValkeyCluster.
+func getTLSConfig(ctx context.Context, c client.Client, secretName, serverName, namespace string) (*tls.Config, error) {
```
sandeepkunusoth: getTLSConfig doesn't need to be moved here as its not related to configmap or valkey.conf. everything else looks good

internal/controller/config.go
```diff
@@ -0,0 +1,247 @@
...
+		sortedKeys := slices.Sorted(maps.Keys(userConfig))
+
+		for _, param := range sortedKeys {
+			writeConfigLine(&configBuilder, param, userConfig[param])
+		}
+	}
+
+	// Add base config
+	writeConfigLine(&configBuilder, "#", "Base Config")
+	for param, val := range baseConfig {
```
sandeepkunusoth: do we need sorting here for baseConfig similar to  we can verify in unit tests if order is same everytime. 		
`sortedKeys := slices.Sorted(maps.Keys(userConfig))`

internal/controller/config.go
```diff
@@ -0,0 +1,247 @@
...
+	baseConfig := getBaseConfig(cluster)
+	userConfig := cluster.Spec.Config
+
+	// Build the config
+	var configBuilder strings.Builder
+	configBuilder.Grow((len(baseConfig) + len(userConfig)) * averageParameterLength)
+
+	// User-provided config goes first, and base config goes last. This prevents
+	// users from overriding key parameters as Valkey uses the last value in the file.
+
```
sandeepkunusoth, line 92: will there be any startup errors when we have duplicate key valus betwee user config and base config?

Makefile
```diff
@@ -96,7 +96,7 @@ setup-test-e2e: ## Set up a Kind cluster for e2e tests if it does not exist
 .PHONY: test-e2e
 test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
 	KIND=$(KIND) KIND_CLUSTER=$(KIND_CLUSTER) go test -tags=e2e ./test/e2e/ -v -ginkgo.v -ginkgo.label-filter "${TEST_LABELS}" -timeout 30m
-	$(MAKE) cleanup-test-e2e
+#	$(MAKE) cleanup-test-e2e
```
sandeepkunusoth: is this comment intended change?

sandeepkunusoth (approved):

jdheyburn:
@utdrmac I rebased this branch from main, looks like its causing some e2e tests to fail - would you mind taking a look?

utdrmac:
🎉  Thanks everyone for helping, commenting, and testing!

Issue #28, opened 2025-12-18
(feat): Add Metrics Exporter Sidecar to ValkeyCluster Pods


jdheyburn:
# Add Metrics Exporter Sidecar to ValkeyCluster Pods

## Summary

Add a redis_exporter sidecar container to ValkeyCluster pods to expose Valkey metrics for Prometheus. The sidecar should be configurable via the existing `ExporterSpec` fields.

## Current State

- ✅ `ExporterSpec` type exists with `Image`, `Resources`, and `Enabled` fields (lines 60-72 in `api/v1alpha1/valkeycluster_types.go`)
- ✅ Default is `enabled: true`
- ❌ No exporter sidecar added to deployments

## Goal

When a ValkeyCluster is created with `spec.exporter.enabled=true` (default), each pod should have:
1. **valkey-server container** - runs Valkey
2. **metrics-exporter container** - exposes metrics on port 9121

### File: `internal/controller/deployment.go`

Update `createClusterDeployment()` function to conditionally add the exporter sidecar:

```go
func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.Deployment {
    image := DefaultImage
    if cluster.Spec.Image != "" {
        image = cluster.Spec.Image
    }
    
    // Start with valkey-server container
    containers := []corev1.Container{
        {
            Name:  "valkey-server",
            Image: image,
            Command: []string{
                "valkey-server",
                "/config/valkey.conf",
            },
            Ports: []corev1.ContainerPort{
                {Name: "client", ContainerPort: DefaultPort},
                {Name: "cluster-bus", ContainerPort: DefaultClusterBusPort},
            },
            // ... existing LivenessProbe, ReadinessProbe, VolumeMounts ...
        },
    }
    
    // Add exporter sidecar if enabled
    if cluster.Spec.Exporter.Enabled {
        exporterImage := DefaultExporterImage
        if cluster.Spec.Exporter.Image != "" {
            exporterImage = cluster.Spec.Exporter.Image
        }
        
        exporterContainer := corev1.Container{
            Name:  "metrics-exporter",
            Image: exporterImage,
            Args: []string{
                "--redis.addr=localhost:6379",
            },
            Ports: []corev1.ContainerPort{
                {
                    Name:          "metrics",
                    ContainerPort: DefaultExporterPort,
                    Protocol:      corev1.ProtocolTCP,
                },
            },
            LivenessProbe: &corev1.Probe{
[...]

PR #49, opened 2026-01-12
(feat): Add Metrics Exporter Sidecar to ValkeyCluster Pods

sandeepkunusoth:
This PR implements https://github.com/valkey-io/valkey-operator/issues/28 by adding redis_exporter sidecar to ValkeyCluster pods (configurable via spec.exporter) so each pod exposes Valkey metrics on port 9121.  also includes e2e tests to validate redis exporter metrics and small refactors to the existing e2e test suite(moved some common code from e2e_test.go to e2e_suite_test.go).

### DIFF
--- a/internal/controller/metrics_exporter.go
+++ b/internal/controller/metrics_exporter.go
@@ -17,6 +17,8 @@ limitations under the License.
 package controller
 
 import (
+	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
@@ -31,7 +33,7 @@ func generateMetricsExporterContainerDef(cluster *valkeyiov1alpha1.ValkeyCluster
 	return corev1.Container{
 		Name:  "metrics-exporter",
 		Image: exporterImage,
-		Args:  []string{"--redis.addr=localhost:6379"},
+		Args:  []string{fmt.Sprintf("--redis.addr=localhost:%d", DefaultPort)},
 		Ports: []corev1.ContainerPort{
 			{
 				Name:          "metrics",
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -36,6 +36,7 @@ import (
 )
 
 // TODO divide this file into multiple files (manager_test.go, valkeycluster_test.go, etc)
+// https://github.com/valkey-io/valkey-operator/issues/51
 var _ = Describe("Manager", Ordered, func() {
 	var controllerPodName string
 	// After each test, check for failures and collect logs, events,
--- a/test/e2e/metrics_exporter_test.go
+++ b/test/e2e/metrics_exporter_test.go
@@ -1,3 +1,22 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2025 Valkey Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package e2e
 
 import (
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -22,6 +22,8 @@ import (
 	valkeyv1 "valkey.io/valkey-operator/api/v1alpha1"
 )
 
+const appName = "valkey"
+
 // Labels returns a copy of user defined labels including recommended:
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
 func labels(cluster *valkeyv1.ValkeyCluster) map[string]string {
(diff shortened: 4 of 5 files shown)


internal/controller/deployment.go
```diff
@@ -23,12 +23,84 @@ import (
 	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
 )
 
-func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.Deployment {
+func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Container {
```
jdheyburn, line 26: Nice

test/e2e/e2e_test.go
```diff
@@ -35,74 +35,9 @@ import (
...
-// metricsServiceName is the name of the metrics service of the project
-const metricsServiceName = "valkey-operator-controller-manager-metrics-service"
-
-// metricsRoleBindingName is the name of the RBAC that will be created to allow get the metrics data
-const metricsRoleBindingName = "valkey-operator-metrics-binding"
-
-// valkeyClientImage is the image used to verify cluster access.
-const valkeyClientImage = "valkey/valkey:9.0.0"
-
+// TODO divide this file into multiple files (manager_test.go, valkeycluster_test.go, etc)
```
jdheyburn, line 38: Could you raise this into an issue and link it back in the comment please?

internal/controller/metrics_exporter.go
```diff
@@ -0,0 +1,66 @@
...
+// generateMetricsExporterContainerDef generates the container definition for the metrics exporter sidecar.
+func generateMetricsExporterContainerDef(cluster *valkeyiov1alpha1.ValkeyCluster) corev1.Container {
+	exporterImage := DefaultExporterImage
+	if cluster.Spec.Exporter.Image != "" {
+		exporterImage = cluster.Spec.Exporter.Image
+	}
+	return corev1.Container{
+		Name:  "metrics-exporter",
+		Image: exporterImage,
+		Args:  []string{"--redis.addr=localhost:6379"},
```
jdheyburn: Can we replace `6379` with `DefaultPort`?

jdheyburn (commented):
Overall looks great! Just had a few minor comments

api/v1alpha1/valkeycluster_types.go
```diff
@@ -70,6 +70,7 @@ type ValkeyClusterSpec struct {
 	Affinity *corev1.Affinity `json:"affinity,omitempty"`
 
 	// Metrics exporter options
+	// +kubebuilder:default:={enabled:true}
```
bjosv, line 73: Is this change needed? And if yes; can we remove `// +kubebuilder:default=true`  on line 88 in the ExporterSpec type now?

test/e2e/metrics_exporter_test.go
```diff
@@ -0,0 +1,313 @@
...
+      requests:
+        memory: "32Mi"
+        cpu: "10m"
+      limits:
+        memory: "64Mi"
+        cpu: "50m"
+`, valkeyName)
+
+			// Create temporary YAML file
+			manifestFile := filepath.Join(os.TempDir(), fmt.Sprintf("%s-%d.yaml", valkeyName, time.Now().UnixNano()))
```
bjosv, line 60: Do we need a unique name here?
We don't do this in the degraded state test where we use `valkeycluster-degraded.yaml` directly.

test/e2e/metrics_exporter_test.go
```diff
@@ -0,0 +1,313 @@
...
+			output, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to apply ValkeyCluster: %s", output))
+
+			By("waiting for deployment to be created")
+			Eventually(func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "deployment", "-l", "app.kubernetes.io/instance="+valkeyName)
+				out, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Deployment not found")
+				g.Expect(out).To(ContainSubstring(valkeyName))
+			}, 2*time.Minute, 5*time.Second).Should(Succeed())
```
bjosv, line 78: The default is 2 minutes, so we could remove all these extra timeouts to get rid of clutter
```suggestion
			}).Should(Succeed())
```

test/e2e/metrics_exporter_test.go
```diff
@@ -0,0 +1,313 @@
...
+  name: %s
+spec:
+  shards: 1
+  replicas: 0
+  exporter:
+    enabled: false
+`, valkeyName)
+
+			// Create temporary YAML file
+			manifestFile := filepath.Join(os.TempDir(), fmt.Sprintf("%s-%d.yaml", valkeyName, time.Now().UnixNano()))
```
bjosv, line 250: same comment here regarding including time in the filename.
If the reason is to enable it to run in parallel, maybe https://pkg.go.dev/os#CreateTemp is a better option to get uniqueness.

test/e2e/metrics_exporter_test.go
```diff
@@ -0,0 +1,313 @@
...
+				err := portForwardCmd.Process.Kill()
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			// Wait a bit for port-forwarding to be ready
+			time.Sleep(2 * time.Second)
+
+			By("verifying common Valkey metrics are exposed")
+			Eventually(func(g Gomega) {
+				cmd := exec.Command("curl", "-s", "http://localhost:9121/metrics")
```
bjosv, line 199: This test now requires curl to be installed on the system.
An option that is to use a curl pod, that how the kubebuilder scaffolding gets the metrics from the controller.

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: Yes this change is needed otherwise by default its becoming false. I will remove other change in next PR.

test/e2e/metrics_exporter_test.go
sandeepkunusoth: yes forgot about this suggestion from past PRs. will update this in next PR.

PR #54, opened 2026-01-15
fixed e2e tests failing intermittently & review comments from past PR 

sandeepkunusoth:
This PR fixes e2e test failing due to https://github.com/valkey-io/valkey-operator/issues/43 temporarily by deleting replica deployment in e2e tests instead of deleting random deployment (either replica or shard previously). We need to add back  e2e test to delete shard and allow it to recover once above issue is fixed.

Also addressed review comments from metrics exporter PR @bjosv https://github.com/valkey-io/valkey-operator/pull/49

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -391,8 +391,6 @@ var _ = Describe("Manager", Ordered, func() {
 				g.Expect(output).To(ContainSubstring("ServiceCreated"), "ServiceCreated event should appear in describe")
 				g.Expect(output).To(ContainSubstring("ConfigMapCreated"), "ConfigMapCreated event should appear in describe")
 				g.Expect(output).To(ContainSubstring("DeploymentCreated"), "DeploymentCreated event should appear in describe")
-				g.Expect(output).To(ContainSubstring("NodeAdding"), "NodeAdding event should appear in describe")
-				g.Expect(output).To(ContainSubstring("NodeAdded"), "NodeAdded event should appear in describe")
 				// TODO PrimaryCreated, ClusterMeet events are not always captured due to rate-limiting issues
 				// fix this removing events which are not important
 				// ReplicaCreated and ClusterReady may not always appear in describe output due to:
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -303,37 +303,45 @@ func GetReplicaDeployment(selector string) (string, error) {
 	}
 	podNames := strings.Fields(output)
 
-	// Compile the regex once
-	replicaPattern := regexp.MustCompile(`(?m)^\s*\d+:S`)
+	// Compile the regex once to capture the role (M or S)
+	rolePattern := regexp.MustCompile(`^\s*\d+:([MS])`)
 
 	for _, podName := range podNames {
-		// Get logs (last 100 lines should be enough to see the role)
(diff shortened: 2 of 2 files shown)


test/e2e/e2e_test.go
```diff
@@ -401,8 +393,8 @@ var _ = Describe("Manager", Ordered, func() {
 				g.Expect(output).To(ContainSubstring("DeploymentCreated"), "DeploymentCreated event should appear in describe")
 				g.Expect(output).To(ContainSubstring("NodeAdding"), "NodeAdding event should appear in describe")
 				g.Expect(output).To(ContainSubstring("NodeAdded"), "NodeAdded event should appear in describe")
```
bjosv: When running `make test-e2e` continuously on my machine I don't always get the `NodeAdded` event either (~1 in 3 runs).

test/e2e/e2e_test.go
```diff
@@ -522,19 +514,13 @@ spec:
...
-
-				deployments := utils.GetNonEmptyLines(output)
-				g.Expect(len(deployments)).To(BeNumerically(">", 0))
-
-				// Get the first deployment
-				deploymentToDelete = deployments[0]
+				// selecting replica deployment to delete as we have intermittent issue
+				// with master deployment deletion resulting in cluster not being able to recover
+				// https://github.com/valkey-io/valkey-operator/issues/43
+				var err error
```
bjosv: remove?

test/utils/utils.go
```diff
@@ -288,3 +289,53 @@ func GetEvents(resourceName string) (map[string]bool, map[string]bool, error) {
...
+		cmd = exec.Command("kubectl", "logs", podName, "--tail=100")
+		logs, err := Run(cmd)
+		if err != nil {
+			// Ignore errors getting logs (pod might be initializing)
+			continue
+		}
+
+		// Check for specific pattern indicating a replica (e.g., "1:S ...")
+		// The user specified "logs start with :S", which refers to the standard Valkey log format column.
+		if replicaPattern.MatchString(logs) {
```
bjosv: We should probably start searching from the last logline to be precise, i.e. in case there has been a role change within that 100 lines of log. Not sure if that can happen in our case, and this is a temporary fix anyway, so its fine as it is I guess.

test/utils/utils.go
```diff
@@ -288,3 +289,53 @@ func GetEvents(resourceName string) (map[string]bool, map[string]bool, error) {
 
 	return normalEvents, warningEvents, nil
 }
+
+// TODO: use this function until we have a better way to find the replica deployment.
+// maybe we need to update valkey cluster status to include master and replica deployment identifiers.
```
bjosv, line 294: I guess we can't fully rely on the K8s valkey status or labels since valkey can change its topology when there is a network issue. This is an acceptable temporary solution.

bjosv (commented):
We probably need to remove the expectation of a `NodeAdded` event as well.
When I remove that I can run 30 times locally without any issues.

test/e2e/e2e_test.go
sandeepkunusoth: removed this for now to be safe.

test/utils/utils.go
sandeepkunusoth: thanks for pointing out updated.

PR #56, opened 2026-01-16
(feat): Support user creation on cluster init

### LINKED ISSUE #36: (feat)  Support ACL user creation on cluster init
Add support for adding Valkey ACL users during ValkeyCluster

utdrmac:
Resolves #36 

This feature request allows for the creation of users on cluster init. Users can be added by creating an entry under `spec.valkeyConfig.auth.users`  Permissions must be provided otherwise the user is skipped. A password, or password reference must be provided otherwise the user is skipped. Passwords can be provided as plaintext, or as sha256 hash directly in the spec, or the password can be searched for in a referenced secret. If the secret is not specified, a secret with the name "${clustername}-secret" will be searched for.

The operator will fetch, or create, a secret named "internal-${clustername}-secret". This secret will be populated with the ACLs built from the spec. It will be automatically mounted to /config/users/ as users.acl. The default config has been adjusted to look for this file.

### DIFF
--- a/internal/controller/users_test.go
+++ b/internal/controller/users_test.go
@@ -32,12 +32,11 @@ func TestBuildAclFileContents(t *testing.T) {
 	// Simulate a Secret
 	testSecrets := &corev1.Secret{
 		Data: map[string][]byte{
-			"charlie":  []byte("charliepassword"),
 			"davidref": []byte("c90502e005ee957f29645e21d1e27f5bbfce539e38c949a00dfc12270f47fc59"),
 		},
 	}
 
-	// alice should succeed as there is permissions, and valid password
+	// alice should succeed as there are permissions, and valid password
 	expected := "user alice on #b6366487efc982bfe450c46753917ee883f71a3f4fa5cdc5bde78d1784842e73 +@list +@connection ~jobs:*"
 	acl := buildAclFileContents(ctx, map[string]valkeyiov1alpha1.UserAcl{
 		"alice": {
@@ -69,7 +68,7 @@ func TestBuildAclFileContents(t *testing.T) {
 		t.Errorf("charlie ACL Failed. Expected %s; got %s", expected, acl)
 	}
 
-	// david should succeed as there is permissions, and valid key ref
+	// david should succeed as there are permissions, and a valid key ref
 	expected = "user david on #c90502e005ee957f29645e21d1e27f5bbfce539e38c949a00dfc12270f47fc59 +@all"
 	acl = buildAclFileContents(ctx, map[string]valkeyiov1alpha1.UserAcl{
 		"david": {


api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,50 @@ type ValkeyClusterSpec struct {
...
+	// Auth-related structure for handling users, and acls
+	Auth AuthSpec `json:"auth,omitempty"`
+}
+
+// AuthSpec contains authorization, user, and ACL-related configurations
+type AuthSpec struct {
+
+	// Enable, or disable protected mode
+	// +kubebuilder:default:=true
+	ProtectedMode bool `json:"protectedmode,omitempty"`
```
sandeepkunusoth: where are we using ProtectedMode?

internal/controller/users.go
```diff
@@ -0,0 +1,216 @@
...
+)
+
+const (
+	hashLength        = 64
+	hashAnnotationKey = "valkey.io/internal-acl-hash"
+)
+
+func getInternalSecretName(cn string) string {
+	return "internal-" + cn + "-secret"
+}
```
sandeepkunusoth, line 93: can we have default name of acl secret name like clusterName-acl or clusterName-acl-secret i prefer lusterName-acl as clusterName-acl-secret becomes too long.

this is better to identify acl secret between many other secrets.

internal/controller/users.go
```diff
@@ -0,0 +1,216 @@
...
+	log := logf.FromContext(ctx)
+
+	// Shortcut
+	auth := &cluster.Spec.ValkeySpec.Auth
+
+	// Look for a Secret matching the user-provided name, or clusterName-secret
+	userSecretName := cluster.Name + "-secret"
+	if auth.UsersSecretRef != "" {
+		userSecretName = auth.UsersSecretRef
+	}
```
sandeepkunusoth, line 106: if auth.UsersSecretRef is optional existing secret when its empty in input spec. i feel we don't need to query for default secret (cluster.Name + "-secret" ) again and log error  "failed to fetch acl secret")

internal/controller/users.go
```diff
@@ -0,0 +1,216 @@
...
+	// Calculate hash of the ACL file contents
+	internalAclHash := fmt.Sprintf("%x", sha256.Sum256(aclFileContents))
+
+	// Compare hash to the one already attached to the internal secret, if present.
+	// If the hashes are different, then we need to update the internal secret with
+	// the new file contents and update the hash annotation. If the hashes are the
+	// same, don't update as that would cause infinite reconciliation
+
+	if needsUpdate := upsertAnnotation(internalAclSecrets, hashAnnotationKey, internalAclHash); !needsUpdate {
+		log.V(2).Info("internal secrets unchanged")
```
sandeepkunusoth: when this update is performed on users acl will this update valkey deployments with latest acl? if not we need to add a TODO here

internal/controller/users.go
```diff
@@ -0,0 +1,216 @@
...
+func buildAclFileContents(ctx context.Context, users map[string]valkeyiov1alpha1.UserAcl, usersSecrets *corev1.Secret) []byte {
+
+	log := logf.FromContext(ctx)
+
+	// Holds the ACLs
+	var aclFileContents string
+
+	// Loop over users and build acl line
+	for un, acl := range users {
+
```
sandeepkunusoth, line 222: is it better to do sorting for user map here? order of keys change can be different while iterating over it each time https://go.dev/blog/maps#:~:text=When%20iterating%20over,in%20key%20order%3A. this may generate different hash and update acl secret across reconciles?

internal/controller/deployment.go
```diff
@@ -141,6 +146,15 @@ func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.De
 								},
 							},
 						},
+						{
+							Name: "users-acl",
+							VolumeSource: corev1.VolumeSource{
+								Secret: &corev1.SecretVolumeSource{
+									SecretName: getInternalSecretName(cluster.Name),
+									Optional:   func(b bool) *bool { return &b }(true),
```
sandeepkunusoth: any reason for adding optional true here?

internal/controller/users.go
utdrmac: I don't understand your statement. The user can choose to put their secrets in either "cluster.Name-secret" (ie: default), or "my-foobar-secret"  If the user specifies a secret ref, then fetch it. If no secret ref specified, then fetch the default. There's only ever 1 API fetch. If that fetch returns not found, that's ok, and we continue on because passwords can still be present.

sandeepkunusoth: in that case we dont need to log error below as UsersSecretRef is optional

api/v1alpha1/valkeycluster_types.go
utdrmac: We are not at this point. I was just trying to flush out more "auth" related items for the AuthSpec struct. Once #44 is merged, then the config function can examine additional params in AuthSpec if there are any we'd like to call out specifically. I can remove it for now.

internal/controller/users.go
utdrmac: done

internal/controller/users.go
utdrmac: ooohhh.. good point

internal/controller/deployment.go
utdrmac: I guess not since the "internal" secrets will always exist. I'll remove it.

internal/controller/users.go
utdrmac: if the user modifies either the CR spec, or the user-defined secret, the internal secret is updated. Since the internal secret is mounted inside each deployment, any changes/updates to it should appear in the file, yes. An improvement to this PR would be to execute ACL LOAD on the primaries, but I didn't want to do too much in 1 PR.

utdrmac: Testing this. I changed the password for 'alice' in the userSpecRef secret, and I can see the internal secret was updated, but the mounted file inside a pod was not updated. Also, I must not be monitoring the userSpecRef secret correctly because after I save the change, I do not see an immediate reconcile called in the operator. The "every 30s" reconcile goes, and that picks up the change.

utdrmac: The users.acl file inside each pod is _eventually_ updated to reflect the new internal secrets contents. It took a little over 2m to get updated. Is something that can be "force refreshed" via K8S api?

utdrmac: This should be fixed now. If a user modifies the user-defined secrets, it will trigger a reconcile and rebuild the internal secrets. However, there is no way to force K8S to update the secret-mount inside the pod; that is done whenever the kubelet does its routine update which could be 60-90s delayed. The next iteration of this feature will need to execute the ACLs manually on each primary, or we could just make the users wait for the file update, and call ACL LOAD when that happens.

internal/controller/users.go
utdrmac: The error is only logged/returned on api failure. a debug log is emitted if either secret is not found.

internal/controller/users.go
sandeepkunusoth: can we not include this automatic update in this PR. i feel like this can be solved in better way. we don't need to automatically listen to all user defined secrets.

utdrmac: How else then do we update the internal-ACL (and subsequently the mounted ACL secret) when users update the secret?

internal/controller/users.go
```diff
@@ -0,0 +1,299 @@
...
+	secretList := &corev1.SecretList{}
+	if err := r.List(ctx, secretList); err != nil {
+		log.Error(err, "failed to list referenced secrets")
+		return []reconcile.Request{}
+	}
+
+	// debug
+	for _, s := range secretList.Items {
+		log.V(1).Info("list item", "name", s.GetName())
+	}
```
sandeepkunusoth: i dont like the idea of watching all secrets across the cluster. and also we should not log secrets here

utdrmac: I changed the behavior on this. By using a label, you can List() secrets, and filter at the same time for matching conditions. It now should only fetch the "internal" secret, and submit the cluster for reconcile.

internal/controller/users.go
```diff
@@ -0,0 +1,284 @@
...
+	internalRefByKey  = "valkey.io/internal-acl-ref-by"
+	instanceLabel     = "app.kubernetes.io/instance"
+)
+
+func getInternalSecretName(cn string) string {
+	return cn + "-acl"
+}
+
+func getDefaultSecretName(cn string) string {
+	return cn + "-secret"
```
jdheyburn, line 49: I am not a fan of adding the resource Kind into the name of the resource, we already know it's a secret. Is there any reason to keep it?

```
kubectl get secret valkey-cluster-secret
```

internal/controller/users.go
```diff
@@ -0,0 +1,284 @@
...
+)
+
+const (
+	hashLength        = 64
+	hashAnnotationKey = "valkey.io/internal-acl-hash"
+	internalRefByKey  = "valkey.io/internal-acl-ref-by"
+	instanceLabel     = "app.kubernetes.io/instance"
+)
+
+func getInternalSecretName(cn string) string {
```
jdheyburn, line 44: Can we avoid acronyms on variable names? I understand `cn` here is referring to the `clusterName` that we're passing from higher up, but if we override the value with something else then it would not align with that. Perhaps `prefix` would be a better alternative? I saw that there are a few places this is being set.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,46 @@ type ValkeyClusterSpec struct {
...
+
+	// sha256 password
+	// kubebuilder:validation:MinLength=64
+	// kubebuilder:validation:MaxLength=65
+	// kubebuilder:XValidation:message="Password should be a sha256 hash"
+	Password string `json:"password,omitempty"`
+
+	// Reference to a key in UsersSecretRef containing the password for this user
+	// Defaults to the username
+	PasswordKeyRef string `json:"passwordKeyRef,omitempty"`
```
jdheyburn, line 115: ACLs can support multiple passwords for a user. Such an example case would be if the password needs to be rolled, you can have both passwords active while the client switches between them. 

Not critical for this PR, but so long as we're open to breaking changes while we're in alpha.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,46 @@ type ValkeyClusterSpec struct {
...
+	UsersSecretRef string `json:"usersSecretRef,omitempty"`
+
+	// Array of users with raw ACL, or reference to a key in the UsersSecretRef
+	Users map[string]UserAcl `json:"users,omitempty"`
+}
+
+type UserAcl struct {
+
+	// Raw ACL line, including password
+	Permissions string `json:"permissions,omitempty"`
```
jdheyburn, line 105: Thinking out loud: I wonder if the operator should help provide an abstraction away from the syntax for permissions. 

https://valkey.io/commands/acl-setuser/

config/samples/v1alpha1_valkeycluster.yaml
```diff
@@ -1,10 +1,33 @@
...
+      usersSecretRef: valkeycluster-sample-users
+      users:
+        alice:
+          permissions: +@list +@connection ~jobs:*
+          passwordKeyRef: alicepw
+        bob:
+          permissions: ~* +@all
+        charlie:
+          permissions: -@all +get ~secretkey ~valkey:*
+          password: sup3rS3cr3t
```
jdheyburn, line 28: Wouldn't this fail the validation that you have in place?

```
                            password:
                              description: |-
                                sha256 password
                                kubebuilder:validation:MinLength=64
                                kubebuilder:validation:MaxLength=65
                                kubebuilder:XValidation:message="Password should be a sha256 hash"
                              type: string
```

internal/controller/users.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// Extract usernames from the map, and sort to keep consistent processing order
+	sortedNames := make([]string, 0, len(users))
+	for k := range users {
+		sortedNames = append(sortedNames, k)
+	}
+	sort.Strings(sortedNames)
+
+	// Loop over users and build acl line
+	for _, un := range sortedNames {
```
jdheyburn, line 218: Could you replace un with userName please

internal/controller/users.go
```diff
@@ -0,0 +1,284 @@
...
+			log.Error(nil, "unknown password format", "username", un)
+			continue
+		}
+
+		// Build and append full ACL string
+		aclFileContents += fmt.Sprintf("user %s on %s %s\n", un, hashedPass, acl.Permissions)
+	}
+
+	return []byte(aclFileContents)
+}
```
jdheyburn, line 284: The acl set user spec mentioned the hashed password has to be converted to hexidecimal. Is that being done here?

```
#<hashedpassword>: Adds the specified hashed password to the list of user passwords. A hashed password is hashed with SHA256 and translated into a hexadecimal string. Example: #c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2.
```

https://valkey.io/commands/acl-setuser/

internal/controller/users.go
```diff
@@ -0,0 +1,284 @@
...
+
+		// We should have a username, cleartext, or hashed password, and acl
+
+		var hashedPass string
+		pw = strings.TrimRight(pw, "\n")
+		pwLen := len(pw)
+
+		if pwLen == hashLength {
+			// If the password string is 64 characters, assume it is hashed, and prefix it
+			hashedPass = "#" + pw
```
jdheyburn, line 255: This assumption could cause confusion to the user when their 64 char password does not work as intended.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,46 @@ type ValkeyClusterSpec struct {
...
+type AuthSpec struct {
+
+	// Users ACL Secret
+	// A reference name to a Secret containing raw user:permission entries, which will be processed first
+	// Defaults to clusterName-secret
+	// +optional
+	UsersSecretRef string `json:"usersSecretRef,omitempty"`
+
+	// Array of users with raw ACL, or reference to a key in the UsersSecretRef
+	Users map[string]UserAcl `json:"users,omitempty"`
```
jdheyburn, line 99: I suggest we use `listType=map` and `listMapKey=name` here to follow API design guidelines.

https://kubernetes.io/docs/reference/using-api/server-side-apply/

internal/controller/users.go
utdrmac: I was following the naming pattern used by other database operators I've been studying. This is just the default name, and users can override if they want in the CR.

internal/controller/users.go
utdrmac: Sure, I can change the acronym

api/v1alpha1/valkeycluster_types.go
utdrmac: IMO, that's an advanced operation outside the scope of the operator. If the community consensus is to have such functionality, there's no argument from me, and I'll implement it. Or we can keep things simple for now, and add rotation later on.

api/v1alpha1/valkeycluster_types.go
utdrmac: We could, but I didn't want the operator to have to be responsible for validating ACL commands, and syntax. That seemed an overly complicated job for the operator.

config/samples/v1alpha1_valkeycluster.yaml
utdrmac: Wow, so much for e2e tests... Shouldn't this have been caught by the K8S api when the e2e test attempted to apply an invalid CR? (Will remove this anyways)

internal/controller/users.go
utdrmac: Sure

internal/controller/users.go
utdrmac: Yes. It's on users.go:266 `// If password string is less than 64 characters, assume plaintext password, and hash it`

internal/controller/users.go
utdrmac: True.

api/v1alpha1/valkeycluster_types.go
utdrmac:
```
valkeycluster_types.go:101:2: must apply listType to an array, found object
valkeycluster_types.go:101:2: must apply listMapKey to an array, found object
```

Issue #53, opened 2026-01-12
(feat) Support StatefulSet workload for ValkeyCluster


sandeepkunusoth:
Today the operator creates a Deployment for ValkeyCluster pods (see internal/controller/deployment.go). Add support to create ValkeyCluster pods via a StatefulSet instead.

Proposed Spec (initial version):
```
spec:
 workload:
   kind: StatefulSet   # default: Deployment
```

Acceptance Criteria:
- [ ] If spec.workload.kind: StatefulSet is set, the operator creates and reconciles a StatefulSet and other dependencies accordingly for the ValkeyCluster pods.
- [ ] Default behavior remains unchanged (operator continues creating a Deployment when the field is not set).

Contribution:
- [X] I’m willing to implement this feature.

jdheyburn:
Closing as done

daanvinken:
Hey, apologies for resurrecting a closed issue. I'm a bit late to these design discussions and feel slightly conflicted about bringing this up after the fact, but I think it's worth asking.

Was using bare Pods (operator-managed, no intermediate controller) ever considered as an alternative to both StatefulSet and Deployment?

The current architecture has 4 layers: `ValkeyCluster → ValkeyNode → StatefulSet/Deployment → Pod`. The issue I see is that pod-level events (OOMKilled, ImagePullBackOff, CrashLoopBackOff, probe failures, evictions) live on the Pod object, but they don't surface through the layers above. The `updateStatus` in the ValkeyNode controller checks `pod.Status.Conditions` for readiness and then reduces everything to a single condition:

```
Message: "Pod is not ready"
```

So the debugging experience becomes:
```
kubectl get valkeycluster      →  "not ready"
kubectl get valkeynode         →  "Pod is not ready" - but why?
kubectl get statefulset        →  replicas 0/1 - but why?
kubectl describe pod <name>    →  Events: OOMKilled  ← finally, the actual reason
```

That's 4 hops to find root cause. And the StatefulSet/Deployment layer in the middle doesn't contribute much from what I see so far. The ValkeyNode controller already:
- Discovers pods directly via label selectors ([`getPod`](https://github.com/valkey-io/valkey-operator/blob/main/internal/controller/valkeynode_controller.go#L361))
- Checks pod readiness directly from `pod.Status.Conditions` ([`updateStatus`](https://github.com/valkey-io/valkey-operator/blob/main/internal/controller/valkeynode_controller.go#L214))
[...]

PR #57, opened 2026-01-17
(feat) Support StatefulSet workload for ValkeyCluster - initial version

sandeepkunusoth:
This PR fixes https://github.com/valkey-io/valkey-operator/issues/53 by adding optional StatefulSet support for ValkeyCluster pods, while keeping the existing Deployment behavior as the default. Updated e2e tests.


api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,30 @@ type ValkeyClusterSpec struct {
...
+)
+
+// WorkloadSpec defines the workload configuration.
+type WorkloadSpec struct {
+	// Kind specifies the type of workload (Deployment or StatefulSet).
+	// +kubebuilder:validation:Enum=Deployment;StatefulSet
+	// +kubebuilder:default=Deployment
+	// +optional
+	Kind WorkloadKind `json:"kind,omitempty"`
 }
```
jdheyburn, line 100: Do you have any other ideas of any other fields we would put into WorkloadSpec? I am trying to think about how it might best be categorised.

internal/controller/valkeycluster_controller.go
```diff
@@ -301,6 +310,36 @@ func (r *ValkeyClusterReconciler) upsertDeployments(ctx context.Context, cluster
...
+	if err := r.List(ctx, existing, client.InNamespace(cluster.Namespace), client.MatchingLabels(labels(cluster))); err != nil {
+		log.Error(err, "failed to list StatefulSets")
+		return err
+	}
+
+	desired := int(cluster.Spec.Shards * (1 + cluster.Spec.Replicas))
+
+	// Create missing statefulsets
+	for i := len(existing.Items); i < desired; i++ {
+		sts := createStatefulSet(cluster)
```
jdheyburn, line 327: Am I understanding this right - does it loop over `desired` number of times, creating a StatefulSet with the same name?

Where does it create a singleton StatefulSet for the shard master, and shard replica?

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: Hi, maybe we can add updateStrategy, podmanagementpolicy under workload which manages configuration of either deployment or statefulset? please let me know if u have any suggestion.
```
workload:
  kind: 
  podManagementPolicy:
  updateStrategy:
```

internal/controller/valkeycluster_controller.go
sandeepkunusoth: this logic is similar to how we are creating deployments, except it creates statefulsets here. for naming it adds extra suffixes as i am using `GenerateName: cluster.Name + "-",`

Should i change this to add some ordering here instead of random suffixes.

```
k get statefulset
NAME                           READY   AGE
valkeycluster-sts-test-2xsjq   1/1     10s
valkeycluster-sts-test-5xgr7   1/1     10s
valkeycluster-sts-test-8n9pl   1/1     10s
valkeycluster-sts-test-lgwvk   1/1     10s
valkeycluster-sts-test-p8942   1/1     10s
valkeycluster-sts-test-vmnt4   1/1     10s
```

test/e2e/e2e_test.go
```diff
@@ -44,41 +44,17 @@ var _ = Describe("Manager", Ordered, func() {
 	AfterEach(func() {
 		specReport := CurrentSpecReport()
 		if specReport.Failed() {
-			By("Fetching controller manager pod logs")
```
bjosv, line 47: This code was generated by kubebuilders scaffolding, maybe we should make this debug collection change upstream instead? They seemed to respond quickly last time.

Issue #55, opened 2026-01-15
[feat] Support scaling


ysqyang:
The current implementation does not support horizontal scaling (scale-out or scale-in).

When increasing the number of shards, new pods are created, but there is no slot migration or rebalancing logic to redistribute slots from existing shards to the new ones. As a result, the new shards remain unused.

Similarly, decreasing the number of shards is currently a no-op, with no data or slot reassignment performed.

jdheyburn:
Closing as done

PR #60, opened 2026-01-20
add scale-out/resharding logic

ysqyang:
## Summary
This PR adds incremental **slot rebalancing for scale‑out** in the Valkey operator (https://github.com/valkey-io/valkey-operator/issues/55 for more context). The controller now migrates slots from existing primaries to newly added primaries in small batches during reconcile, enabling shard scale‑out without manual resharding.

Key behavior:
- New nodes are joined via `CLUSTER MEET`, then **slot migration** begins once all pending nodes are joined.
- Rebalancing is **deterministic** and **incremental** (small batch per reconcile).
- Slot migration uses standard `CLUSTER SETSLOT` + `GETKEYSINSLOT` + `MIGRATE` flow.

## Details
- `BuildRebalanceMove` computes a small, deterministic `SlotMove` (src, dst, slots) based on target slot counts for the expected shard count.
- Rebalancing migrates up to `rebalanceSlotBatchSize` slots per reconcile and migrates keys in batches.
- Status conditions are updated to reflect ongoing rebalancing.

## Testing
- Unit tests updated/added:
  - `internal/valkey/rebalance_test.go` (existing tests + scale‑out scenarios)
- Manual validation:
  - Before (3 shards):
    - Screenshot 2026-01-20 at 12 23 07 PM
  - After (5 shards):
    - Screenshot 2026-01-20 at 12 43 27 PM
  - It can be seen that each master now owns 3276-3277 slots after the resharding.

### DIFF
--- a/internal/controller/rebalance.go
+++ b/internal/controller/rebalance.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	rebalanceSlotBatchSize    = 20
+	rebalanceSlotBatchSize    = 200
 	rebalanceKeysBatchSize    = 100
 	rebalanceMigrateTimeoutMs = 5000
 )
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"embed"
 	"errors"
+	"fmt"
 	"slices"
 	"strconv"
+	"strings"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -182,6 +184,33 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// --- Reclassify replica-intended nodes ---
+	// When all 16384 slots are assigned, GetClusterState categorizes
+	// slot-less masters with cluster_known_nodes > 1 into state.Shards
+	// (each as its own single-node empty shard) rather than PendingNodes.
+	// This is correct for new primaries awaiting rebalance during
+	// scale-out. However, nodes labeled as replicas (valkey.io/role=
+	// replica) still need CLUSTER REPLICATE. Move them back to
+	// PendingNodes so Phase 3 can process them, and drop their
+	// placeholder shards.
+	{
+		var keptShards []*valkey.ShardState
+		for _, shard := range state.Shards {
+			if countSlots(shard.Slots) > 0 || len(shard.Nodes) != 1 {
+				keptShards = append(keptShards, shard)
+				continue
+			}
+			node := shard.Nodes[0]
+			role, _ := r.podRoleAndShard(node.Address, pods)
+			if node.IsPrimary() && role == RoleReplica {
+				state.PendingNodes = append(state.PendingNodes, node)
+			} else {
+				keptShards = append(keptShards, shard)
+			}
+		}
+		state.Shards = keptShards
+	}
+
 	// --- Phase 3: REPLICATE all replica-labeled pending nodes ---
 	// By this point all primaries have slots and appear in state.Shards.
 	// CLUSTER REPLICATE for different replicas targets different primaries,
@@ -213,21 +242,11 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 
-	replicaAttached, err := r.attachReplica(ctx, cluster, state)
-	if err != nil {
-		log.Error(err, "unable to add cluster replica")
-		r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "ReplicaCreationFailed", "CreateReplica", "Failed to create replica: %v", err)
-		setCondition(cluster, valkeyiov1alpha1.ConditionDegraded, valkeyiov1alpha1.ReasonNodeAddFailed, err.Error(), metav1.ConditionTrue)
-		_ = r.updateStatus(ctx, cluster, state)
-		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
-	}
-	if replicaAttached {
-		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonMissingReplicas, "Waiting for all replicas to be created", metav1.ConditionFalse)
-		setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonReconciling, "Creating replicas", metav1.ConditionTrue)
-		setCondition(cluster, valkeyiov1alpha1.ConditionClusterFormed, valkeyiov1alpha1.ReasonMissingReplicas, "Waiting for replicas", metav1.ConditionFalse)
-		_ = r.updateStatus(ctx, cluster, state)
-		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
-	}
+	// NOTE: Replica attachment is handled entirely by Phase 3
+	// (replicatePendingReplicas) which uses pod labels to ensure each
+	// replica attaches to the correct shard primary. The old
+	// label-unaware attachReplica function was removed because it could
+	// race with Phase 3 and re-attach replicas to wrong primaries.
 
 	for _, shard := range state.Shards {
 		if countSlots(shard.Slots) == 0 {
@@ -269,6 +288,10 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 	if rebalanced {
+		// Clear any stale Degraded condition from earlier phases (e.g.
+		// NodeAddFailed set when replicas couldn't attach to primaries
+		// that hadn't received slots yet during scale-out).
+		meta.RemoveStatusCondition(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionDegraded)
 		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonReconciling, "Cluster is Reconciling", metav1.ConditionFalse)
 		setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonRebalancingSlots, "Rebalancing slots across primaries", metav1.ConditionTrue)
 		_ = r.updateStatus(ctx, cluster, state)
@@ -622,9 +645,17 @@ func (r *ValkeyClusterReconciler) assignSlotsToPendingPrimaries(ctx context.Cont
 	return assigned, nil
 }
 
+// errPrimaryNotReady is returned by replicateToShardPrimary when the
+// primary for a shard hasn't received slots yet (e.g. during scale-out,
+// while the rebalancer is still migrating slots). This is not a fatal
+// error — the replica will be retried on a future reconcile.
+var errPrimaryNotReady = errors.New("primary not yet in cluster state (awaiting rebalance)")
+
 // replicatePendingReplicas issues CLUSTER REPLICATE for all pending nodes
-// labeled as replicas. By the time this runs, all primaries should already
-// have slots and appear in state.Shards.
+// labeled as replicas. During initial creation all primaries should already
+// have slots, but during scale-out new primaries may still be empty while the
+// rebalancer migrates slots to them. Replicas whose primary isn't ready yet
+// are silently skipped and retried on the next reconcile.
 // Returns the number of replicas that were attached.
 func (r *ValkeyClusterReconciler) replicatePendingReplicas(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, state *valkey.ClusterState, pods *corev1.PodList) (int, error) {
 	log := logf.FromContext(ctx)
@@ -637,6 +668,10 @@ func (r *ValkeyClusterReconciler) replicatePendingReplicas(ctx context.Context, 
 		}
 
 		if err := r.replicateToShardPrimary(ctx, cluster, state, node, shardIndex, pods); err != nil {
+			if errors.Is(err, errPrimaryNotReady) {
+				log.V(1).Info("skipping replica; primary not ready yet", "node", node.Address, "shard", shardIndex)
+				continue
+			}
 			log.Error(err, "failed to replicate pending node", "node", node.Address, "shard", shardIndex)
 			return replicated, err
 		}
@@ -702,11 +737,18 @@ func (r *ValkeyClusterReconciler) replicateToShardPrimary(ctx context.Context, c
 		}
 	}
 	if primaryNodeId == "" {
-		return errors.New("primary Valkey node not found in cluster state for shard " + strconv.Itoa(shardIndex))
+		return fmt.Errorf("shard %d: %w", shardIndex, errPrimaryNotReady)
 	}
 
 	log.V(1).Info("add a new replica", "primary IP", primaryIP, "primary Id", primaryNodeId, "replica address", node.Address, "shardIndex", shardIndex)
 	if err := node.Client.Do(ctx, node.Client.B().ClusterReplicate().NodeId(primaryNodeId).Build()).Error(); err != nil {
+		// "Unknown node" means gossip hasn't propagated the primary's ID to
+		// this replica yet. This is transient and will resolve on the next
+		// reconcile once gossip catches up — treat it as retriable.
+		if strings.Contains(err.Error(), "Unknown node") {
+			log.V(1).Info("replica does not yet know primary (gossip pending); will retry", "replica", node.Address, "primaryId", primaryNodeId)
+			return fmt.Errorf("shard %d: %w", shardIndex, errPrimaryNotReady)
+		}
 		log.Error(err, "command failed: CLUSTER REPLICATE", "nodeId", primaryNodeId)
 		r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "ReplicaCreationFailed", "CreateReplica", "Failed to create replica: %v", err)
 		return err


internal/controller/rebalance.go
```diff
@@ -0,0 +1,94 @@
...
+	rebalanceKeysBatchSize    = 100
+	rebalanceMigrateTimeoutMs = 5000
+)
+
+func (r *ValkeyClusterReconciler) rebalanceSlots(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, state *valkey.ClusterState) (bool, error) {
+	if state == nil {
+		return false, nil
+	}
+
+	move, err := valkey.BuildRebalanceMove(state, int(cluster.Spec.Shards), rebalanceSlotBatchSize)
```
sandeepkunusoth, line 43: can u add e2e tests to test rebalancing?

ysqyang: done

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+	for {
+		keys, err := src.Client.Do(ctx, src.Client.B().ClusterGetkeysinslot().Slot(int64(slot)).Count(int64(rebalanceKeysBatchSize)).Build()).AsStrSlice()
+		if err != nil {
+			return fmt.Errorf("getkeysinslot failed for slot %d on %s: %w", slot, src.Address, err)
+		}
+		if len(keys) == 0 {
+			break
+		}
+		for _, key := range keys {
+			if err := src.Client.Do(ctx, src.Client.B().Migrate().Host(dst.Address).Port(int64(dst.Port)).Key(key).DestinationDb(0).Timeout(rebalanceMigrateTimeoutMs).Replace().Build()).Error(); err != nil {
```
xtllli, line 195: `MIGRATE` allows multiple keys in one command to speed things up.

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+		if !isSlotMigrationTerminal(state) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func isSlotMigrationTerminal(state string) bool {
+	switch strings.ToLower(state) {
+	case "success", "failed", "canceled", "cancelled":
```
jdheyburn, line 129: Would we expect to see both `canceled` and `cancelled`, or just one of them?

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+			dst.Id,
+		)
+	}
+	if err := src.Client.Do(ctx, cmd.Build()).Error(); err != nil {
+		return fmt.Errorf("migrateslots failed from %s to %s: %w", src.Address, dst.Address, err)
+	}
+	return nil
+}
+
+func migrateSlotLegacy(ctx context.Context, src *valkey.NodeState, dst *valkey.NodeState, slot int) error {
```
jdheyburn, line 174: There's a few client commands in this function which are being chained. Are we able to place them behind a human friendly function name?

e.g. 

```golang
if err := src.Client.Do(ctx, src.Client.B().ClusterSetslot().Slot(int64(slot)).Migrating().NodeId(dst.Id).Build()).Error(); err != nil {

# to
if err := setClusterSlot(src, dst, slot); err != nil {
```

internal/controller/rebalance.go
jdheyburn: @ysqyang Is this comment still applicable?

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+	log.V(1).Info("rebalancing slots", "src", move.Src.Address, "dst", move.Dst.Address, "slots", len(move.Slots))
+	r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "SlotsRebalancing", "RebalanceSlots", "Moving %d slots from %s to %s", len(move.Slots), move.Src.Address, move.Dst.Address)
+
+	ranges := slotsToRanges(move.Slots)
+	if err := migrateSlotsAtomic(ctx, move.Src, move.Dst, ranges); err != nil {
+		if isSlotsNotServedByNode(err) {
+			log.V(1).Info("slots no longer served by source; will retry with fresh state", "src", move.Src.Address, "dst", move.Dst.Address)
+			return true, nil
+		}
+		if !isAtomicMigrationUnsupported(err) {
```
jdheyburn, line 80: Could we switch the function call to be `isAtomicMigrationSupported`? The "is not unsuppported" is hard to parse.

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "unknown command") ||
+		strings.Contains(msg, "unknown subcommand") ||
+		strings.Contains(msg, "syntax error") ||
+		strings.Contains(msg, "wrong number of arguments") ||
+		strings.Contains(msg, "not supported")
+}
+
+func isSlotsNotServedByNode(err error) bool {
```
jdheyburn, line 244: Similarly to `isAtomicMigrationUnsupported`, switching the boolean of this function can improve readability

`isSlotsServedByNode`

internal/controller/rebalance.go
ysqyang: Actually there's a reason I went with "unsupported" here. This is a helper function to check if an error is due to "atomic migration unsupported", as you can see from function argument.

internal/controller/rebalance.go
ysqyang: Similar to my comment above, this is a helper function that basically checks "is this error due to 'slots not server by node?'", hence the negation here.

internal/controller/rebalance.go
ysqyang: @jdheyburn This is for legacy slot migration which is not applicable to valkey >= 9.0, so I think we can leave this as is for now.

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+			return true, nil
+		}
+		state := strings.ToLower(values["state"])
+		if state == "" {
+			log.V(1).Info("slot migration entry missing state; treating as in progress", "src", src.Address)
+			return true, nil
+		}
+		if !isSlotMigrationTerminal(state) {
+			return true, nil
+		}
```
sandeepkunusoth, line 122: can we add some comments wherever required. i am not very familiar with valkey cluster rebalancing. it will be helpful for new contributors too.

internal/controller/rebalance.go
ysqyang: This is just to be defensive as both spellings are valid.

internal/controller/valkeycluster_controller.go
```diff
@@ -360,7 +400,7 @@ func (r *ValkeyClusterReconciler) addValkeyNode(ctx context.Context, cluster *va
 		slotEnd := slotStart + (16384 / shardsRequired) - 1
```
sandeepkunusoth: nit: can we reuse total slots variable here which is already defined in rebalance.go

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+func nodeKnownToSource(ctx context.Context, src *valkey.NodeState, dst *valkey.NodeState) (bool, error) {
+	if src == nil || dst == nil || src.Client == nil {
+		return false, fmt.Errorf("missing source or destination node")
+	}
+	if dst.Id == "" {
+		return false, nil
+	}
+	cmd := src.Client.B().ClusterNodes().Build()
+	nodes, err := src.Client.Do(ctx, cmd).ToString()
+	if err != nil {
```
sandeepkunusoth, line 145: NodeState already has clusterNodes information? can we use that directly? also maybe this function can be moved to clusterstate?

```
// NodeState represents the current state of an inspected cluster node.
type NodeState struct {
	Client       vclient.Client
	Address      string
	Port         int
	Id           string
	Flags        []string
	ShardId      string
	Info         map[string]string
	ClusterInfo  map[string]string
	ClusterNodes string
}
```

internal/controller/rebalance.go
jdheyburn: I wonder then if there is a more proactive check we can make to determine if the cluster supports atomic slot migration, so that we're not reactively inspecting the error to see if we should do legacy. That might help with readability too.

```
if supportsAtomicMigration(cluster):
  doAtomicMigration(cluster)
else:
  migrateLegacy(cluster)
```

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+		if !isAtomicMigrationUnsupported(err) {
+			return false, err
+		}
+		log.V(1).Info("atomic slot migration unsupported; falling back to legacy migration", "src", move.Src.Address, "dst", move.Dst.Address)
+		for _, slot := range move.Slots {
+			if err := migrateSlotLegacy(ctx, move.Src, move.Dst, slot); err != nil {
+				return false, err
+			}
+		}
```
jdheyburn, line 88: What happens if a cluster upgrade is being performed from 8.X -> 9.X? Which slot migration would be used against a cluster that is transitioning to a version that supports atomic slot migration?

internal/controller/rebalance.go
jdheyburn: If its not applicable to valkey >= 9.0 then perhaps we can remove it and add it in a future iteration?

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+	nodes, err := src.Client.Do(ctx, cmd).ToString()
+	if err != nil {
+		return false, fmt.Errorf("cluster nodes failed on %s: %w", src.Address, err)
+	}
+	return strings.Contains(nodes, dst.Id), nil
+}
+
+func migrateSlotsAtomic(ctx context.Context, src *valkey.NodeState, dst *valkey.NodeState, ranges []valkey.SlotsRange) error {
+	if src == nil || dst == nil {
+		return fmt.Errorf("missing source or destination node")
```
sandeepkunusoth, line 153: is src.Client == nil issing here? i see it other places but not here?

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	return nil
+}
+
+func slotsToRanges(slots []int) []valkey.SlotsRange {
+	if len(slots) == 0 {
+		return nil
+	}
+	ordered := append([]int(nil), slots...)
+	sort.Ints(ordered)
```
sandeepkunusoth, line 215: do we need creation of copy here cant we do in place if original can be modfied?

internal/controller/rebalance.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	if err := src.Client.Do(ctx, src.Client.B().ClusterSetslot().Slot(int64(slot)).Migrating().NodeId(dst.Id).Build()).Error(); err != nil {
+		return fmt.Errorf("setslot migrating failed for slot %d on %s: %w", slot, src.Address, err)
+	}
+	if err := dst.Client.Do(ctx, dst.Client.B().ClusterSetslot().Slot(int64(slot)).Importing().NodeId(src.Id).Build()).Error(); err != nil {
+		return fmt.Errorf("setslot importing failed for slot %d on %s: %w", slot, dst.Address, err)
+	}
+
+	for {
```
sandeepkunusoth, line 186: should we add any timeout for this infinite loop?

internal/controller/replicas.go
```diff
@@ -0,0 +1,153 @@
...
+				return true, nil
+			}
+			log.V(1).Info("add a new replica", "primary address", target.Address, "primary Id", target.Id, "replica address", candidate.Address)
+			if err := candidate.Client.Do(ctx, candidate.Client.B().ClusterReplicate().NodeId(target.Id).Build()).Error(); err != nil {
+				log.Error(err, "command failed: CLUSTER REPLICATE", "nodeId", target.Id)
+				r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "ReplicaCreationFailed", "CreateReplica", "Failed to create replica: %v", err)
+				return false, err
+			}
+			r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "ReplicaCreated", "CreateReplica", "Created replica for primary %v", target.Id)
+			return true, nil
```
sandeepkunusoth, line 121: are we adding only 1 replica per reconciliation here even if multiple candidates can be available?  Its just that there are too many nested loops for each reconciliation.  maybe we can optimize this to add multiple replicas per reconciliation.

internal/controller/rebalance.go
ysqyang: Good question. Right now our controller doesn't support rolling upgrades (something we should create an issue for). The rolling upgrade should be implemented to ensure that resharding only happens after all pods are rolled.

internal/controller/rebalance.go
ysqyang: It isn't necessary today, but it's cheap insurance (a single allocation of an array of `len(slots)` which is at most 20 * 8 = 160 bytes) against a future refactor where `move.Slots` gets reused or inspected after the call.

internal/controller/rebalance.go
ysqyang: The loop terminates when `CLUSTER GETKEYSINSLOT` returns zero keys, meaning the slot is empty on the source. So as long as the number of keys is finite, the loop breaks eventually. We can add a timeout here, but since this is only the legacy fallback path (for valkey < 9.0), it might not be worth the extra complexity. For valkey >= 9.0, the atomic migration path doesn't block.

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
 			}
 			Eventually(verifyClusterAccess).Should(Succeed())
 		})
+
+	It("rebalances slots on scale out", func() {
+		const baseShards = 2
+		const scaleOutShards = 3
+		valkeyClusterName = "valkeycluster-scaleout"
+
+		By("ensuring the controller pod name is set")
```
bjosv, line 280: The previous test-steps doesn't check if the operator is running since it's expected to be started in the testsuite-initiation. Is it needed here, or can we remove it for less code maintenance?

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+		By("creating a smaller ValkeyCluster for scale-out")
+		scaleOutManifest := fmt.Sprintf(`apiVersion: valkey.io/v1alpha1
+kind: ValkeyCluster
+metadata:
+  name: %s
+spec:
+  shards: %d
+  replicas: 1
+  resources:
+    requests:
```
bjosv, line 303: Do we need the resource requests/limits?

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+    limits:
+      memory: "512Mi"
+      cpu: "500m"
+`, valkeyClusterName, baseShards)
+		manifestFile := filepath.Join(os.TempDir(), "valkeycluster-scaleout.yaml")
+		err = os.WriteFile(manifestFile, []byte(scaleOutManifest), 0644)
+		Expect(err).NotTo(HaveOccurred(), "Failed to write scale-out manifest file")
+		defer os.Remove(manifestFile)
+
+		cmd = exec.Command("kubectl", "delete", "valkeycluster", valkeyClusterName, "--ignore-not-found=true")
```
bjosv, line 315: Are we deleting an old run here? I though we deleted it at the end of then testcase?

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+		Expect(err).NotTo(HaveOccurred(), "Failed to apply scale-out ValkeyCluster CR")
+
+		By("waiting for the cluster to be ready before scaling")
+		verifyReadyForScaleOut := func(g Gomega) {
+			cr, err := utils.GetValkeyClusterStatus(valkeyClusterName)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+			g.Expect(cr.Status.ReadyShards).To(Equal(int32(baseShards)))
+		}
+		Eventually(verifyReadyForScaleOut, 10*time.Minute, 2*time.Second).Should(Succeed())
```
bjosv, line 328: Other tests only waits the default time, i.e. 2 minutes. Is set to 10 for a reason?

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+					continue
+				}
+				if !strings.Contains(fields[2], "master") {
+					continue
+				}
+				masterWithSlots++
+			}
+			g.Expect(masterWithSlots).To(Equal(scaleOutShards), "Expected all primaries to own slots after rebalance")
+		}
+		Eventually(verifySlotRebalance, 10*time.Minute, 2*time.Second).Should(Succeed())
```
bjosv, line 364: I guess the extra 2sec here is not needed? Can be removed, same below.

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+
+		By(fmt.Sprintf("waiting for the cluster to report %d ready shards", scaleOutShards))
+		verifyScaledOut := func(g Gomega) {
+			cr, err := utils.GetValkeyClusterStatus(valkeyClusterName)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(cr.Status.Shards).To(Equal(int32(scaleOutShards)))
+			g.Expect(cr.Status.ReadyShards).To(Equal(int32(scaleOutShards)))
+			g.Expect(cr.Status.State).To(Or(
+				Equal(valkeyiov1alpha1.ClusterStateReady),
```
bjosv, line 373: I expected the state to only be in ClusterStateReady before passing this test? Maybe I have missed something

test/e2e/valkeycluster_test.go
```diff
@@ -269,6 +271,115 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+				Equal(valkeyiov1alpha1.ClusterStateReconciling),
+			))
+		}
+		Eventually(verifyScaledOut, 10*time.Minute, 2*time.Second).Should(Succeed())
+
+		By("cleaning up the scale-out cluster")
+		cmd = exec.Command("kubectl", "delete", "valkeycluster", valkeyClusterName, "--wait=false")
+		_, _ = utils.Run(cmd)
+	})
```
bjosv, line 382: Hum, is there a weird scoping here since we have two `})` with the same indentation?

api/v1alpha1/valkeycluster_types.go
```diff
@@ -85,6 +85,7 @@ type ExporterSpec struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
 	// Enable or disable the exporter sidecar container
+	// +kubebuilder:default=true
 	Enabled bool `json:"enabled,omitempty"`
```
sandeepkunusoth, line 89: nit: this enabled is not needed, it will not have any effect as we already have this initial state set https://github.com/valkey-io/valkey-operator/blob/9a724fb61e4af19254f88d4165795eec82a2d3b6/api/v1alpha1/valkeycluster_types.go#L73

internal/controller/deployment.go
```diff
@@ -116,22 +118,53 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
...
+// Both deployLabels and podLabels are built from fresh copies of labels()
+// (which clones the user-defined labels map) so mutations here don't leak.
+func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex int, role string) *appsv1.Deployment {
 	containers := generateContainersDef(cluster)
+
+	// Deployment-level labels carry shard/role for easy listing/filtering.
+	deployLabels := labels(cluster)
+	deployLabels[LabelShardIndex] = strconv.Itoa(shardIndex)
+	deployLabels[LabelRole] = role
+
```
sandeepkunusoth, line 147: was this labels change required as part of this PR? this PR changes became even more after these changes in last commit.

ysqyang: This change is actually part of https://github.com/valkey-io/valkey-operator/pull/81 and will be gone once it's merged to `main`.

PR #63, opened 2026-01-23
(rfc): Design Proposal for additional CRDs

jdheyburn:
Would love to get as many opinions on this.

Please leave comments on the PR where appropriate.

I don't anticipate merging this in its current form; rather when we have agreed on a design for each CRD they can be merged in individually (or discussed on Discussions).

There are still some CRD design guidelines that I need to enforce:

- [x] Nested sub-problems
- [x] Avoid bool fields
- [x] Use +listType=map
- [ ] Conditions use metav1.Condition
- [ ] observedGeneration
- [ ] Enum extensibility documented
- [ ] No cross-namespace refs by default
- [ ] Sentence test for naming
- [ ] Status uses adjectives/past-tense
- [ ] Glossary defined

The guidelines:

- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

Here's a mock diagram of the CRD relationships. ValkeyCluster is similar, but it only provisions ValkeyNode.

```text

⏺ ┌─────────────────────────────────────────────────────────────────────────────┐
  │                              ValkeyPool                                     │
  │                              (cache)                                        │
  │  ┌─────────────────────────────────────────────────────────────────────┐   │
  │  │ spec:                                                                │   │
  │  │   shards: 4                                                          │   │
  │  │   replicasPerShard: 2                                                │   │
  │  │   sentinel:                                                          │   │
  │  │     enabled: true ─────────────────────────────────┐                 │   │
  │  │   template: ...                                    │                 │   │
  │  └────────────────────────────────────────────────────│─────────────────┘   │
  └───────────────────────────────────────────────────────│─────────────────────┘
                            │                             │
                            │ creates                     │ creates
                            ▼                             ▼
  ┌─────────────────────────────────────────┐   ┌─────────────────────────┐
  │              Valkey (cache-0)           │   │    ValkeySentinel       │
  │  ┌───────────────────────────────────┐  │   │    (cache-sentinel)     │
  │  │ spec:                             │  │   │  ┌───────────────────┐  │
  │  │   replicas: 2                     │  │   │  │ spec:             │  │
  │  │   failover:                       │  │   │  │   replicas: 3     │  │
  │  │     mode: sentinel                │  │   │  └───────────────────┘  │
  │  │     sentinel:                     │  │   │                         │
  │  │       ref: ───────────────────────│──│───│───▶ (referenced)        │
  │  │         name: cache-sentinel      │  │   │                         │
  │  │       config:                     │  │   │  Creates 3 Sentinel     │
  │  │         quorum: 2                 │  │   │  pods that monitor      │
  │  │         downAfterMillis: 30000    │  │   │  all Valkey instances   │
  │  └───────────────────────────────────┘  │   │  referencing it         │
  │                   │                      │   └─────────────────────────┘
  │                   │ creates              │               ▲
  │                   ▼                      │               │
  │  ┌───────────────────────────────────┐  │               │
  │  │ ValkeyNode (cache-0-primary)      │  │               │
  │  │ ValkeyNode (cache-0-replica-0)    │  │               │
  │  │ ValkeyNode (cache-0-replica-1)    │  │               │
  │  └───────────────────────────────────┘  │               │
  └─────────────────────────────────────────┘               │
                                                            │
  ┌─────────────────────────────────────────┐               │
  │              Valkey (cache-1)           │               │
  │  ┌───────────────────────────────────┐  │               │
  │  │ spec.failover.sentinel.ref: ──────│──│───────────────┘
  │  │   name: cache-sentinel            │  │
  │  └───────────────────────────────────┘  │
  │                   │                      │
  │                   ▼                      │
  │  ┌───────────────────────────────────┐  │
  │  │ ValkeyNode (cache-1-primary)      │  │
  │  │ ValkeyNode (cache-1-replica-0)    │  │
  │  │ ValkeyNode (cache-1-replica-1)    │  │
  │  └───────────────────────────────────┘  │
  └─────────────────────────────────────────┘

    ... (cache-2, cache-3 similar)
```

### DIFF
--- a/docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
+++ b/docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
@@ -1,0 +1,1681 @@
+> Represents the CRD design I submitted for review in this PR: TODO
+
+
+# Valkey Operator CRD Architecture - Design Proposal
+
+**Date:** 2026-01-22
+**Status:** Draft
+**Authors:** jdheyburn
+
+---
+
+## Table of contents
+
+- [Overview](#overview)
+- [Design Goals and Principles](#design-goals-and-principles)
+- [CRD Architecture Overview](#crd-architecture-overview)
+- [ValkeyNode CRD (Internal)](#valkeynode-crd-internal)
+- [Valkey CRD (Standalone/Replicated)](#valkey-crd-standalonereplicated)
+- [ValkeyPool CRD (Client-Side Sharding)](#valkeypool-crd-client-side-sharding)
+- [ValkeyCluster CRD (Server-Side Sharding)](#valkeycluster-crd-server-side-sharding)
+- [Sentinel CRD (HA Failover)](#sentinel-crd-ha-failover)
+- [Common Configuration Patterns](#common-configuration-patterns)
+- [Key Design Decisions](#key-design-decisions)
+- [Design Rationale and Rejected Alternatives](#design-rationale-and-rejected-alternatives)
+- [Migration Guide](#migration-guide)
+- [Future Enhancements](#future-enhancements)
+- [References](#references)
+- [Appendices](#appendices)
+- [Questions](#questions)
+
+---
+
+## Overview
+
+This document aims to outline the design of CRDs that would be required for the valkey-operator. For the 0.1.0 release we have suggested a minimal ValkeyCluster that is open to breaking changes. This document proposes additional CRDs to meet the requirements outlined in the references.
+
+### Valkey deployment methods
+
+Valkey is popular because of the breadth and depth of the data types and workloads it support, which then open it up to several methods of deployment.
+
+- Standalone
+    - no replication, just one primary instance
+- HA replication,
+    - one primary and one or more replicas
+- Sentinel-backed HA replication
+    - one primary and one or more replicas
+    - a separate process (Valkey Sentinel) that has watchdog, healthchecker, and primary discovery capabilities
+- Cluster
+    - Server-side sharding
+    - Horizontally scalable
+    - Independent of above deployment methods
+
+### Five-CRD architecture
+
+| CRD | Purpose | User-Facing |
+|-----|---------|-------------|
+| **ValkeyNode** | Single-pod abstraction | ❌ No (operator-managed) |
+| **Valkey** | Standalone/replicated instance | ✅ Yes |
+| **ValkeyPool** | Multiple independent instances | ✅ Yes |
+| **ValkeyCluster** | Sharded cluster mode | ✅ Yes |
+| **Sentinel** | HA monitoring and failover | ✅ Yes |
+
+## Design Goals and Principles
+
+### Primary Goals
+
+1. **User-first approach**: Deploy functional, highly available Valkey clusters with minimal configuration
+2. **Kubernetes-native**: Follow established Kubernetes conventions for consistency
+3. **Production focus**: Prioritise cluster and replication modes for production workloads
+4. **Upgrade paths**: Enable seamless progression from standalone → replicated → HA → sharded
+
+### Design principles
+
+TODO comment to ask if there are any others that should be included
+
+- Progressive enhancement
+    - Start simple and upgrade as required
+    - No forced migrations between CRD types
+        - i.e. `ValkeyStandalone` -> `ValkeyReplicated` -> `ValkeySentinel`
+- Explicit over implicit
+    - Failover modes explicitly declared
+    - No external dependencies
+    - Self-documenting specs
+- Separation of concerns
+    - ValkeyNode handles infrastructure
+    - Parent controllers handle topology
+    - Sentinel handles failover
+- Nested sub-problems
+    - Related fields are grouped (`persistence.rdb`, `tls.clusterBus`)
+    - Better organisation and extensibility
+- Configuration-driven behaviour
+    - ValkeyNode behaviour is determined by config fields
+    - Not coupled to parent CRD type
+    - Generic and reusable
+
+## CRD Architecture Overview
+
+### Resource Hierarchy
+
+```
+User Creates:
+├─ ValkeyPool
+│  └─ Creates: Multiple Valkey resources
+│     └─ Creates: ValkeyNode per pod
+│        └─ Creates: Deployment/StatefulSet + Service + PVC
+│
+├─ Valkey
+│  └─ Creates: ValkeyNode per pod
+│     └─ Creates: Deployment/StatefulSet + Service + PVC
+│
+├─ ValkeyCluster
+│  └─ Creates: ValkeyNode per pod
+│     └─ Creates: Deployment/StatefulSet + Service + PVC
+│
+└─ Sentinel
+   └─ Creates: ValkeyNode per Sentinel pod
+      └─ Creates: Deployment/StatefulSet + Service
+```
+
+### Controller Responsibilities
+
+| Controller        | Responsibilities                                                                                                               |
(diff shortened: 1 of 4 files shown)


docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1682 @@
...
+### Primary Goals
+
+1. **User-first approach**: Deploy functional, highly available Valkey clusters with minimal configuration
+2. **Kubernetes-native**: Follow established Kubernetes conventions for consistency
+3. **Production focus**: Prioritise cluster and replication modes for production workloads
+4. **Upgrade paths**: Enable seamless progression from standalone → replicated → HA → sharded
+
+### Design principles
+
+TODO comment to ask if there are any others that should be included
```
jdheyburn, line 71: Anything else I might've missed?

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1682 @@
...
+  # Observed replication state (queried via INFO replication)
+  observedRole: primary  # "primary" or "replica"
+  observedReplicaOf: ""
+
+  ready: true
+```
+
+### Replication and Failover Semantics
+
+TODO these semantics needs to be finalised
```
jdheyburn, line 227: We'll need another RFC to discuss how we want replication/failover/etc, to be managed. I have a great understanding of how we would do this for Sentinel, but I might need some help when it is operator managed.

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1835 @@
...
+    size: 10Gi
+    # +optional
+    storageClassName: gp3
+
+  # Service configuration
+  service:
+    type: ClusterIP
+    annotations: {}
+
+  # Valkey-specific configuration (only when type=valkey)
```
bjosv: Not sure what `type=valkey` means here?

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1835 @@
...
+  valkeyConfig:
+    # Cluster configuration (presence-based)
```
bjosv, line 201: Shouldn't this be used for standalone mode as well as in cluster mode, i.e. to set a config with some options that a user wants?

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1835 @@
...
+    # If set, node runs in cluster mode with assigned slots
+    # Omit entirely for standalone/replicated mode
+    # cluster:
```
bjosv, line 204: Slots are provided to the Valkey Cluster instance when the controller has decided that this specific pod should be a primary, then connects to the valkey process via a client and then sends commands to add slots.
If we want to use configs as an extra step it makes it abit more complex I think. Something needs to give commands to the instance.

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1835 @@
...
+    cluster-node-timeout: "15000"
+    cluster-replica-validity-factor: "10"
+
+status:
+  state: Ready  # Initializing | Reconciling | Ready | Degraded | Suspended | Failed
+  reason: ClusterHealthy
+  message: "All shards healthy, all slots assigned"
+
+  totalShards: 3
+  readyShards: 3
```
bjosv, line 776: Its tough for a cluster-user to get a grip on that we are using `shard` from the Valkey domain, but the `replicas` is from the K8s domain and not Valkey.. the primary-replica concept is so deeply rooted..

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1835 @@
...
+
+  conditions: []
+```
+
+### Naming Convention
+
+**ValkeyNode naming:** `<cluster-name>-<shard-index>-<replica-index>`
+
+Examples:
+- `mycluster-0-0`: Shard 0, replica index 0 (primary)
```
bjosv, line 811: I think we should use a short hash here instead of indexes, at least we shouldn't give an example where  replica index 0 always is a primary. The primary will be moved when there is a failure.

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
```diff
@@ -0,0 +1,1835 @@
...
+Valkey Cluster uses **16384 hash slots** distributed across primaries:
+
+**For 3 shards:**
+- Shard 0: slots 0-5461 (5462 slots)
+- Shard 1: slots 5462-10922 (5461 slots)
+- Shard 2: slots 10923-16383 (5461 slots)
+
+**Formula:**
+```go
+slotsPerShard := 16384 / shards
```
bjosv, line 829: Just a feeling, but maybe could skip code in this doc to tighten it, it's a big document

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
jdheyburn: In either case, the slots allocated to a ValkeyNode would need to be defined somewhere in the spec - then the ValkeyNode controller can issue the commands to the node. Does that sound about right? 

Perhaps this section can be reworked to:


```
# saved to ConfigMap, mounted to pod at valkey.conf
# also 
config: 
  key: value

# if defined, this is in cluster mode, and issues the commands to the cluster server
cluster:
  slots: 0-10000
```

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
jdheyburn: `valkeyConfig` or `config` is defined in standalone mode too. This comment refers to the `cluster` section below it

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
jdheyburn: I am removing it. In a previous iteration I had Sentinel config here too, but I then decided to keep ValkeyNode exclusive for a valkey server.

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
jdheyburn: Very fair. I'll work on condensing it down so that it focuses on CRD spec.

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
jdheyburn: I think we need a discussion about how the pods are named, everyone has their own opinion of it.

docs/rfc/INITIAL_CRD_DESIGN_STRATEGY.md
jdheyburn: I think it is a grey-area, one that I am happy to go with the consensus on. We could have `replicasPerShard` or `podsPerShard`. For now, I wanted to reuse the terminology from K8s domain.

jdheyburn:
Closing this PR, but will aim to reference this for future discussions

PR #68, opened 2026-01-29
add startup probe to valkey containers

ysqyang:
As titled. Without a startup probe, the existing liveness check can fail repeatedly and restart the container before Valkey finishes starting. This PR adds a startup probe that uses the existing `liveness-check.sh` with a longer failure budget (about 150s) to allow slow boots.

### DIFF
--- a/internal/controller/deployment_test.go
+++ b/internal/controller/deployment_test.go
@@ -17,6 +17,9 @@ limitations under the License.
 package controller
 
 import (
+	"os"
+	"os/exec"
+	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


internal/controller/scripts_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+func TestLivenessCheckScript(t *testing.T) {
+	scriptPath := filepath.Join("scripts", "liveness-check.sh")
```
sandeepkunusoth: why cant we add these in deployment_test.go

internal/controller/deployment_test.go
```diff
@@ -237,3 +240,68 @@ func findContainer(containers []corev1.Container, name string) *corev1.Container
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			if err := runProbeScript(t, scriptPath, test.response); (err != nil) != test.wantErr {
+				t.Fatalf("unexpected result: err=%v wantErr=%v", err, test.wantErr)
+			}
+		})
+	}
+}
+
```
bjosv, line 289: Adding a comment that the script is run with a stubbed `valkey-cli` might be helpful

internal/controller/deployment_test.go
```diff
@@ -237,3 +240,68 @@ func findContainer(containers []corev1.Container, name string) *corev1.Container
...
+func runProbeScript(t *testing.T, scriptPath, response string) error {
+	t.Helper()
+
+	binDir := t.TempDir()
+	valkeyCli := filepath.Join(binDir, "valkey-cli")
+	script := []byte("#!/bin/sh\n" +
+		"echo \"${VALKEY_RESPONSE:-PONG}\"\n")
```
bjosv, line 297: An alternative is
```suggestion
	script := `#!/bin/sh
echo "${VALKEY_RESPONSE:-PONG}"
`		
```
and then `[]byte(script)` below

bjosv (approved):

internal/controller/deployment_test.go
ysqyang: done (line 293)

Issue #31, opened 2025-12-18
(feat): Support NodeSelector Configuration for ValkeyCluster


jdheyburn:
## Summary

Apply `spec.nodeSelector` from ValkeyCluster to pods to constrain which nodes the pods can be scheduled on.

## Current State

- ✅ `spec.nodeSelector` field exists in API (line 46-48 in `api/v1alpha1/valkeycluster_types.go`)
- ❌ Not applied to pod spec in deployment

## Implementation

**File:** `internal/controller/deployment.go`

Update the pod spec in `createClusterDeployment()`:

```go
Spec: corev1.PodSpec{
    Containers:   containers,
    NodeSelector: cluster.Spec.NodeSelector,  // ADD THIS
    Volumes:      volumes,
}
```

## Testing

```yaml
apiVersion: valkey.io/v1alpha1
kind: ValkeyCluster
metadata:
  name: valkeycluster-nodeselector
spec:
  shards: 2
  replicas: 1
  nodeSelector:
    disktype: ssd
    zone: us-west-1a
```

```bash
kubectl apply -f test.yaml

# Verify nodeSelector is applied
kubectl get pods -l app.kubernetes.io/instance=valkeycluster-nodeselector \
  -o jsonpath='{.items[0].spec.nodeSelector}'

# Should output:
# {"disktype":"ssd","zone":"us-west-1a"}

# Verify pods only schedule on matching nodes
kubectl get pods -l app.kubernetes.io/instance=valkeycluster-nodeselector -o wide
# Check NODE column matches nodeSelector constraints
```

## Acceptance Criteria

- [ ] `spec.nodeSelector` applied to pod spec
- [ ] Single label selector works
- [ ] Multiple label selectors work
- [ ] Works when omitted (no nodeSelector set)
- [ ] Pods only schedule on nodes matching selector
- [ ] Pods stay pending if no nodes match selector

## References

- [Kubernetes Node Selection](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector)

jdheyburn:
I'll give this one a go at some point this week.

PR #69, opened 2026-01-30
propagate nodeSelector to pod spec

ysqyang:
[Issue](https://github.com/valkey-io/valkey-operator/issues/31)

Testing in an EKS cluster with 2 values for label `topology.kubernetes.io/zone`:
Screenshot 2026-01-29 at 6 10 52 PM

Without adding `nodeSelector`, when applying `config/samples/v1alpha1_valkeycluster.yaml`, valkey pods were placed on nodes with labels `topology.kubernetes.io/zone=us-east-1b` and `topology.kubernetes.io/zone=us-east-1c`:

Screenshot 2026-01-29 at 4 51 43 PM

After adding `nodeSelector: topology.kubernetes.io/zone: us-east-1b` in the manifest, nodeSelector was successfully applied:
Screenshot 2026-01-29 at 5 27 54 PM

and valkey pods were only placed on nodes with label `topology.kubernetes.io/zone=us-east-1b`.
Screenshot 2026-01-29 at 4 52 38 PM

If we change the node selector to `topology.kubernetes.io/zone: us-east-1a`, the pods stayed pending indefinitely because there's no node with this label:
Screenshot 2026-01-29 at 11 16 01 PM

Multiple selectors:
Adding `nodeSelector:
  topology.kubernetes.io/zone: us-east-1b
  kubernetes.io/hostname: ip-10-9-58-63.ec2.internal` in the manifest put all valkey pods on the same node in `us-east-1b`:
Screenshot 2026-01-31 at 3 43 06 PM

### DIFF
--- a/internal/controller/deployment.go
+++ b/internal/controller/deployment.go
@@ -118,10 +118,6 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
 
 func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.Deployment {
 	containers := generateContainersDef(cluster)
-	nodeSelector := cluster.Spec.NodeSelector
-	if cluster.Spec.Affinity != nil {
-		nodeSelector = nil
-	}
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: cluster.Name + "-",
@@ -140,8 +136,8 @@ func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.De
 				Spec: corev1.PodSpec{
 					Containers:   containers,
 					Affinity:     cluster.Spec.Affinity,
-					NodeSelector: nodeSelector,
-					Tolerations: cluster.Spec.Tolerations,
+					NodeSelector: cluster.Spec.NodeSelector,
+					Tolerations:  cluster.Spec.Tolerations,
 					Volumes: []corev1.Volume{
 						{
 							Name: "scripts",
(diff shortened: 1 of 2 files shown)


internal/controller/deployment.go
```diff
@@ -102,6 +102,10 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
 
 func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.Deployment {
 	containers := generateContainersDef(cluster)
+	nodeSelector := cluster.Spec.NodeSelector
+	if cluster.Spec.Affinity != nil {
+		nodeSelector = nil
```
bjosv: I guess you could have both Affinity and a NodeSelector configured at the same time (See note at the end of [this section](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity)).
I might have missed something, is there a reason for blocking a user to finetune placement by resetting nodeSelector here?

ysqyang: This is just to align with line 67 in https://github.com/valkey-io/valkey-operator/blob/main/api/v1alpha1/valkeycluster_types.go. I'm not sure why we wanted the override in the first place TBH, but that’s probably a discussion for another time.

ysqyang:
@jdheyburn Could you kindly review this? Thanks!

internal/controller/deployment.go
jdheyburn: I think we should remove this logic. A user may specify a nodeSelector, and a podAffinity / podAntiAffinity and not a nodeAffinity. If a user has declared a config we should allow it to propagate and not remove on their behalf.

internal/controller/deployment_test.go
```diff
@@ -86,6 +86,44 @@ func TestCreateClusterDeployment_SetsPodAntiAffinity(t *testing.T) {
...
+			Image:        "container:version",
+			NodeSelector: nodeSelector,
+			Affinity:     affinity,
+		},
+	}
+
+	d := createClusterDeployment(cluster)
+
+	assert.Nil(t, d.Spec.Template.Spec.NodeSelector, "node selector should be nil when affinity set")
+}
```
jdheyburn: This test can be removed once the previous comment is addressed.

PR #73, opened 2026-02-02
Pin GitHub Actions

SuperQ:
It is recommended for supply chains security to ping GitHub Actions to specific commit hashes in order to avoid stealth updates. Apply the latest commit hash for each of the used actions.
* Update all GitHub actions to latest.
* Enable monthly dependabot config to keep actions up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
--- a/.github/workflows/auto_update.yml
+++ b/.github/workflows/auto_update.yml
@@ -24,7 +24,7 @@ jobs:
     # Step 1: Checkout the repository.
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
         fetch-depth: 0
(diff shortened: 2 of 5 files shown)


.github/workflows/auto_update.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     # Step 1: Checkout the repository.
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
```
bjosv: I've seen there is an update to 6.0.2 lately

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
bjosv, line 6: Its quite nice to get all monthly updates grouped into a single PR, like

```suggestion
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
    groups:
      github-actions:
        patterns:
          - "*"
```

SuperQ: I don't typically like to group unrelated updates since it makes it harder to bisect issues.

.github/workflows/auto_update.yml
SuperQ: Oh yea, that's an issue with my fixup script. Well, feature. It uses GitHub API to access the "latest" release. They released v6.0.2 without making it "latest".

SuperQ: Fixed.

.github/dependabot.yml
bjosv: Its fine, we don't use that many actions. At this stage it wont pollute the Git history with never ending dependabot commits.

jdheyburn (approved):
Thank you!

PR #81, opened 2026-02-09
Use deterministic deployment names to encode shard topology

ysqyang:
## Context

https://github.com/valkey-io/valkey-operator/issues/55. This is a precursor to the resharding PR: https://github.com/valkey-io/valkey-operator/pull/60

## What changed

- Deployments are now **deterministically named** (`<cluster>-<N>-<M>`, e.g. `mycluster-0-0`) instead of using `GenerateName` with random suffixes.
- Each Deployment and its pod template are stamped with `valkey.io/shard-index` and `valkey.io/node-index` labels at creation time. By convention, node 0 is the initial primary and nodes 1+ are replicas.
- `addValkeyNode` reads the pod's `node-index` label (via `podRoleAndShard`) to determine whether a pending node should become a primary (assign slots) or a replica (`CLUSTER REPLICATE` to the matching shard's primary).
- A new `pickPendingNode` helper ensures primaries are processed before replicas, since `CLUSTER REPLICATE` requires the primary to already have slots.
- Slot-assignment and replication logic are extracted into dedicated `assignSlotsToNewPrimary` and `replicateToShardPrimary` functions.

## Why

The old code guessed node roles by counting existing shards — if fewer shards existed than desired, the next pending node became a primary; otherwise it became a replica. This heuristic was fragile: it could misassign roles when pods started out of order or when gossip hadn't yet propagated topology changes. Labels make the intended role explicit at Deployment creation time, removing all guesswork from the reconciler.

Deterministic names make it easy to correlate pods, deployments, and Valkey nodes during debugging (`kubectl get pods` immediately shows which shard and position each pod belongs to), and make `upsertDeployments` idempotent — it simply tries to create each name and ignores `AlreadyExists`.

## Testing

Screenshot 2026-02-13 at 9 19 04 AM

### DIFF
--- a/internal/controller/deployment.go
+++ b/internal/controller/deployment.go
@@ -123,20 +123,20 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
 // node Deployment. The name encodes the shard index and a node index within
 // the shard:
 //
-//	<cluster>-shard<N>-<M>    e.g. "mycluster-shard0-0", "mycluster-shard1-2"
+//	<cluster>-<N>-<M>    e.g. "mycluster-0-0", "mycluster-1-2"
 //
 // By convention, node 0 is the initial primary and nodes 1, 2, … are replicas.
 // The name deliberately avoids "primary"/"replica" because failover can swap
 // roles at any time. The authoritative role lives in the pod labels.
 func deploymentName(clusterName string, shardIndex int, nodeIndex int) string {
-	return fmt.Sprintf("%s-shard%d-%d", clusterName, shardIndex, nodeIndex)
+	return fmt.Sprintf("%s-%d-%d", clusterName, shardIndex, nodeIndex)
 }
 
 // createClusterDeployment builds a single-pod Deployment for a Valkey node.
 //
 // Each Deployment manages exactly one Pod (Replicas=1) and represents one
 // logical node in the Valkey cluster. The Deployment name encodes the shard
-// and node index (e.g. "mycluster-shard0-0", "mycluster-shard1-2"). Two
+// and node index (e.g. "mycluster-0-0", "mycluster-1-2"). Two
 // labels (valkey.io/shard-index and valkey.io/node-index) provide selector
 // uniqueness so no two Deployments fight over the same Pod.
 func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex int, nodeIndex int) *appsv1.Deployment {
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -17,11 +17,9 @@ limitations under the License.
 package controller
 
 import (
-	"fmt"
 	"maps"
 	"slices"
 	"strconv"
-	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	valkeyv1 "valkey.io/valkey-operator/api/v1alpha1"
@@ -35,7 +33,7 @@ const appName = "valkey"
 // Every Deployment (and therefore every Pod) encodes the node's position in
 // the Valkey cluster in its *name*:
 //
-//	<cluster>-shard<N>-<M>    e.g. "mycluster-shard0-0", "mycluster-shard1-2"
+//	<cluster>-<N>-<M>    e.g. "mycluster-0-0", "mycluster-1-2"
 //
 // where N is the shard index and M is the node index within the shard. By
 // convention, node 0 is the initial primary and nodes 1, 2, … are replicas.
@@ -47,12 +45,15 @@ const appName = "valkey"
 //	valkey.io/shard-index  – which shard ("0", "1", …)
 //	valkey.io/node-index   – node within shard ("0", "1", …)
 //
-// The reconciler parses the pod *name* (via podRoleAndShard) to decide
-// whether to assign slots (node 0 = primary) or issue CLUSTER REPLICATE
-// (node 1+ = replica), and for which shard.
+// The reconciler reads pod labels (via podRoleAndShard) to decide whether
+// to assign slots (node 0 = initial primary) or issue CLUSTER REPLICATE
+// (node 1+ = initial replica), and for which shard. This convention only
+// applies during initial cluster creation; after a failover, Valkey may
+// promote a replica to primary, making node 0 a replica. The labels are
+// not updated to reflect this — the live role is always read from
+// CLUSTER NODES, not from the labels.
 //
 // Names are set by deploymentName, labels by createClusterDeployment.
-// Both are consumed by addValkeyNode → podRoleAndShard.
 const (
 	// LabelShardIndex identifies which shard a pod belongs to (e.g. "0", "1", "2").
 	LabelShardIndex = "valkey.io/shard-index"
@@ -88,43 +89,23 @@ func annotations(cluster *valkeyv1.ValkeyCluster) map[string]string {
 	return maps.Clone(cluster.Annotations)
 }
 
-// podRoleAndShard finds the pod matching the given IP address and parses its
-// Deployment-generated name to extract the role and shard index.
+// podRoleAndShard finds the pod matching the given IP address and reads its
+// labels to determine the intended role and shard index.
 //
-// Pod names follow the pattern:
-//
-//	<cluster>-shard<N>-<M>-<replicaset>-<hash>
-//
-// where N is the shard index and M is the node index within the shard. By
-// convention, node 0 is the initial primary and nodes 1+ are replicas. Returns
-// ("", -1) if the pod is not found or the name doesn't match.
+// The role is derived from valkey.io/node-index: node 0 is the initial
+// primary, nodes 1+ are replicas. Returns ("", -1) if the pod is not
+// found or the labels are missing.
 func podRoleAndShard(address string, pods *corev1.PodList) (string, int) {
 	idx := slices.IndexFunc(pods.Items, func(p corev1.Pod) bool { return p.Status.PodIP == address })
 	if idx == -1 {
 		return "", -1
 	}
-	return parseRoleAndShard(pods.Items[idx].Name)
-}
-
-// parseRoleAndShard extracts the role and shard index from a pod name like
-// "mycluster-shard2-0-abc12-xyz". Node index 0 → primary, 1+ → replica.
-func parseRoleAndShard(podName string) (string, int) {
-	shardPos := strings.Index(podName, "-shard")
-	if shardPos == -1 {
-		return "", -1
-	}
-	// After "-shard": "2-0-abc12-xyz"
-	rest := podName[shardPos+len("-shard"):]
-	// Split: ["2", "0", "abc12", "xyz"]
-	parts := strings.SplitN(rest, "-", 3)
-	if len(parts) < 2 {
-		return "", -1
-	}
-	shardIndex, err := strconv.Atoi(parts[0])
+	pod := &pods.Items[idx]
+	shardIndex, err := strconv.Atoi(pod.Labels[LabelShardIndex])
 	if err != nil {
 		return "", -1
 	}
-	nodeIndex, err := strconv.Atoi(parts[1])
+	nodeIndex, err := strconv.Atoi(pod.Labels[LabelNodeIndex])
 	if err != nil {
 		return "", -1
 	}
@@ -134,12 +115,12 @@ func parseRoleAndShard(podName string) (string, int) {
 	return RoleReplica, shardIndex
 }
 
-// primaryPodIP finds the IP of the primary pod (node 0) for a given shard by
-// parsing pod names. Returns "" if not found.
+// primaryPodIP finds the IP of the node-0 (primary) pod for a given shard by
+// reading pod labels. Returns "" if not found.
 func primaryPodIP(pods *corev1.PodList, shardIndex int) string {
-	target := fmt.Sprintf("-shard%d-0-", shardIndex)
+	si := strconv.Itoa(shardIndex)
 	idx := slices.IndexFunc(pods.Items, func(p corev1.Pod) bool {
-		return strings.Contains(p.Name, target)
+		return p.Labels[LabelShardIndex] == si && p.Labels[LabelNodeIndex] == "0"
 	})
 	if idx == -1 {
 		return ""
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -76,9 +76,8 @@ var scripts embed.FS
 //  1. Ensure the headless Service exists (upsertService).
 //  2. Ensure the ConfigMap with valkey.conf and health-check scripts exists
 //     (upsertConfigMap).
-//  3. Ensure one Deployment per (shard, role) pair exists, each named
-//     deterministically (e.g. mycluster-shard0-primary) and labelled with
-//     valkey.io/shard-index and valkey.io/role (upsertDeployments).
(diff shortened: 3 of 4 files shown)


internal/controller/deployment.go
```diff
@@ -116,22 +118,53 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
...
+	// Deployment-level labels carry shard/role for easy listing/filtering.
+	deployLabels := labels(cluster)
+	deployLabels[LabelShardIndex] = strconv.Itoa(shardIndex)
+	deployLabels[LabelRole] = role
+
+	// Pod labels include shard/role so the controller can look up a pod's
+	// intended role via pod.Labels without touching Valkey.
+	podLabels := labels(cluster)
+	podLabels[LabelShardIndex] = strconv.Itoa(shardIndex)
+	podLabels[LabelRole] = role
```
sandeepkunusoth: is there any difference between deployLabels and podLabels can we use common func for both which takes shard and returns labels maybe in  utils.go?

ysqyang: For now, there's no difference. The small redundancy here is just to ensure that deployLabels and podLabels are 2 independent copies so changing one won't affect the other.

internal/controller/valkeycluster_controller.go
```diff
@@ -345,61 +451,136 @@ func (r *ValkeyClusterReconciler) addValkeyNode(ctx context.Context, cluster *va
...
-			log.Error(err, "command failed: CLUSTER ADDSLOTSRANGE", "slotStart", slotStart, "slotEnd", slotEnd)
-			r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "SlotAssignmentFailed", "AssignSlots", "Failed to assign slots: %v", err)
-			return err
+	return errors.New("pod has no valkey.io/role label; cannot determine node role")
+}
+
+// podRoleAndShard looks up the pod matching the given IP address and returns
+// its valkey.io/role and valkey.io/shard-index labels. Returns ("", -1) if
+// the pod is not found or has no labels.
+func (r *ValkeyClusterReconciler) podRoleAndShard(address string, pods *corev1.PodList) (string, int) {
```
sandeepkunusoth: i feel these functions dont need to be part of main controller.go reconciler it ca be moved to utils?

ysqyang: moved

internal/controller/deployment.go
sandeepkunusoth: As we have 1-1 mapping between pods and deployments, labels will stay same. we can add later if required. just to keep it simple.

internal/controller/valkeycluster_controller.go
```diff
@@ -281,26 +332,65 @@ func (r *ValkeyClusterReconciler) upsertDeployments(ctx context.Context, cluster
...
-	// TODO: update existing
+	// TODO: update existing Deployments when the spec changes (e.g. image upgrade).
 
 	return nil
 }
 
+// createLabeledDeployment creates a single Deployment for the given (shard, role)
+// pair, sets the ValkeyCluster as its owner, and emits a Kubernetes event.
+func (r *ValkeyClusterReconciler) createLabeledDeployment(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, shard int, role string, created, expected int) error {
+	deployment := createClusterDeployment(cluster, shard, role)
```
sandeepkunusoth: why did we add this extra layer createLabeledDeployment. createClusterDeployment is the function which is actually adding labels. and we are passing created expected vars just for logging? i would recommend to use same function keep the diff simple createClusterDeployment

internal/controller/deployment.go
ysqyang: fixed

internal/controller/valkeycluster_controller.go
ysqyang: fixed

internal/controller/deployment.go
```diff
@@ -116,22 +118,43 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
...
 
-func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.Deployment {
+// createClusterDeployment builds a single-pod Deployment for a Valkey node.
+//
+// Each Deployment manages exactly one Pod (Replicas=1) and represents one
+// logical node in the Valkey cluster. The shardIndex and role parameters are
+// stamped as Kubernetes labels on both the Deployment and the Pod template:
+//
+//	valkey.io/shard-index = "<shardIndex>"
+//	valkey.io/role        = "primary" | "replica"
```
jdheyburn: Does this represent the nodes desired replication role, or current replication role? I think we should be explicit here: `valkey.io/desired-replication-role`

internal/controller/valkeycluster_controller.go
```diff
@@ -115,16 +132,33 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	// lookup would fail.
 	if len(state.PendingNodes) > 0 {
 		node := state.PendingNodes[0]
+		for _, n := range state.PendingNodes {
+			role, _ := podRoleAndShard(n.Address, pods)
+			if role == RolePrimary {
+				node = n
+				break
+			}
+		}
```
jdheyburn: Could we extract this to a function? Something like `getCandidatePendingNode`

internal/controller/deployment.go
ysqyang: It is desired replication role, but I think we're better off just using `valkey.io/shard-index` and `valkey.io/node-index`. Roles are subject to change during failover.

internal/controller/valkeycluster_controller.go
ysqyang: done

internal/controller/deployment.go
```diff
@@ -116,22 +119,47 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
...
 
-func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster) *appsv1.Deployment {
+// deploymentName returns a deterministic, human-readable name for a Valkey
+// node Deployment. The name encodes the shard index and a node index within
+// the shard:
+//
+//	<cluster>-shard<N>-<M>    e.g. "mycluster-shard0-0", "mycluster-shard1-2"
+//
+// By convention, node 0 is the initial primary and nodes 1, 2, … are replicas.
+// The name deliberately avoids "primary"/"replica" because failover can swap
```
bjosv, line 129: I also think its good that we avoid primary/replica in the name.
We only have to keep in mind, as stated here, that after a failure the primary can be at index 2 (of 0-2).
If the user want to scale down replicas either the replica at index 0 or index 1 needs to go, i.e. there can be indexes missing and the index doesn't say that much anymore.

In StatefulSets indexes are used since it's always the last index/replica that is removed, but in Deployments, where any replica can be removed (depending of other factors), they have chosen to use a unique name/hash instead.
I hesitate a bit to use "index" instead of a random suffix, but I understand the reason for using it..

internal/controller/deployment.go
```diff
@@ -116,22 +119,47 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
...
+	return fmt.Sprintf("%s-shard%d-%d", clusterName, shardIndex, nodeIndex)
+}
+
+// createClusterDeployment builds a single-pod Deployment for a Valkey node.
+//
+// Each Deployment manages exactly one Pod (Replicas=1) and represents one
+// logical node in the Valkey cluster. The Deployment name encodes the shard
+// and node index (e.g. "mycluster-shard0-0", "mycluster-shard1-2"). Two
+// labels (valkey.io/shard-index and valkey.io/node-index) provide selector
+// uniqueness so no two Deployments fight over the same Pod.
```
bjosv, line 141: Not sure about the "fight" comment here, how does two Deployments fight over a pod?

internal/controller/utils.go
```diff
@@ -17,13 +17,57 @@ limitations under the License.
...
+// The name deliberately avoids "primary"/"replica" because failover can swap
+// roles at any time.
+//
+// Two labels are set for selector uniqueness and kubectl convenience:
+//
+//	valkey.io/shard-index  – which shard ("0", "1", …)
+//	valkey.io/node-index   – node within shard ("0", "1", …)
+//
+// The reconciler parses the pod *name* (via podRoleAndShard) to decide
+// whether to assign slots (node 0 = primary) or issue CLUSTER REPLICATE
```
bjosv: Not sure if we should say `(node 0 = primary)` here, or should it mention that its only when creating the cluster? Similar comment below.

internal/controller/utils.go
```diff
@@ -43,3 +87,76 @@ func labels(cluster *valkeyv1.ValkeyCluster) map[string]string {
...
+	shardIndex, err := strconv.Atoi(parts[0])
+	if err != nil {
+		return "", -1
+	}
+	nodeIndex, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return "", -1
+	}
+	if nodeIndex == 0 {
+		return RolePrimary, shardIndex
```
bjosv, line 113: Is this function only expected to be used on pending nodes not yet add?
I guess so since this is only correct until there is an failover.
We could also label deployments with its expected role, i.e not using index but some label describing expected cluster role.

internal/controller/valkeycluster_controller.go
```diff
@@ -69,8 +69,26 @@ var scripts embed.FS
 // +kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
 
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
+// Reconcile is the main reconciliation loop. On each invocation it drives the
```
bjosv, line 72: Nice addition, we need to make sure its updated after each change.

internal/controller/valkeycluster_controller.go
```diff
@@ -69,8 +69,26 @@ var scripts embed.FS
...
+// cluster one step closer to the desired state described by the ValkeyCluster
+// spec. The pipeline runs in the following order:
+//
+//  1. Ensure the headless Service exists (upsertService).
+//  2. Ensure the ConfigMap with valkey.conf and health-check scripts exists
+//     (upsertConfigMap).
+//  3. Ensure one Deployment per (shard, role) pair exists, each named
+//     deterministically (e.g. mycluster-shard0-primary) and labelled with
```
bjosv: `mycluster-shard0-primary` here is probably an old iteration, but I'm in favor of using a label that could be updated after a failover. These need to be updated in the reconciliation loop.
If we go for using index, are we aiming to force the primary to be placed at index 0, i.e. failback to index 0 as soon as possible?

internal/controller/utils.go
jdheyburn: > We could also label deployments with its expected role, i.e not using index but some label describing expected cluster role.

I was thinking along this line in the below comment. We may want to query for pods where a failback is due, and so querying for desired-replication-role vs current-replication-role may be useful.

https://github.com/valkey-io/valkey-operator/pull/81#discussion_r2791732459

internal/controller/deployment.go
jdheyburn: > We only have to keep in mind, as stated here, that after a failure the primary can be at index 2 (of 0-2).
If the user want to scale down replicas either the replica at index 0 or index 1 needs to go, i.e. there can be indexes missing and the index doesn't say that much anymore.

In this scenario, I would expect a failover to either index 0 or 1, and index 2 is removed.

internal/controller/deployment.go
jdheyburn: I believe it means that a Deployment selector will always have a 1:1 mapping to a pod (singleton Deployments).

internal/controller/utils.go
```diff
@@ -77,20 +88,75 @@ func annotations(cluster *valkeyv1.ValkeyCluster) map[string]string {
...
 
-// podRoleAndShard looks up the pod matching the given IP address and returns
-// its valkey.io/role and valkey.io/shard-index labels. Returns ("", -1) if
-// the pod is not found or has no labels.
+// podRoleAndShard finds the pod matching the given IP address and parses its
+// Deployment-generated name to extract the role and shard index.
+//
+// Pod names follow the pattern:
+//
+//	<cluster>-shard<N>-<M>-<replicaset>-<hash>
```
jdheyburn: When StatefulSets are introduced, would we expect this to be `<cluster>-shard<N>-<M>-<p>`, e.g. `valkey-mycluster-shard0-0-0`?

internal/controller/utils.go
```diff
@@ -77,20 +88,75 @@ func annotations(cluster *valkeyv1.ValkeyCluster) map[string]string {
...
-	pod := &pods.Items[idx]
-	role := pod.Labels[LabelRole]
-	shardStr := pod.Labels[LabelShardIndex]
-	shardIndex, err := strconv.Atoi(shardStr)
+	return parseRoleAndShard(pods.Items[idx].Name)
+}
+
+// parseRoleAndShard extracts the role and shard index from a pod name like
+// "mycluster-shard2-0-abc12-xyz". Node index 0 → primary, 1+ → replica.
+func parseRoleAndShard(podName string) (string, int) {
```
jdheyburn: Since we already have the pod from the calling parent, couldn't we extract the labels on the pod to get the role and shard-index, instead of having to parse the pod name?

internal/controller/deployment.go
ysqyang: Without `valkey.io/shard-index` and `valkey.io/node-index`, all deployments have overlapping selectors, which means each Deployment's controller will independently count the matching Pods and try to reconcile to its desired replica count. So if Deployment A (replicas=1) and Deployment B (replicas=1) both select the same 2 Pods, each thinks "I want 1 but see 2" and tries to delete one. Then the other sees 0 and creates one. They'd thrash back and forth.

internal/controller/utils.go
ysqyang: correct

internal/controller/valkeycluster_controller.go
ysqyang: Thanks for catching this - this comment needs to be updated. We're no longer putting roles (primary vs replica) in deployments' or pods' names or labels, just shard indexes and node indexes. That said, I’m not planning to add controller logic to force the primary to be placed at index 0, as the benefits seem minimal relative to the added complexity. Users are expected to know that node-index = 0 does not always mean "primary".

internal/controller/utils.go
ysqyang: This is only used on pending nodes for now. After a node has been bootstrapped (slots assigned, replicas attached), it moves out of PendingNodes into state.Shards and this function is never called for it again.

bjosv: After a pod-failure with a primary Valkey node, a newly created pod will also be a pending node.
Since a pod with nodeIndex=1 might have take on that role in this shard nodeIndex=0 doesn't mean that is should become a primary, i.e in most failure cases the name will not be deterministic. 
I guess we also need to check the current topology before adding this node.

internal/controller/deployment.go
jdheyburn: Failing logic is going to be addressed in a future PR

PR #82, opened 2026-02-10
(feat) Support for creating users on cluster init

### LINKED ISSUE #36: (feat)  Support ACL user creation on cluster init

utdrmac:
Resolves #36 

This feature allows for the creation of Valkey users on cluster initialization. The feature abstracts Valkey ACL permissions into several objects, giving better readability, and creation flexibility to users that may not be intimately familiar with Valkey ACLs.

### DIFF
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -36,6 +36,7 @@ import (
 
 const (
 	hashAnnotationKey = "valkey.io/internal-acl-hash"
+	aclFilename       = "users.acl"
 )
 
 func getInternalSecretName(clusterName string) string {
@@ -115,10 +116,11 @@ func (r *ValkeyClusterReconciler) reconcileUsersAcl(ctx context.Context, cluster
 	}
 	usersAclsBytes := []byte(usersAcls.String())
 
+	// Calculate hash of the ACL file contents
+	internalAclHash := fmt.Sprintf("%x", sha256.Sum256(usersAclsBytes))
+
 	// An "internal" secrets object is used for synchronization
 	internalSecretName := getInternalSecretName(cluster.Name)
-	needCreateInternal := false
-
 	internalAclSecret := &corev1.Secret{}
 	if err := r.Get(ctx, types.NamespacedName{
 		Name:      internalSecretName,
@@ -131,53 +133,53 @@ func (r *ValkeyClusterReconciler) reconcileUsersAcl(ctx context.Context, cluster
 
 		// Internal secret was not found.
 		// Init, and add metadata to the new Secret object
-		needCreateInternal = true
 		log.V(2).Info("creating internal secret", "secretName", internalSecretName)
 
-		internalAclSecret.Data = make(map[string][]byte)
 		internalAclSecret.ObjectMeta = metav1.ObjectMeta{
 			Name:      internalSecretName,
 			Namespace: cluster.Namespace,
 			Labels:    labels(cluster),
+			Annotations: map[string]string{
+				hashAnnotationKey: internalAclHash,
+			},
 		}
-	}
+		internalAclSecret.Data = map[string][]byte{
+			aclFilename: usersAclsBytes,
+		}
 
-	// Register ownership of the internal Secret so that it is GC'd by K8S on CR delete
-	if err := controllerutil.SetControllerReference(cluster, internalAclSecret, r.Scheme); err != nil {
-		log.Error(err, "Failed to grab ownership of internal secret")
-		r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "InternalSecretsCreationFailed", "ReconcileUsers", "Failed to grab ownership of internal secret: %v", err)
-		return err
-	}
+		// Register ownership of the new internal Secret
+		if err := controllerutil.SetControllerReference(cluster, internalAclSecret, r.Scheme); err != nil {
+			log.Error(err, "Failed to grab ownership of internal secret")
+			r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "InternalSecretsCreationFailed", "ReconcileUsers", "Failed to grab ownership of internal secret: %v", err)
+			return err
+		}
 
-	// Calculate hash of the ACL file contents
-	internalAclHash := fmt.Sprintf("%x", sha256.Sum256(usersAclsBytes))
-
-	// Compare hash to the one already attached to the internal secret, if present.
-	// If the hashes are different, then we need to update the internal secret with
-	// the new file contents and update the hash annotation. If the hashes are the
-	// same, don't update as that would cause infinite reconciliation
-
-	if needsUpdate := upsertAnnotation(internalAclSecret, hashAnnotationKey, internalAclHash); !needsUpdate {
-		log.V(1).Info("internal ACLs unchanged")
-		return nil
-	}
-
-	// Add the acl contents to the internal secret, replacing anything preexisting
-	internalAclSecret.Data["users.acl"] = usersAclsBytes
-
-	// Create the internal secret, if needed
-	if needCreateInternal {
+		// Create the internal Secret
 		if err := r.Create(ctx, internalAclSecret); err != nil {
 			log.Error(err, "Failed to create internal secret")
 			r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "InternalSecretsCreationFailed", "ReconcileUsers", "Failed to create internal secret: %v", err)
 			return err
-		} else {
-			r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "InternalSecretsCreated", "ReconcileUsers", "Created internal ACLs")
-			return nil
 		}
+
+		r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "InternalSecretsCreated", "ReconcileUsers", "Created internal ACLs")
+
+		// All good; new internal Secret with contents created
+		return nil
 	}
 
-	// Otherwise update it
+	// Internal Secret exists; Calculate, and compare the hashes to
+	// determine if anything needs updating. If the hashes are the
+	// same, don't update as that would cause infinite reconciliation
+
+	if !upsertAnnotation(internalAclSecret, hashAnnotationKey, internalAclHash) {
+		log.V(1).Info("internal ACLs unchanged")
+		return nil
+	}
+
+	// Hashes are different; Update the acl contents of the internal secret
+	internalAclSecret.Data[aclFilename] = usersAclsBytes
+
+	// Update secret
 	if err := r.Update(ctx, internalAclSecret); err != nil {
 		log.Error(err, "Failed to update internal secret")
 		r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "InternalSecretsUpdateFailed", "ReconcileUsers", "Failed to update internal secret: %v", err)


test/e2e/e2e_suite_test.go
```diff
@@ -65,9 +65,14 @@ func TestE2E(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	By("building the manager image")
-	cmd := exec.Command("make", "docker-build", fmt.Sprintf("IMG=%s", managerImage))
+	By("purging old events")
```
utdrmac, line 68: K8S holds on to events for an extended period of time, even after a cluster is deleted. This call to purge events was added because subsequent e2e tests would fail if a previous test failed due to the e2e test fetching _all_ events.

Makefile
```diff
@@ -105,7 +105,7 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 ##@ Build
 
 .PHONY: build
-build: manifests generate fmt vet ## Build manager binary.
+build: manifests generate fmt vet lint ## Build manager binary.
```
utdrmac, line 108: The golangci-lint was only ran during github e2e tests, and it was frustrating for this to fail remotely. Simple change to check this locally first.

utdrmac (commented):
I added some comments for 2 changes outside the scope of the PR.

internal/controller/valkeycluster_controller.go
```diff
@@ -506,6 +514,10 @@ func (r *ValkeyClusterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Service{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&appsv1.Deployment{}).
+		Watches(
+			&corev1.Secret{},
+			handler.EnqueueRequestsFromMapFunc(r.findReferencedClusters),
+		).
```
sandeepkunusoth, line 793: i feel in initial version we dont need to watch for secrets. we can just read exisiting secret for given valkey cluster and use that during valkey cluster creation. 

watching secrets, and updating existing valkey cluster can be seperate PR. As i think update is not fully working in current PR.

utdrmac: Can you explain "update not fully working"? When you update any referenced secret for a user, cluster reconciliation is triggered, as expected. Yes, it is acknowledged that current behavior does fully update the ACL inside Valkey, and that will be in another PR. This PR supports each user having their own secret; it's not just a single secret for all users, as per the proposed spec in #36

internal/controller/users.go
```diff
@@ -0,0 +1,311 @@
...
+			log.Error(nil, "missing password key in secret", "user", user.Name, "secret", userSecretName, "key", key)
+			return []string{}, fmt.Errorf("missing password key in secret")
+		}
+
+		// If the Secret begins with # (byte 35) and is 65 total characters long,
+		// we assume this is a pre-hashed password
+		if password[0] == 35 && len(password) == 65 {
+			passwords = append(passwords, string(password[1:]))
+			continue
+		}
```
jdheyburn: I still don't know how I feel about this. Do you mind providing some context for this?

internal/controller/users.go
```diff
@@ -0,0 +1,311 @@
...
+			log.Error(err, "failed to fetch acl secret")
+			return []string{}, err
+		}
+		log.V(1).Info("Users secret not found", "userSecretName", userSecretName)
+
+		// The Secret was not found; Also, if NoPassword is false, then we cannot add this user
+		if !user.NoPassword {
+			return []string{}, fmt.Errorf("no password or reference found")
+		}
+	}
```
jdheyburn, line 276: Because of line 250, user.NoPassword will always be false. Do we need the conditional check here?

internal/controller/users.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// Holds the ACL as we build it
+	var acl strings.Builder
+
+	// Helper for repeated actions
+	appendAcl := func(acl *strings.Builder, permissions []string, prefix string) {
+		for _, permission := range permissions {
+			fmt.Fprintf(acl, " %s%s", prefix, permission)
+		}
+	}
```
jdheyburn: Would it be better to declare this outside of a function? Otherwise on every `buildUserAcl` call there would be a new function being declared and saved to the `appendAcl` variable.

internal/controller/users.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// If enabled, append password(s), which should already be prefix-hashed
+	if user.NoPassword {
+		fmt.Fprintf(&acl, " nopass")
+	} else {
+		appendAcl(&acl, passwords, "#")
+	}
+
+	// Add key restrictions
+	appendAcl(&acl, user.Keys.ReadWrite, "~")
```
jdheyburn, line 226: `user.Keys.ReadWrite`, and other attributes, are optional. Do we need to wrap this behind a if condition? Same for elsewhere that you're appending to ACL.

internal/controller/valkeycluster_controller.go
```diff
@@ -249,7 +256,8 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
 			"valkey.conf": `
 cluster-enabled yes
 protected-mode no
-cluster-node-timeout 2000`,
+cluster-node-timeout 2000
+aclfile /config/users/users.acl`,
```
jdheyburn, line 341: Will the aclfile always exist, even if a user hasn't defined any users?

internal/controller/users.go
```diff
@@ -0,0 +1,311 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
+)
+
+const (
+	hashAnnotationKey = "valkey.io/internal-acl-hash"
+)
+
+func getInternalSecretName(clusterName string) string {
+	return "internal-" + clusterName + "-acl"
```
jdheyburn, line 43: Would `clusterName + "acl"` suffice? I don't think its necessary to mention this is internal.

internal/controller/valkeycluster_controller.go
sandeepkunusoth: my point was we are watching secrets only to update secrets of the users and later update deployment with updated secrets. but today we are only updating individual users secrets. it doesn't help anyone when user updates secrets today. unless they want to manually restart the deployments after acl is updated.

jdheyburn: My understanding is that the external Secret manifest that is deployed outside of a ValkeyCluster needs to be watched for updates, so that the controller can update the Secret containing the ACL file.

Live reloading the ACL file (perhaps via `ACL LOAD`) on the server after that change is made isn't in scope for this PR.

internal/controller/users.go
jdheyburn: Perhaps `aclSecretName` would be a better name here.

utdrmac: "internal-" is a pattern followed by all of the Percona [operators](https://github.com/percona/percona-xtradb-cluster-operator/blob/main/pkg/controller/pxc/secrets.go#L23) for secrets and config objects which are created/owned by the operator. I was simply following that pattern. It's a naming convention indication that users shouldn't modify it.

internal/controller/valkeycluster_controller.go
utdrmac: Yes, it always exists. The file is created, empty, if there's no users defined.

internal/controller/users.go
utdrmac: They don't need if conditions. The undefined state of a slice is an empty slice, thus the `range` call in `appendAcl` has nothing to loop over.

internal/controller/users.go
utdrmac: Good catch.

internal/controller/users.go
utdrmac: The context is that users can put in the secret either A) a pre-hashed password, which would be 65 characters long, beginning with '#' (valkey format), or B) anything else, which would be considered plain-text, and the operator hashes it and appends the '#'. Some users will be super-password-secure, and not want plaintext in their Secrets, so we support that. Other users won't mind, and will put plaintext in their Secrets, which we should also support.

internal/controller/users.go
utdrmac: Gotcha. Fixed.

api/v1alpha1/valkeyacls_types.go
```diff
@@ -0,0 +1,102 @@
...
+	Enabled bool `json:"enabled,omitempty"`
+
+	// Reference information to a Secret containing user passwords
+	// +optional
+	PasswordSecret PasswordSecretSpec `json:"passwordSecret,omitempty"`
+
+	// Do not apply a password to this user
+	// +kubebuilder:default=false
+	NoPassword bool `json:"nopass,omitempty"`
+
```
sandeepkunusoth, line 37: i think passwordEnabled or enabledPassword would be good naming here instead of NoPassword. Also i am not sure this is general use case of having nopass attached to user. if they really want to acheieve this they can always use rawacl.

internal/controller/utils.go
```diff
@@ -19,14 +19,16 @@ package controller
...
 	valkeyv1 "valkey.io/valkey-operator/api/v1alpha1"
 )
 
 const appName = "valkey"
 
 // Labels returns a copy of user defined labels including recommended:
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
-func labels(cluster *valkeyv1.ValkeyCluster) map[string]string {
+// nolint:unparam
+func labels(cluster *valkeyv1.ValkeyCluster, extraLabels ...map[string]string) map[string]string {
```
sandeepkunusoth: any reason for adding nolint here. is it really needed?

internal/controller/users.go
jdheyburn: Fair, I'm not familiar with Percona's operators, but if its a recognised pattern then I'm game for it.

api/v1alpha1/valkeyacls_types.go
jdheyburn: It might not be general or recommended, but if Valkey exposes it as an option it might be useful to abstract it away.

https://valkey.io/topics/acl/#configure-acls-with-the-acl-command

This boolean is used in the PR in a few places. If we were to remove NoPassword here, would we change those conditions to be `'nopass' in RawAcl` (pseudocode) ?

internal/controller/users.go
jdheyburn: Okay, I think that's fine. It's not a pattern I've seen elsewhere. So long as we have this documented then it works for me.

jdheyburn: Maybe we can have a function for this to help readability:

```go
func isPreHashedPassword(password) {
  return password[0] == 35 && len(password) == 65
}
```

internal/controller/users.go
sandeepkunusoth: FYI, adding the internal- prefix reduces the available character budget for generated Kubernetes Secret names.
To avoid Secret creation failures, we’d need to enforce a stricter max length for ValkeyCluster names ( depending on other suffixes/prefixes as length of these increase, we need to restrict max length for cluster).

internal/controller/utils.go
utdrmac: Yes, because golangci-lint complains about `extraLabels` not being used. I guess I can remove the added functionality until some future time it is needed.

api/v1alpha1/valkeyacls_types.go
utdrmac: I think the idea was since the ACL keyword is `nopass` then this follows-ish 1:1. Since the default is false if not specified, is there a strong reason to remove it? As the code is now, if removed, then passwords are always searched for even in a nopass=true case. The code would then need to check rawAcl for presence of `nopass` to achieve same logic.

internal/controller/users.go
jdheyburn: Can you raise an issue for that @sandeepkunusoth ?

test/e2e/e2e_suite_test.go
bjosv: Good catch. We could move it to valkeycluster_test.go to have it close to where its needed.
This is also the only testcase/context where we use `involvedObject.name=valkeycluster-sample`.
I guess it needs to be in a `BeforeEach` instead then.

internal/valkey/clusterstate.go
```diff
@@ -124,7 +124,7 @@ func (s *ClusterState) GetUnassignedSlots() []SlotsRange {
 	remaining := []SlotsRange{{0, 16383}}
 	for _, shard := range s.Shards {
 		for _, slot := range shard.Slots {
-			var next []SlotsRange
+			var next []SlotsRange //nolint:prealloc
```
bjosv, line 130: Interesting that this isn't showing in CI, our current main (given by kubebuilder) uses
```
make lint            
...
Downloading github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.7.2
0 issues.
```
which is using
go: downloading github.com/alexkohler/prealloc v1.0.0

Is this seen with newer versions of golangci-lint/prealloc?

utdrmac: Must be. I'm using `golangci-lint has version 2.8.0 built with go1.25.5`

internal/controller/utils.go
```diff
@@ -57,23 +96,97 @@ func annotations(cluster *valkeyv1.ValkeyCluster) map[string]string {
...
+	}
+	return false
+}
+
+// findShardPrimary scans all pods with the given shard-index label and returns
+// the Valkey node ID + IP of whichever pod is currently the slot-bearing
+// primary, regardless of its node-index label. This handles the post-failover
+// case where node-index=1 (or higher) was promoted by Valkey.
+// Returns ("", "") if no primary is found.
+func findShardPrimary(state *valkey.ClusterState, shardIndex int, pods *corev1.PodList) (nodeID, ip string) {
```
jdheyburn, line 177: Is this, and the other functions above it, the result of a bad merge?

utdrmac: `upsertAnnotation` is added by this PR. `findShardPrimary` was added in d777c83  I'm not seeing any merge issues.

jdheyburn: My apologies, the comparison I was making made it appear as something different :(

jdheyburn (approved):

utdrmac:
@jdheyburn That is correct due to the logic flow. The first thing checked for is a password. If NoPassword is true, `fetchUserPasswords` exits early, and then then continues building the user. If NoPassword is false, (as is the case for charlie), the search inside the Secret takes place. This returns error because we can't add a user without password. If charlie did have a password, or NoPassword was true, he would get added in disabled ("user charlie off").

I guess it comes down to a design decision: Should users that are disabled appear in the ACL file? If yes, (current behavior), then the error is correct. If no, then the logic can be changed to completely skip users which are disabled (and print info message that user was skipped due to being disabled).

PR #83, opened 2026-02-13
RFC for ValkeyNode CRD proposal

jdheyburn:
Based off of the long term proposal I created below:
- https://github.com/jdheyburn/valkey-operator/blob/f042e6f63d206e67f1680fe9d66e6022254590db/docs/plans/2026-02-04-valkeynode-mvp-design.md

Github discussion poll:
- https://github.com/valkey-io/valkey-operator/discussions/84

### DIFF
--- a/docs/rfc/VALKEY_NODE_CRD_PROPOSAL.md
+++ b/docs/rfc/VALKEY_NODE_CRD_PROPOSAL.md
@@ -1,0 +1,274 @@
+# ValkeyNode CRD - Design Proposal
+
+**Date:** 2026-02-13
+**Status:** Draft
+**Authors:** jdheyburn
+
+---
+
+## Table of contents
+
+- [Overview](#overview)
+- [Goals](#goals)
+- [Non-Goals](#non-goals)
+- [Resource Hierarchy](#resource-hierarchy)
+- [Spec](#spec)
+  - [Design Philosophy](#design-philosophy)
+  - [Specification](#specification)
+- [Status](#status)
+  - [Conditions](#conditions)
+  - [`kubectl get valkeynode`](#kubectl-get-valkeynode)
+- [Reconciliation Logic](#reconciliation-logic)
+- [Naming Conventions](#naming-conventions)
+- [Lifecycle and Deletion Behaviour](#lifecycle-and-deletion-behaviour)
+- [Interaction Contract with Parent Controllers](#interaction-contract-with-parent-controllers)
+- [References](#references)
+
+---
+
+## Overview
+
+This document proposes a ValkeyNode CRD which will be used to abstract away the concrete implementation of how the Valkey server workloads are provisioned to Kubernetes.
+
+We wish to support both Deployments and StatefulSets, depending on the user requirements. Having an abstraction on top of these primitives will provide a common interface for the ValkeyCluster to abstract away this detail. We can use this opportunity to provide other abstractions that do not necessarily concern the parent controllers, such as management of volumes.
+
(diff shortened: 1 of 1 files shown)


docs/rfc/VALKEY_NODE_CRD_PROPOSAL.md
```diff
@@ -0,0 +1,274 @@
...
+
+```yaml
+apiVersion: valkey.io/v1alpha1
+kind: ValkeyNode
+metadata:
+  name: cache-0-0
+  namespace: default
+  ownerReferences: # Set by parent controller
+    - apiVersion: valkey.io/v1alpha1
+      kind: ValkeyCluster
```
sandeepkunusoth, line 103: will we embed ValkeyNode cr spec within ValkeyCluster? how this change will effect ValkeyCluster spec? An example of ValkeyCluster updated CRD would be helpful. if we are keeping it same as today, any new fields added to ValkeyNode spec needs to be duplicated to ValkeyCluster too.

jdheyburn: The ValkeyNode CR spec would not be integrated in ValkeyCluster, in the same way that StatefulSet/Deployment spec is not integrated in ValkeyCluster today. 

> any new fields added to ValkeyNode spec needs to be duplicated to ValkeyCluster too.

Yes, given that ValkeyCluster is the abstraction on top of ValkeyNode. The duplication I don't believe is an issue, since it would be better to consolidate the resources required of a Valkey Node in one controller which can then be reused in other Valkey CRDs

docs/rfc/VALKEY_NODE_CRD_PROPOSAL.md
```diff
@@ -0,0 +1,274 @@
...
+
+Parent controllers are responsible for executing Valkey commands (`REPLICAOF`, `CLUSTER MEET`, `CLUSTER ADDSLOTS`, `FAILOVER`, etc.) directly against the pod using `status.podIP`. ValkeyNode does **not** run topology or replication commands — it only observes the resulting state via `INFO replication` and reports it in status.
+
+This keeps ValkeyNode deliberately "dumb" and avoids coupling it to any specific topology model.
+
+### Change discovery
+
+Future extensibility includes the ability for parent controllers to set up **watches** on ValkeyNode resources.
+
+For example, when a ValkeyNode's status changes (e.g. pod IP changes, `observedRole` flips during a failover, or `ready` transitions), the watch triggers re-reconciliation of the parent controller, allowing it to react accordingly.
```
bjosv, line 269: I guess its ok initially to skip the `watches` if we keep the current periodic reconciliation in the parent controller. But hopefully we don't needed that periodic check after adding it though.

If we let this controller poll the Valkey instance via the valkey go-client and maybe get status & stats, we also get a quick feedback when there is a netsplit, or find out when a worker node is down.

docs/rfc/VALKEY_NODE_CRD_PROPOSAL.md
```diff
@@ -0,0 +1,274 @@
...
+
+We can add fields that would be beneficial for the user to query and filter on.
+
+- `observedRole` and `observedReplicaOf` are queried from `INFO replication`
+- During failovers, status may temporarily differ from desired state until reconciliation
+
+This can then plug into other interfaces as they can query the K8s API for the state of the Cluster and its ValkeyNodes.
+
+## Reconciliation Logic
+
```
bjosv, line 212: #63 mentions that a headless service should be created.
Should we let the parent controller create that now?
..or maybe that is a implementations specific question and we don't need to specify everything in this section.

bjosv (commented):
LGTM!

jdheyburn:
Closing this PR, but will aim to reference this for future discussions

PR #86, opened 2026-02-16
Handle post-failover pod replacement gracefully

ysqyang:
## Summary

When Valkey promotes a replica to primary during automatic failover and Kubernetes recreates the old node-index=0 pod, the reconciler previously tried to assign a new slot range to the replacement pod. This failed in a loop with `"no slots range to assign"` because all 16384 slots were already owned by the promoted replica, leaving the cluster stuck in `Degraded/NodeAddFailed`.

**Fix:** before assigning slots, `assignSlotsToPendingPrimaries` now checks the live Valkey topology (`shardExistsInTopology`) for any existing member in the same shard. If one exists (post-failover or mid-failover), the replacement pod joins as a replica instead. `replicateToShardPrimary` also gains a fallback (`findShardPrimary`) that scans all shard pods for the actual primary, since after failover node-index=0 is no longer the Valkey primary.

### What changed

- **`utils.go`** — two new helpers:
  - `shardHasLivePrimary`: checks if another pod in the same shard-index group is a slot-bearing primary in the live topology
  - `findShardPrimary`: scans all pods in a shard to find the actual Valkey primary regardless of node-index label
- **`valkeycluster_controller.go`**:
  - `addValkeyNode` step 2 now detects post-failover replacements and falls through to the replica path
  - `replicateToShardPrimary` tries node-index=0 first (fast path), then falls back to `findShardPrimary`

### Test plan

**1. Deploy a 3-shard, 1-replica cluster and wait for `Ready`**

```bash
```

Screenshot 2026-02-16 at 2 54 38 PM
Screenshot 2026-02-16 at 2 57 59 PM

**2. Kill shard-0 primary**

```bash
kubectl delete pod valkeycluster-sample-0-0
```

Screenshot 2026-02-16 at 3 04 55 PM

**3. Verify `valkeycluster-sample-0-1` is now the primary and `valkeycluster-sample-0-0` is its replica**

```bash
kubectl exec valkeycluster-sample-0-1 -- valkey-cli cluster nodes
```

Screenshot 2026-02-16 at 3 05 47 PM

### DIFF
--- a/internal/controller/utils.go
+++ b/internal/controller/utils.go
@@ -128,11 +128,11 @@ func podRoleAndShard(address string, pods *corev1.PodList) (string, int) {
 // assignSlotsToNewPrimary. Instead it should fall through to
 // replicateToShardPrimary, which will either succeed (case 1) or return an
 // error and retry on the next reconcile (case 2).
-func shardExistsInTopology(state *valkey.ClusterState, shardIndex int, selfAddress string, pods *corev1.PodList) bool {
+func shardExistsInTopology(state *valkey.ClusterState, shardIndex int, pods *corev1.PodList) bool {
 	si := strconv.Itoa(shardIndex)
 	for i := range pods.Items {
 		p := &pods.Items[i]
-		if p.Labels[LabelShardIndex] != si || p.Status.PodIP == selfAddress || p.Status.PodIP == "" {
+		if p.Labels[LabelShardIndex] != si || p.Status.PodIP == "" {
 			continue
 		}
 		for _, shard := range state.Shards {
@@ -151,11 +151,11 @@ func shardExistsInTopology(state *valkey.ClusterState, shardIndex int, selfAddre
 // primary, regardless of its node-index label. This handles the post-failover
 // case where node-index=1 (or higher) was promoted by Valkey.
 // Returns ("", "") if no primary is found.
-func findShardPrimary(state *valkey.ClusterState, shardIndex int, selfAddress string, pods *corev1.PodList) (nodeID, ip string) {
+func findShardPrimary(state *valkey.ClusterState, shardIndex int, pods *corev1.PodList) (nodeID, ip string) {
 	si := strconv.Itoa(shardIndex)
 	for i := range pods.Items {
 		p := &pods.Items[i]
-		if p.Labels[LabelShardIndex] != si || p.Status.PodIP == selfAddress || p.Status.PodIP == "" {
+		if p.Labels[LabelShardIndex] != si || p.Status.PodIP == "" {
 			continue
 		}
 		for _, shard := range state.Shards {
--- a/test/e2e/valkeycluster_test.go
+++ b/test/e2e/valkeycluster_test.go
@@ -24,8 +24,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
@@ -442,7 +440,7 @@ spec:
 
 		// This test was temporarily disabled in PR #54 because the operator
 		// could not recover from a primary deletion (issue #43). The failover
-		// fix (shardHasLivePrimary + findShardPrimary) now handles this: when
+		// fix (shardExistsInTopology + findShardPrimary) now handles this: when
 		// Valkey promotes the replica, the replacement node-index=0 pod joins
 		// as a replica of the promoted primary instead of trying to claim slots.
 		It("should detect and recover when a primary deployment is deleted", func() {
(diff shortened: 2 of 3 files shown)


test/e2e/valkeycluster_test.go
```diff
@@ -445,5 +438,114 @@ spec:
...
+				cmd := exec.Command("kubectl", "get", "deployments",
+					"-l", fmt.Sprintf("app.kubernetes.io/instance=%s,valkey.io/node-index=0", failoverClusterName),
+					"-o", "go-template={{ (index .items 0).metadata.name }}")
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred(), "Failed to find a primary deployment")
+				g.Expect(output).NotTo(BeEmpty())
+				primaryDeployment = output
+			}
+			Eventually(getPrimaryDeployment).Should(Succeed())
+
```
bjosv, line 493: When I debugged issue #43 I came to the conclusion that we need to make sure that the replica is in sync with the primary before we can delete the primary here, else the failover might not be a clean one (i.e. a flaky test).
Handling cluster failures is a separate issue and was not expected to be covered in this test.

For simplicity we can probably run
`cmd = exec.Command("kubectl", "exec", podName, "-c", "valkey-server", "--", "valkey-cli", "INFO", "REPLICATION")`
on all pods and check that no cmd-result contains `master_link_status:down`. Eventually all replicas are up and then we can delete a primary.

Maybe @sandeepkunusoth was already looking at this in #43 ?

ysqyang: great point!

internal/controller/utils.go
```diff
@@ -115,17 +116,56 @@ func podRoleAndShard(address string, pods *corev1.PodList) (string, int) {
...
+	}
+	return false
+}
+
+// findShardPrimary scans all pods with the given shard-index label and returns
+// the Valkey node ID + IP of whichever pod is currently the slot-bearing
+// primary, regardless of its node-index label. This handles the post-failover
+// case where node-index=1 (or higher) was promoted by Valkey.
+// Returns ("", "") if no primary is found.
+func findShardPrimary(state *valkey.ClusterState, shardIndex int, selfAddress string, pods *corev1.PodList) (nodeID, ip string) {
```
bjosv: Is the argument `selfAddress` really needed? It's an optimization to save a few cpu cycles right?
The api gets cleaner if we remove it.
Same comment for `shardExistsInTopology`

test/e2e/valkeycluster_test.go
```diff
@@ -445,5 +439,137 @@ spec:
 			}
 			Eventually(verifyClusterDeleted).Should(Succeed())
 		})
+
+		// This test was temporarily disabled in PR #54 because the operator
+		// could not recover from a primary deletion (issue #43). The failover
+		// fix (shardHasLivePrimary + findShardPrimary) now handles this: when
```
bjosv: The comment needs to be updated here, shardHasLivePrimary does not exist in this PR.

bjosv: ..also update the PR description with the latest changes, shardHasLivePrimary is mentioned there too.

test/e2e/valkeycluster_test.go
```diff
@@ -445,5 +439,137 @@ spec:
...
+				g.Expect(cr.Status.ReadyShards).To(Equal(int32(3)))
+			}
+			Eventually(verifyClusterReady).Should(Succeed())
+
+			By("verifying all replicas are fully synced before triggering failover")
+			verifyReplicationHealthy := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "pods",
+					"-l", fmt.Sprintf("app.kubernetes.io/instance=%s", failoverClusterName),
+					"-o", "go-template={{ range .items }}{{ .metadata.name }}{{ \"\\n\" }}{{ end }}")
+				output, err := utils.Run(cmd)
```
sandeepkunusoth: can u pull the changes from main branch i added change to check replication status before marking cluster ready. so i think we don't need to check this replication status in e2e tests for now.

internal/controller/utils.go
ysqyang: Good catch. I just realized that both functions are called with a pending node's address, which can't possibly appear in `state.Shards` (it wasn't part of any shard when the state was captured).

test/e2e/valkeycluster_test.go
ysqyang: done

test/e2e/valkeycluster_test.go
ysqyang: done

sandeepkunusoth: Hi @ysqyang is transient failure happening after removal of this?

PR #87, opened 2026-02-18
added replication check before marking cluster ready status

### LINKED ISSUE #43: ValkeyCluster Intermittently getting struck in Degraded after deployment deletion E2E Test

sandeepkunusoth:
Fixes #43 

Added replica status check before marking cluster ready status.

E2e tests for deleting primary is already covered as part of other PR @ysqyang So not duplicating it again.

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -198,6 +198,20 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
 
+	// Check that all replicas have their replication link up (master_link_status:up).
+	// before marking the cluster Ready, we need to make sure all replicas are in sync with their primary.
+	for _, shard := range state.Shards {
+		for _, node := range shard.Nodes {
+			if !node.IsReplicationInSync() {
+				log.V(1).Info("replica not yet in sync, requeue..", "address", node.Address)
+				setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonReconciling, "Waiting for replicas to sync with primary", metav1.ConditionFalse)
+				setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonReconciling, "Waiting for replica sync", metav1.ConditionTrue)
+				_ = r.updateStatus(ctx, cluster, state)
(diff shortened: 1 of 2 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -198,6 +198,20 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 
+	// Check that all replicas have their replication link up (master_link_status:up).
+	// before marking the cluster Ready, we need to make sure all replicas are in sync with their primary.
+	for _, shard := range state.Shards {
+		for _, node := range shard.Nodes {
+			if !node.IsReplicationInSync() {
+				log.V(1).Info("replica not yet in sync, requeue..", "address", node.Address)
+				setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonReconciling, "Waiting for replicas to sync with primary", metav1.ConditionFalse)
+				setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonReconciling, "Waiting for replica sync", metav1.ConditionTrue)
+				_ = r.updateStatus(ctx, cluster, state)
```
jdheyburn, line 209: Is it possible to do this instead?




```suggestion
				r.updateStatus(ctx, cluster, state)
```

bjosv: The linter will probably warn if we remove it, other than that it _should_ work (famous last words..)

sandeepkunusoth: yeah getting linter error

PR #90, opened 2026-02-19
Batch pending node processing across MEET, slot assignment, and replication

ysqyang:
## Summary

Refactors the pending-node reconciliation loop from a **one-node-per-reconcile** model into a **three-phase batched pipeline**, significantly reducing the number of reconcile cycles needed to form or scale a cluster.

### Before

Each reconcile processed **one** pending node: MEET it, requeue, assign slots or replicate, requeue. A fresh 3-shard / 1-replica cluster (6 nodes) required ~12 reconcile cycles just for node introduction.

### After

Pending nodes are processed in three batched phases per reconcile:

1. **Phase 1 – MEET**: All isolated nodes (`cluster_known_nodes <= 1`) are introduced to the cluster via `CLUSTER MEET` in a single pass with bidirectional MEET to prevent gossip fragmentation. Requeue once to let gossip propagate.
2. **Phase 2 – Assign slots**: All primary-labeled pending nodes receive their slot ranges via `CLUSTER ADDSLOTSRANGE` in one pass. Slot ranges are pre-computed and assigned sequentially. Post-failover replacement primaries are detected via `shardExistsInTopology` and skipped (they join as replicas in phase 3).
3. **Phase 3 – Replicate**: All replica-labeled pending nodes (and post-failover replacement primaries) are attached to their shard's primary via `CLUSTER REPLICATE` in one pass. Replicas whose primary isn't ready yet (e.g., during scale-out rebalancing) are silently skipped and retried on the next reconcile.

### Other changes

- **`pickPendingNode` removed** — no longer needed since primaries and replicas are handled in separate phases.
- **`SubtractSlotsRange` exported** (was `subtractSlotsRange`) so `assignSlotsToPendingPrimaries` can update remaining slot ranges after each assignment.
- **`errPrimaryNotReady` sentinel error** introduced so that `replicatePendingReplicas` can distinguish retriable failures (primary not yet in topology / gossip pending) from hard errors.
- **"Unknown node" errors from `CLUSTER REPLICATE`** (gossip hasn't propagated the primary's ID to the replica yet) are now treated as retriable rather than fatal.

## Test plan

- [ ] Unit tests pass (`make test`)
- [ ] E2E tests pass (cluster creation, replica deletion recovery, primary deletion/failover recovery)
- [ ] Manual test: apply a 4-shard/2-replica manifest and verify the cluster reaches Ready in fewer reconcile cycles than before

**Before** (n = 2):

Screenshot 2026-02-19 at 11 43 03 AM
Screenshot 2026-02-19 at 11 43 15 AM

**After** (n = 2):

Screenshot 2026-02-19 at 11 43 28 AM
Screenshot 2026-02-19 at 11 43 41 AM

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -139,11 +139,11 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// --- Phase 1: MEET all isolated nodes in one batch ---
 	// A node with cluster_known_nodes <= 1 hasn't been introduced to the
 	// cluster yet. CLUSTER MEET is idempotent and has no ordering
-	// dependencies, so we issue it for every isolated node in a single
-	// reconcile pass — this includes both pending nodes AND shard primaries
-	// that were assigned slots while still isolated. After MEET, we requeue
-	// to let gossip propagate before proceeding to slot assignment or
-	// replication.
+	// dependencies, so we issue it for every isolated pending node in a
+	// single reconcile pass. Phase 2 refuses to assign slots to isolated
+	// nodes, so every node is guaranteed to pass through here first.
+	// After MEET, we requeue to let gossip propagate before proceeding
+	// to slot assignment or replication.
 	{
 		met, err := r.meetIsolatedNodes(ctx, cluster, state)
 		if err != nil {
@@ -184,7 +184,7 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	// --- Phase 3: REPLICATE all replica-labeled pending nodes ---
-	// By this point all primaries have slots and appear in state.Shards.
+	// By this point all currently known primaries have slots and appear in state.Shards.
 	// CLUSTER REPLICATE for different replicas targets different primaries,
 	// so they can all be issued in one pass.
 	if len(state.PendingNodes) > 0 {
@@ -422,6 +422,25 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pod
 	return valkey.GetClusterState(ctx, ips, DefaultPort)
 }
 
+// findMeetTarget picks the best node to MEET all isolated nodes against.
+// Priority: (1) a non-isolated shard primary — already has slots, so gossip
+// will propagate slot info; (2) a non-isolated pending node from a previous
+// MEET batch; (3) the first isolated node as a bootstrap seed when every
+// single node is isolated (fresh bootstrap, first reconcile).
+func findMeetTarget(state *valkey.ClusterState, isolated []*valkey.NodeState) *valkey.NodeState {
+	for _, shard := range state.Shards {
+		if p := shard.GetPrimaryNode(); p != nil && !p.IsIsolated() {
+			return p
+		}
+	}
+	for _, node := range state.PendingNodes {
+		if !node.IsIsolated() {
+			return node
+		}
+	}
+	return isolated[0]
+}
+
 // meetIsolatedNodes issues CLUSTER MEET for every isolated pending node
 // (cluster_known_nodes <= 1). Phase 2 (assignSlotsToPendingPrimaries)
 // refuses to assign slots to isolated nodes, so every node is guaranteed
@@ -452,35 +471,11 @@ func (r *ValkeyClusterReconciler) meetIsolatedNodes(ctx context.Context, cluster
 		return 0, nil
 	}
 
-	// Find a well-connected node (cluster_known_nodes > 1) to use as the
-	// MEET target. We check shard primaries first (they already have slots,
-	// so gossip will propagate slot info), then fall back to any
-	// non-isolated pending node (from a previous MEET batch that hasn't
-	// received slots yet). Only if every single node is isolated (true
-	// fresh bootstrap, first reconcile) do we pick an isolated node as a
-	// bootstrap seed.
-	var meetTarget *valkey.NodeState
-	for _, shard := range state.Shards {
-		p := shard.GetPrimaryNode()
-		if p != nil && !p.IsIsolated() {
-			meetTarget = p
-			break
-		}
-	}
-	if meetTarget == nil {
-		for _, node := range state.PendingNodes {
-			if !node.IsIsolated() {
-				meetTarget = node
-				break
-			}
-		}
-	}
-
-	if meetTarget == nil {
-		// Every node is isolated (true fresh bootstrap). Use the first
-		// isolated node as a bootstrap seed: every other node MEETs it,
-		// and gossip takes care of full propagation.
-		meetTarget = isolated[0]
+	// Find a well-connected node to MEET all isolated nodes against.
+	// Falls back to an isolated node as a bootstrap seed if every node
+	// is isolated (fresh bootstrap).
+	meetTarget := findMeetTarget(state, isolated)
+	if meetTarget == isolated[0] {
 		isolated = isolated[1:]
 	}
 
@@ -619,8 +614,6 @@ func (r *ValkeyClusterReconciler) replicatePendingReplicas(ctx context.Context, 
 			}
 			log.V(1).Info("post-failover: attaching replacement node as replica",
 				"shardIndex", shardIndex, "node", node.Address)
-		} else if role != RoleReplica {
-			continue
 		}
 
 		if err := r.replicateToShardPrimary(ctx, cluster, state, node, shardIndex, pods); err != nil {


internal/controller/valkeycluster_controller.go
```diff
@@ -132,35 +137,71 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	// and the lookup would fail.
+	// --- Phase 1: MEET all isolated nodes in one batch ---
+	// A node with cluster_known_nodes <= 1 hasn't been introduced to the
+	// cluster yet. CLUSTER MEET is idempotent and has no ordering
+	// dependencies, so we issue it for every isolated node in a single
+	// reconcile pass — this includes both pending nodes AND shard primaries
+	// that were assigned slots while still isolated. After MEET, we requeue
+	// to let gossip propagate before proceeding to slot assignment or
+	// replication.
+	{
```
bjosv, line 147: Do we need the extra scope here?

internal/controller/valkeycluster_controller.go
```diff
@@ -382,126 +423,225 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pod
...
+//     > 1), use it as the target so new nodes join the existing cluster.
+//   - If all nodes are isolated (fresh bootstrap), use the first isolated
+//     node as a bootstrap seed. All others MEET this seed, and gossip
+//     propagates the full topology from there.
 //
-//  3. REPLICA: if node index is >= 1 (or a post-failover replacement), find
-//     the actual primary for the same shard and issue CLUSTER REPLICATE.
-func (r *ValkeyClusterReconciler) addValkeyNode(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, state *valkey.ClusterState, node *valkey.NodeState, pods *corev1.PodList) error {
+// Returns the number of nodes that were MEET'd.
+func (r *ValkeyClusterReconciler) meetIsolatedNodes(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, state *valkey.ClusterState, _ *corev1.PodList) (int, error) {
```
bjosv: The corev1.PodList argument is not used, should it be removed?

internal/controller/valkeycluster_controller.go
```diff
@@ -382,126 +423,225 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pod
...
+	sval, ok := node.ClusterInfo["cluster_known_nodes"]
+	if !ok {
+		return false
+	}
+	val, err := strconv.Atoi(sval)
+	return err == nil && val <= 1
+}
+
+// meetIsolatedNodes issues CLUSTER MEET for every isolated node in the
+// cluster — both pending nodes (no slots) AND shard primaries that were
```
bjosv: oh, interesting. So we can assign slots before we meet. Is that when the first, and sofar only, node is a primary? 
I guess it works since we don't have overlapping slots?  We are not currently setting a specific unique epoch for each created node but that is recommended, not sure if that effects this.

internal/controller/valkeycluster_controller.go
ysqyang: This isn’t strictly required; it’s a stylistic choice to keep `err` scoped consistently across all three phases. Phases 2 and 3 already achieve this naturally as top-level if blocks.

internal/controller/valkeycluster_controller.go
ysqyang: It should be removed. I used it in an earlier iteration and temporarily left it in as a stub while moving toward the current version, but forgot to remove it entirely. Good catch!

internal/controller/valkeycluster_controller.go
ysqyang: Great question! This was actually for a corner case where exactly one pod is up in the first reconcile and the cluster would end up fragmented (I actually ran into this during local testing). But I can actually make the whole thing cleaner by not allowing slot assignment to isolated nodes.

internal/controller/valkeycluster_controller.go
```diff
@@ -382,126 +423,229 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pod
...
 }
 
-// addValkeyNode introduces a pending Valkey node into the cluster. The node's
-// intended role is derived from its pod labels (e.g. shard-index=0, node-index=0
-// for the initial primary, shard-index=1, node-index=1 for a replica), which
-// was set at Deployment-creation time by upsertDeployments. This removes all
-// guesswork from the reconciler:
+// isNodeIsolated returns true if the node's cluster_known_nodes is <= 1,
+// meaning it hasn't been introduced to any other cluster member yet.
+func isNodeIsolated(node *valkey.NodeState) bool {
```
sandeepkunusoth: can we move this function to clusterstate

ysqyang: done

internal/controller/valkeycluster_controller.go
```diff
@@ -382,126 +422,218 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pod
...
-	if role == RoleReplica {
-		return r.replicateToShardPrimary(ctx, cluster, state, node, shardIndex, pods)
+	if meetTarget == nil {
+		for _, node := range state.PendingNodes {
+			if !node.IsIsolated() {
+				meetTarget = node
+				break
+			}
+		}
 	}
```
jdheyburn: Could you extract this into its own function? Then you can avoid the breaks in the for loops and simplify slightly.

```go
func getMeetTarget() *valkey.NodeState {
	for _, shard := range state.Shards {
		node := shard.GetPrimaryNode()
		if node != nil && !node.IsIsolated() {
			return node
		}
	}
	for _, node := range state.PendingNodes {
		if !node.IsIsolated() {
			return node
		}
	}
    return nil
```

internal/controller/valkeycluster_controller.go
```diff
@@ -382,126 +422,218 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pod
...
+		// assignSlotsToPendingPrimaries.
+		if role == RolePrimary {
+			if !shardExistsInTopology(state, shardIndex, pods) {
+				continue // genuine new primary, handled by assignSlotsToPendingPrimaries
+			}
+			log.V(1).Info("post-failover: attaching replacement node as replica",
+				"shardIndex", shardIndex, "node", node.Address)
+		} else if role != RoleReplica {
+			continue
+		}
```
jdheyburn: Wouldn't `role != RoleReplica` mean that the role is primary, which is already covered in line 616?

internal/controller/valkeycluster_controller.go
```diff
@@ -132,35 +136,71 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	// This is important because replicateToShardPrimary needs the primary
-	// to already be in state.Shards (i.e. have slots assigned). If we
-	// processed a replica first, its primary might still be in PendingNodes
-	// and the lookup would fail.
+	// --- Phase 1: MEET all isolated nodes in one batch ---
+	// A node with cluster_known_nodes <= 1 hasn't been introduced to the
+	// cluster yet. CLUSTER MEET is idempotent and has no ordering
+	// dependencies, so we issue it for every isolated node in a single
+	// reconcile pass — this includes both pending nodes AND shard primaries
+	// that were assigned slots while still isolated. After MEET, we requeue
```
bjosv: This can be removed now " AND shard primaries that were assigned slots while still isolated."

internal/controller/valkeycluster_controller.go
```diff
@@ -132,35 +136,71 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonReconciling, "Cluster is Reconciling", metav1.ConditionFalse)
+			_ = r.updateStatus(ctx, cluster, state)
+			// Requeue so that the next reconcile sees the newly-created shards
+			// in state.Shards, which replicas need for their lookup.
+			return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+		}
+	}
+
+	// --- Phase 3: REPLICATE all replica-labeled pending nodes ---
+	// By this point all primaries have slots and appear in state.Shards.
```
bjosv:
```suggestion
	// By this point all currently known primaries have slots and appear in state.Shards.
```

internal/controller/valkeycluster_controller.go
bjosv: Is this when podRoleAndShard don't find a shardIndex- or nodeIndex-label?
Nowadays we should have added them in he pod template while creating the deployment.
If this is to catch some inconsistency we probably should log an error here.

ysqyang: `role` comes from the helper function `podRoleAndShard` which can return an empty role under some corner cases. But with the way we use the helper today, such corner cases won't happen. So we can skip the `role != RoleReplica` check entirely.

internal/controller/valkeycluster_controller.go
ysqyang: done

PR #91, opened 2026-02-23
Added Initial TLS support for ValkeyCluster

### LINKED ISSUE #59: (feat) TLS support for ValkeyCluster
## **Summary**

Add TLS support for Valkey nodes managed by valkey-operator, including:

* Enable TLS for **client traffic**, **replication traffic**, and **cluster bus**.
* Support both **server-only TLS** and **mutual TLS (mTLS)**.
* Optional **mTLS** (require client certs).
* Existing clusters without TLS must behave unchanged (backwards compatible).
* Update **status conditions** for missing/invalid TLS secrets.

**Followups (optional)**

* Auto-generate certificates via **cert-manager**
* Certificate rotation automation

---

## **Proposed CRD Spec Changes**

Add an optional `spec.tls`:

```yaml
kind: ValkeyCluster
metadata:
  name: valkeycluster-tls
spec:
  shards: 3
  replicas: 0
  tls:
    # Enable TLS
    enabled: false

    # Name of the Secret containing TLS keys (required when enabled=true)
    existingSecret: ""

    # Secret key name containing server public certificate (default=server.crt)
    cert: server.crt

    # Secret key name containing server private key (default=server.key)
    key: server.key

    # Secret key name containing Certificate Authority public certificate (default=ca.crt)
    ca: ca.crt

    # Require that clients authenticate with a certificate (mTLS)
    requireClientCertificate: false

    # Future mode: cert-manager (optional, follow-up) issuerRef:
    issuerRef:
          name: my-issuer
          kind: ClusterIssuer
```

### Validation

* Backwards compatible: `spec.tls` omitted => current behavior.
* If `tls.enabled=true`:

  * Must specify **either** `existingSecret` **or** `issuerRef` (not both).
  * If `existingSecret` is used:

    * Secret must contain keys: `<cert>`, `<key>`, `<ca>` (using the configured names).
* Defaults:

  * `cert=server.crt`, `key=server.key`, `ca=ca.crt`
  * `requireClientCertificate=false`

---

### Backwards compatibility

* [ ] If `spec.tls` is omitted or `tls.enabled=false`, behavior is unchanged.

### TLS (server-only)

* [ ] With `tls.enabled=true` and `requireClientCertificate=false`, pods become Ready.
* [ ] TLS client connections succeed (PING/SET/GET via TLS).
* [ ] Replication + cluster bus use TLS (where applicable).

### mTLS

* [ ] With `requireClientCertificate=true`, clients without cert are rejected.
* [ ] Clients with valid cert can connect and perform basic ops.

### Status + validation

* [ ] Missing/invalid secret surfaces a clear status condition + event.
* [ ] Missing keys in secret surfaces which keys are missing.

---

## Follow-ups (optional)

* cert-manager integration:

  * If `issuerRef` set, operator creates `Certificate` resources and waits for Secret.
* Certificate rotation automation:

  * Controlled rolling restart when Secret changes (or documented behavior).

Contribution:
- [X] I’m willing to implement this feature.

sandeepkunusoth:
This PR adds initial TLS support to the Valkey Cluster, fixes #59.  TLS enabled for ValkeyCluster using pre-created / existing Kubernetes Secrets. Updated e2e tests to test TLS.

Followup features not included in this PR:
- Automatic certificate provisioning using cert-manager
- Certificate rotation on renewal.


internal/controller/utils.go
```diff
@@ -181,3 +209,39 @@ func pickPendingNode(nodes []*valkey.NodeState, pods *corev1.PodList) *valkey.No
...
+		return nil, fmt.Errorf("failed to parse TLS key pair: %w", err)
+	}
+
+	caCertPool := x509.NewCertPool()
+	caCertPool.AppendCertsFromPEM(caData)
+
+	return &tls.Config{
+		Certificates:       []tls.Certificate{cert},
+		RootCAs:            caCertPool,
+		InsecureSkipVerify: true,
```
bjosv, line 254: I guess we aim to remove InsecureSkipVerify soon? Or should it be configurable?

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,41 @@ type ValkeyClusterSpec struct {
...
+}
+
+// TLSConfig defines the TLS configuration for a Valkey cluster.
+type TLSConfig struct {
+	// Enable TLS
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled,omitempty"`
+
+	// Name of the Secret containing TLS keys
+	ExistingSecret string `json:"existingSecret,omitempty"`
```
bjosv, line 94: Since we use the name "existing" here, will the cert-manager create a separate secret which it updates?
Or it named so that a user should know that is needs to exist?

internal/controller/deployment.go
```diff
@@ -31,15 +31,60 @@ func generateContainersDef(cluster *valkeyiov1alpha1.ValkeyCluster) []corev1.Con
...
+			"--tls-cluster", "yes",
+			"--tls-replication", "yes",
+			"--tls-cert-file", certPath,
+			"--tls-key-file", keyPath,
+			"--tls-ca-cert-file", caPath,
+		}
+
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  "VALKEY_TLS_ARGS",
+			Value: fmt.Sprintf("--tls --cert %s --key %s --cacert %s", certPath, keyPath, caPath),
```
bjosv, line 71: Ah, we don't need `--insecure` here but we have InsecureSkipVerify=true?

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: maybe we can change ExistingSecret to secretRef so that it can be used for both exisiting secret or secret generated by Certificate CRD.

internal/controller/utils.go
sandeepkunusoth: i tried removing it. getting connection errors due to cert validation will check on this. the same works from valkey-cli but go client throws error

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,41 @@ type ValkeyClusterSpec struct {
...
+	// TLS configuration
+	// +optional
+	TLS *TLSConfig `json:"tls,omitempty"`
+}
+
+// TLSConfig defines the TLS configuration for a Valkey cluster.
+type TLSConfig struct {
+	// Enable TLS
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled,omitempty"`
```
jdheyburn, line 91: Could we remove the `enabled` field and use the presence of `tls` as whether it should be enabled?

internal/controller/utils.go
```diff
@@ -17,15 +17,40 @@ limitations under the License.
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	valkeyv1 "valkey.io/valkey-operator/api/v1alpha1"
 	"valkey.io/valkey-operator/internal/valkey"
 )
 
+// getTLSFileNames returns the parsed cert, key, and ca file names for a Valkey cluster.
+func getTLSFileNames(tlsConfig *valkeyv1.TLSConfig) (string, string, string) {
+	if tlsConfig == nil || !tlsConfig.Enabled {
+		return "", "", ""
+	}
```
jdheyburn, line 39: Would this block ever be hit? Seems we're always checking for this before we call `getTLSFileNames`

internal/controller/valkeycluster_controller.go
```diff
@@ -229,6 +235,35 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+// upsertTLSCertificate will create or update the TLS certificate for the cluster
+// It will return the existing secret name if it exists, otherwise it should create a new Certificate
+// TODO create Certificate CR for server and client based on issuerRef specified in the cluster spec
+func (r *ValkeyClusterReconciler) upsertTLSCertificate(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+	if cluster.Spec.TLS == nil || !cluster.Spec.TLS.Enabled {
+		return nil
+	}
+
+	if cluster.Spec.TLS.ExistingSecret == "" {
+		// TODO: cert-manager integration
```
jdheyburn, line 293: How common is it for operators to depend on cert-manager? I checked ElasticSearch and CloudNativePG and they don't do this. Instead they self-manage their own certs, while allowing the user to override this by providing their own certs (possibly managed by cert-manager).

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,41 @@ type ValkeyClusterSpec struct {
...
+	// +kubebuilder:default=server.crt
+	Cert string `json:"cert,omitempty"`
+
+	// Secret key name containing server private key (default=server.key)
+	// +kubebuilder:default=server.key
+	Key string `json:"key,omitempty"`
+
+	// Secret key name containing Certificate Authority public certificate (default=ca.crt)
+	// +kubebuilder:default=ca.crt
+	CA string `json:"ca,omitempty"`
```
jdheyburn, line 106: Are these defaults standard with the `kubernetes.io/tls` Secret type?

```
tls.crt
tls.key
ca.crt
```

api/v1alpha1/valkeycluster_types.go
jdheyburn: Could we name it `certificateRef` to provide a clearer name of intent? And have it be of a custom struct so that we can extend on it later on if we wanted to? This is what the elasticsearch operator does:

```
type SecretRef struct {
    name string `json:"name,omitempty"`
}
```

api/v1alpha1/valkeycluster_types.go
```diff
@@ -73,6 +73,41 @@ type ValkeyClusterSpec struct {
...
+
+	// TODO: cert-manager integration and rotation
+	// IssuerRef references a cert-manager Issuer or ClusterIssuer.
+	IssuerRef *IssuerRef `json:"issuerRef,omitempty"`
```
jdheyburn, line 110: Similar comment here re cert-manager integration

internal/controller/valkeycluster_controller.go
```diff
@@ -229,6 +235,35 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		err := errors.New("tls.existingSecret is required when tls.enabled is true")
```
jdheyburn, line 294: I think we should remove `tls.enabled`, and have presence on `tls.certificateRef` (renamed from `tls.existingSecret`) to determine whether TLS should be enabled. Otherwise `tls.enabled` isn't really doing a lot here. Plus it can be confusing to the user: "which part of TLS is enabled?"

sandeepkunusoth:
> This PR adds initial TLS support to the Valkey Cluster, fixes #59. TLS enabled for ValkeyCluster using pre-created / existing Kubernetes Secrets. Updated e2e tests to test TLS.
> 
> Followup features not included in this PR:
> 
> * Automatic certificate provisioning using cert-manager
> * Certificate rotation on renewal.

will open new PR with these comments addressed. closing this PR as there is some problem when rebasing

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: @jdheyburn i was trying to follow structure from valkey-helm https://github.com/valkey-io/valkey-helm/blob/e0a84af747e7684d656064b3809c53615ffc4bc4/valkey/values.yaml#L265

we can remove if not required.

internal/controller/valkeycluster_controller.go
sandeepkunusoth: hi  @jdheyburn,

in tls enabled there can be 2 cases. 

user must specify either existingSecret or issuerRef (not both).

1. if existingSecret  is not empty, 1 user wants to pass existing secrets they are managing by themself.
or 
2. if they want to pass cert issuer settings, we could automatically create Certificate CRD within valkey operator. In this case user doesnt have to create certificate manually. but for now i can keep it simple with only 1st use case

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: @jdheyburn if we are renaming it to certificateRef are we expecting users to pass certificate name here to create certificate CR within operator? or use exisitng certificate? i think passing secret name would work for existing secret. if user doesnt want to use cert-manager to manage certificates?

PR #92, opened 2026-02-24
Fix slot-range parsing during migration and add rebalance planning logic

ysqyang:
## Summary

- Fix `parseSlotsRanges` to skip `[slot->-nodeid]` / `[slot-<-nodeid]` entries that Valkey appends to `CLUSTER NODES` output during active slot migrations, which previously caused parse errors.
- Add `BuildRebalanceMove` — a deterministic, incremental slot rebalance planner that computes a single slot migration from an overloaded primary to an underloaded (or empty) one. This is the planning layer for scale-out support; execution is wired in a follow-up PR.

## Test plan

- [x] Unit test for `parseSlotsRanges`: normal ranges, migrating/importing entries, empty input, migration-only input
- [x] Unit tests for `PlanRebalanceMove`: scale-out (2→3 shards), balanced cluster (no-op), mismatched shard count, extra empty shards, zero max-slots, slot extraction from ranges


internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+
+	moveCount := min(src.count-src.target, dst.target-dst.count, maxSlots)
+	slots, err := takeSlotsFromRanges(src.ranges, moveCount)
+	if err != nil {
+		return nil, err
+	}
+	return &SlotMove{Src: src.node, Dst: dst.node, Slots: slots}, nil
+}
+
+func takeSlotsFromRanges(ranges []SlotsRange, count int) ([]int, error) {
```
jdheyburn: Can you use a more descriptive variable name than `count`, what is being counted? numSlotsToMove?

internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// Tolerate ±1-slot rounding differences that arise because Phase 2's
+	// slot assignment order may not match the address-sorted target order.
+	if src.count-src.target <= 1 && dst.target-dst.count <= 1 {
+		return nil, nil
+	}
+
+	moveCount := min(src.count-src.target, dst.target-dst.count, maxSlots)
+	slots, err := takeSlotsFromRanges(src.ranges, moveCount)
```
jdheyburn: Similar to below comment, I think `numSlotsToMove` might be a more descriptive variable name over moveCount (feel free to pick a more appropriate name).

internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+type SlotMove struct {
+	Src   *NodeState
+	Dst   *NodeState
+	Slots []int
+}
+
+type primarySlots struct {
+	node   *NodeState
+	ranges []SlotsRange
+	count  int
```
jdheyburn: What are we counting here? Can a more descriptive name be used?

internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+	target int
```
jdheyburn: What is target in this context?

internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+			dst = p
+		}
+	}
+	if src == nil || dst == nil {
+		return nil, nil
+	}
+
+	// Tolerate ±1-slot rounding differences that arise because Phase 2's
+	// slot assignment order may not match the address-sorted target order.
+	if src.count-src.target <= 1 && dst.target-dst.count <= 1 {
```
jdheyburn: Could this be a boolean function so that it has a descriptive name?

internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+// It returns nil when the cluster is already balanced or not ready for rebalancing.
+func BuildRebalanceMove(shards []*ShardState, expectedShards int, maxSlots int) (*SlotMove, error) {
+	if expectedShards <= 0 || maxSlots <= 0 {
+		return nil, nil
+	}
+	if len(shards) != expectedShards {
+		return nil, nil
+	}
+
+	all := make([]*primarySlots, 0, len(shards))
```
jdheyburn: Could we use a more descriptive name than `all`? `slotAllocations` maybe?

internal/valkey/rebalance.go
```diff
@@ -0,0 +1,117 @@
...
+			count:  count,
+		})
+	}
+	// Sort by address for deterministic target assignment.
+	sort.Slice(all, func(i, j int) bool { return all[i].node.Address < all[j].node.Address })
+
+	// Assign per-shard slot targets and find the first imbalance.
+	base, rem := TotalSlots/expectedShards, TotalSlots%expectedShards
+	var src, dst *primarySlots
+	for i, p := range all {
```
jdheyburn: Can we use more descriptive variable names over `i` and `p`?

internal/controller/valkeycluster_controller.go
```diff
@@ -553,15 +553,13 @@ func (r *ValkeyClusterReconciler) assignSlotsToPendingPrimaries(ctx context.Cont
...
-		shardsLeft := shardsRequired - (shardsExists + assigned)
-		if shardsLeft == 1 {
-			// Last shard: absorb everything that remains.
-			if len(slots) != 1 {
-				return assigned, errors.New("assigning multiple ranges to shard not yet supported")
-			}
-			slotEnd = slots[0].End
+		base := valkey.TotalSlots / shardsRequired
+		idx := shardsExists + assigned
+		perShard := base
```
bjosv: Do we need `base`?
We can set it directly  to `valkey.TotalSlots / shardsRequired` and rename it to `slotsPerShard`?

internal/controller/valkeycluster_controller.go
```diff
@@ -553,15 +553,13 @@ func (r *ValkeyClusterReconciler) assignSlotsToPendingPrimaries(ctx context.Cont
...
+		if idx < valkey.TotalSlots%shardsRequired {
+			perShard++
```
bjosv: We could add a comment that we distribute the remaining slots to the first N shards.

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,117 @@
+/*
```
bjosv, line 1: Since we might use this directory, ie, the valkey package, for other CRD controllers like sentinel or so, should we rename this file to cluster_rebalance.go? (an open question..)

internal/valkey/rebalance.go
ysqyang: done

internal/valkey/rebalance.go
ysqyang: done

internal/valkey/rebalance.go
ysqyang: done

internal/valkey/rebalance.go
ysqyang: done

internal/valkey/rebalance.go
ysqyang: done

internal/valkey/rebalance.go
ysqyang: renamed

internal/valkey/rebalance.go
ysqyang: Not sure if it's worth making this a helper function given that it's just a simple arithmetic operation

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	// Tolerate ±1-slot rounding differences that arise because Phase 2's
+	// slot assignment order may not match the address-sorted target order.
+	if src.numSlots-src.targetNumSlots <= 1 && dst.targetNumSlots-dst.numSlots <= 1 {
+		return nil, nil
+	}
+
+	numSlotsToMove := min(src.numSlots-src.targetNumSlots, dst.targetNumSlots-dst.numSlots, maxSlots)
+	slots := takeSlotsFromRanges(src.ranges, numSlotsToMove)
```
bjosv, line 89: `takeSlotsFromRanges` can return nil; should that trigger a error (or a `return nil, nil`) instead of a `SlotMove{Slots: nil}`?
Maybe we should check the result of numSlotsToMove instead?
takeSlotsFromRanges is a package internal helper that could expect valid input.

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,111 @@
...
+type primarySlots struct {
+	node           *NodeState
+	ranges         []SlotsRange
+	numSlots       int
+	targetNumSlots int
+}
+
+// BuildRebalanceMove computes a single, deterministic slot move to improve balance.
+// It returns nil when the cluster is already balanced or not ready for rebalancing.
+func BuildRebalanceMove(shards []*ShardState, expectedShards int, maxSlots int) (*SlotMove, error) {
```
bjosv: We build a `SlotMove` in this function, is the name of this function misguiding?
Maybe its a PlanRebalanceMove?

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+	"sort"
+)
+
+// SlotMove describes a small, incremental slot migration between two primaries.
+type SlotMove struct {
+	Src   *NodeState
+	Dst   *NodeState
+	Slots []int
```
bjosv, line 28: Not for this PR; but maybe we should have a defined type for a Slot in clusterstate.go (or alias).
That might avoid any confusion if an `int` means a slot, a number of slots, a length or size.

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,111 @@
...
+			node:     primary,
+			ranges:   append([]SlotsRange(nil), shard.Slots...),
+			numSlots: numSlots,
+		})
+	}
+	// Sort by address for deterministic target assignment.
+	sort.Slice(slotAllocations, func(i, j int) bool { return slotAllocations[i].node.Address < slotAllocations[j].node.Address })
+
+	// Assign per-shard slot targets and find the first imbalance.
+	baseSlotsPerShard, rem := TotalSlots/expectedShards, TotalSlots%expectedShards
```
bjosv: In valkeycluster_controller.go‎ we changed base to numSlotsPerShard

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		if dst == nil && alloc.numSlots < alloc.targetNumSlots {
+			dst = alloc
+		}
+	}
+	if src == nil || dst == nil {
+		return nil, nil
+	}
+
+	// Tolerate ±1-slot rounding differences that arise because Phase 2's
```
bjosv: Hum, maybe we should remove the mention of Phase 2 here since its not stated what it is in this file.
It's the initial slot assignment right?

internal/valkey/cluster_rebalance.go
bjosv: also, the `shards []*ShardState`, is that from the current ClusterState, or is it planned to be a subset?

internal/valkey/cluster_rebalance.go
```diff
@@ -0,0 +1,111 @@
...
+	// slot assignment order may not match the address-sorted target order.
+	if src.numSlots-src.targetNumSlots <= 1 && dst.targetNumSlots-dst.numSlots <= 1 {
```
bjosv: Isn't there a risk we get the same src and dst when calling the BuildRebalanceMove repeatedly,
i.e. we will not get a better rebalancing suggestion next time either?
Could that be mitigated by having his check in the loop over slotAllocations above somehow, and skip them there.
..or maybe I missed something and just need more coffee :)

internal/valkey/rebalance.go
jdheyburn: It may be simple, but me reading this as a layman is hard to grasp what its checking for. I'm a fan of including functions if it improves readability for contributors. It's not anything blocking though. :)

internal/valkey/cluster_rebalance.go
ysqyang: > also, the `shards []*ShardState`, is that from the current ClusterState, or is it planned to be a subset?

It's actually a superset of current ClusterState, which will become apparent in a follow-up PR.

internal/valkey/cluster_rebalance.go
ysqyang: This logic can be refined indeed:)

internal/valkey/cluster_rebalance.go
ysqyang: renamed

internal/valkey/rebalance.go
ysqyang: the whole rounding check can be refined, which hopefully eliminates the need for a helper function:)

PR #97, opened 2026-02-27
chore(kubebuilder): Upgrade the Scaffold: v4.11.1 -> v4.13.0

sandeepkunusoth:
fixes https://github.com/valkey-io/valkey-operator/issues/95

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,9 @@
 name: Lint
 
 on:
+  push:
+    branches:
+      - 'main'
   pull_request:
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -17,7 +16,9 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Check linter configuration
+        run: make lint-config
+
       - name: Run linter
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20  # v9.2.0
```
bjosv: Remove the `uses:` line too

.github/workflows/test-e2e.yml
```diff
@@ -1,7 +1,6 @@
 name: E2E Tests
 
 on:
-  push:
```
bjosv, line 4: I think its nice if we run CI after merging to main, i.e. to know we have a stable base.
Wouldn't this just stop that?

sandeepkunusoth: updated to run only on main

PR #100, opened 2026-02-27
Add scale-out support with slot rebalancing

ysqyang:
### Summary

- Enables scaling out a `ValkeyCluster` by increasing `spec.shards`. New shard deployments are created automatically, and hash slots are incrementally migrated from existing primaries to new ones.
- Slot migration supports two strategies: `CLUSTER MIGRATESLOTS` for atomic batch migration on newer Valkey versions, with a per-key `MIGRATE` fallback for older versions.
- Each reconcile loop moves a bounded batch of slots (default 400) so the operator stays responsive and restartable. Rebalancing converges over multiple reconcile cycles.

### Key changes

**New file — `internal/controller/migration.go`:**
- `rebalanceSlots`: orchestrates `PlanRebalanceMove` to compute a single bounded move, checks for in-progress migrations via `CLUSTER GETSLOTMIGRATIONS`, then executes via `migrateSlotsAtomic` or `migrateSlotLegacy` fallback.
- `slotMigrationInProgress` / `nodeKnownToSource`: pre-migration safety checks (active migration detection, gossip propagation verification).
- `slotsToRanges` / `countSlots`: slot range conversion and counting utilities.
- `isAtomicMigrationUnsupported` / `isSlotsNotServedByNode`: error classifiers for migration edge cases.

**`internal/controller/valkeycluster_controller.go`:**
- `effectiveShards`: includes slot-less pending primaries in health checks and rebalancing decisions, since `GetClusterState` classifies new primaries as `PendingNodes` until they own slots.
- Reconcile loop now calls `rebalanceSlots` after health checks pass.
- `assignSlotsToPendingPrimaries` gracefully returns 0 when no unassigned slots are available (scale-out case), allowing the rebalancer to redistribute slots instead.

**`internal/controller/utils.go`:**
- `findShardPrimary` skips shards with no slots, preventing replicas from being attached to primaries that haven't received slots yet during scale-out.
- `countSlots` helper for slot range counting.

**`api/v1alpha1/valkeycluster_types.go`:**
- New status reasons: `RebalancingSlots`, `RebalanceFailed`.

**`test/e2e/valkeycluster_test.go`:**
- New test: creates a 2-shard cluster, scales to 3 shards, verifies all 3 primaries own slots and the cluster reaches Ready state.

### Test plan

- [ ] Unit tests pass (`go test ./internal/valkey/...`)
- [ ] E2E: scale out from 2 to 3 shards — verify slot redistribution and Ready state
- [ ] E2E: scale out from 3 to 5 shards — verify rebalancing converges
- [ ] Verify no unnecessary rebalancing during initial cluster creation (±1 tolerance)
- [ ] Verify replicas are not attached until their primary receives slots


internal/valkey/cluster_migration.go
```diff
@@ -0,0 +1,222 @@
+/*
+Copyright 2025 Valkey Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
```
sandeepkunusoth, line 6: can we move this under valkey folder and rename it to cluster_migration.go?

internal/valkey/cluster_migration.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return nil
+}
+
+func migrateSlotLegacy(ctx context.Context, src *valkey.NodeState, dst *valkey.NodeState, slot int) error {
+	if err := src.Client.Do(ctx, src.Client.B().ClusterSetslot().Slot(int64(slot)).Migrating().NodeId(dst.Id).Build()).Error(); err != nil {
+		return fmt.Errorf("setslot migrating failed for slot %d on %s: %w", slot, src.Address, err)
+	}
+	if err := dst.Client.Do(ctx, dst.Client.B().ClusterSetslot().Slot(int64(slot)).Importing().NodeId(src.Id).Build()).Error(); err != nil {
+		return fmt.Errorf("setslot importing failed for slot %d on %s: %w", slot, dst.Address, err)
```
sandeepkunusoth: just a query. what happens when these slot migrations fail? does it resume on next reconcile?  can the particular slot be struck in migrating state?

test/e2e/valkeycluster_test.go
```diff
@@ -310,6 +313,98 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+
+			By(fmt.Sprintf("waiting for the cluster to report %d ready shards", scaleOutShards))
+			verifyScaledOut := func(g Gomega) {
+				cr, err := utils.GetValkeyClusterStatus(valkeyClusterName)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(cr.Status.Shards).To(Equal(int32(scaleOutShards)))
+				g.Expect(cr.Status.ReadyShards).To(Equal(int32(scaleOutShards)))
+				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+			}
+			Eventually(verifyScaledOut, 10*time.Minute, 2*time.Second).Should(Succeed())
```
sandeepkunusoth: can we remove hardcoded timeouts these are already present in common e2e_suite 10*time.Minute, 2*time.Second

internal/controller/valkeycluster_controller.go
```diff
@@ -779,6 +811,23 @@ func (r *ValkeyClusterReconciler) countReadyShards(state *valkey.ClusterState, c
...
+// effectiveShards returns state.Shards plus any pending primaries that are
+// scale-out leaders (slot-less Valkey masters labeled as RolePrimary).
+func effectiveShards(state *valkey.ClusterState, pods *corev1.PodList) []*valkey.ShardState {
+	shards := append([]*valkey.ShardState(nil), state.Shards...)
+	for _, node := range state.PendingNodes {
+		role, _ := podRoleAndShard(node.Address, pods)
+		if role == RolePrimary {
+			shards = append(shards, &valkey.ShardState{
+				Id:        node.ShardId,
+				Nodes:     []*valkey.NodeState{node},
```
sandeepkunusoth, line 871: is there any case where this can result in duplicate shard entry?

test/e2e/valkeycluster_test.go
ysqyang: The default timeout is `2*time.Minute` which might be too short for resharding. I can change the polling interval to the default of `1*time.Second`.

internal/controller/valkeycluster_controller.go
ysqyang: Yes, there can be, but it's benign. Consider `node-0-0` failing and coming back after a successful failover to `node-0-1`. The `ClusterState` now has `node-0-0` as a pending node (because it comes back as an empty master). It goes through the 3 phases as usual: `meetIsolatedNodes`, `assignSlotsToPendingPrimaries` (a no-op because of the `shardExistsInTopology` check), and finally `replicatePendingReplicas` (which issues `CLUSTER REPLICATE` on node-0-0 to replicate node-0-1). In most cases we requeue before hitting `effectiveShards`, and `node-0-0` will no longer be a pending node in the next reconcile — so no duplicate shard entry is created. The only case a duplicate entry would appear is when `CLUSTER REPLICATE` fails because `node-0-0` doesn't know about `node-0-1` yet (gossip delay). But this temporary inconsistency is caught immediately by the replica-count check (the real shard is missing a node), so the duplicate never makes its way downstream. It resolves itself in subsequent reconciles once gossip propagates.

internal/valkey/cluster_migration.go
ysqyang: done

internal/valkey/cluster_migration.go
ysqyang: If the migration fails partway, the slot can be left in `MIGRATING` state on the source. On the next reconcile, `PlanRebalanceMove` computes the same imbalance and retries the same slot. This effectively resumes: `SETSLOT MIGRATING/IMPORTING` are idempotent for the same destination, and `GETKEYSINSLOT` returns only remaining keys, so it picks up where it left off.

A slot can get stuck in `MIGRATING` state if the same step fails persistently (e.g. destination unreachable), but it resolves once the root cause clears. We don't currently issue `CLUSTER SETSLOT STABLE` to roll back.

I'm inclined to keep it this way for now since this is the legacy migration path (Valkey < 9.0). I'll create a poll in the Slack channel on whether we should only support Valkey >= 9.0 — if the majority are in favor, we can drop this code path entirely in favor of the atomic `CLUSTER MIGRATESLOTS`.

jdheyburn: @ysqyang Would it make implementation easier if we only supported 9.0+, then add support for <9.0 in another PR in the future?

internal/controller/valkeycluster_controller.go
```diff
@@ -779,6 +811,76 @@ func (r *ValkeyClusterReconciler) countReadyShards(state *valkey.ClusterState, c
 	return readyCount
 }
 
+const rebalanceSlotBatchSize = 400
```
jdheyburn, line 821: What is the significance of this?

internal/valkey/cluster_migration.go
```diff
@@ -0,0 +1,182 @@
...
+	"sort"
+	"strconv"
+	"strings"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	keysBatchSize    = 100
+	migrateTimeoutMs = 5000
```
jdheyburn: Would we want to make these configurable in the future? (Cluster layman here :) )

internal/valkey/cluster_migration.go
jdheyburn, line 1: Are there any companion unit tests that could be included for this? Or are they covered elsewhere?

internal/valkey/cluster_migration.go
ysqyang: @jdheyburn Good idea. Let's do that!

internal/controller/valkeycluster_controller.go
ysqyang: When scaling out (e.g. from 3 to 4 shards), each existing shard might need to give up ~1365 slots. Rather than migrating all of them in one shot — which would be a long-running, potentially disruptive operation — the controller moves at most `rebalanceSlotBatchSize` slots per reconcile loop, then requeues. This gives the operator a chance to re-read cluster state between batches, react to failures, and avoid tying up the cluster with one massive migration. It's essentially an incremental/chunked approach to slot rebalancing.

internal/valkey/cluster_migration.go
ysqyang: I'm removing these now that we've decided to only support valkey >= 9.0.0 in this PR:)

internal/valkey/cluster_migration.go
ysqyang: added

internal/controller/valkeycluster_controller.go
jdheyburn: Sounds good, in the future perhaps we should make this configurable.

internal/valkey/cluster_rebalance.go
```diff
@@ -87,9 +87,6 @@ func PlanRebalanceMove(shards []*ShardState, expectedShards int, maxSlots int) (
 
 	numSlotsToMove := min(src.numSlots-src.targetNumSlots, dst.targetNumSlots-dst.numSlots, maxSlots)
 	slots := takeSlotsFromRanges(src.ranges, numSlotsToMove)
-	if len(slots) == 0 {
```
bjosv, line 90: This was added since takeSlotsFromRanges could return nil, and to avoid having SlotMove(Slots: nil), is there a reason for removing it?

internal/controller/valkeycluster_controller.go
```diff
@@ -779,6 +811,68 @@ func (r *ValkeyClusterReconciler) countReadyShards(state *valkey.ClusterState, c
...
+			log.V(1).Info("slots no longer served by source; will retry with fresh state", "src", move.Src.Address, "dst", move.Dst.Address)
+			return true, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
+
+// effectiveShards returns state.Shards plus any pending primaries that are
+// scale-out leaders (slot-less Valkey masters labeled as RolePrimary).
```
bjosv:
```suggestion
// scale-out leaders (slot-less Valkey primaries labeled as RolePrimary).
```

internal/controller/valkeycluster_controller.go
```diff
@@ -259,6 +269,28 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	}
 
+	// Rebalance slots across primaries if needed (scale-out).
```
bjosv: Since we have added a very detailed comment of what the `Reconcile` function does (its function doc), I guess we also need to add something about this?

internal/valkey/cluster_migration.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// NodeKnownToSource reports whether dst's node ID appears in
+// src's CLUSTER NODES output (i.e. gossip has propagated).
+func NodeKnownToSource(ctx context.Context, src *NodeState, dst *NodeState) (bool, error) {
+	if dst.Id == "" {
+		return false, nil
+	}
+	cmd := src.Client.B().ClusterNodes().Build()
+	nodes, err := src.Client.Do(ctx, cmd).ToString()
```
bjosv: `src.ClusterNodes` already contain the result from `CLUSTER NODES`, can it be used to avoid another (costly) query?

internal/valkey/cluster_rebalance.go
ysqyang: This is not needed. First, `numSlotsToMove` from line 88 is guaranteed to be at least 1 because `src.numSlots-src.targetNumSlots`  is at least 2 due to line 77, `dst.targetNumSlots-dst.numSlots` is at least 2 based due to 80, and `maxSlots` is at least 1 due to line 41. And `src.ranges` is guaranteed to be non-empty since `src.numSlots - src.targetNumSlots` > 1. Therefore `takeSlotsFromRanges` will always return a non-empty slice.

internal/valkey/cluster_migration.go
ysqyang: Good catch. We don't actually need a function for this at all; a simple `strings.Contains(move.Src.ClusterNodes, move.Dst.Id)` would do.

test/e2e/valkeycluster_test.go
```diff
@@ -310,6 +313,98 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+
+			By("creating a smaller ValkeyCluster for scale-out")
+			scaleOutManifest := fmt.Sprintf(`apiVersion: valkey.io/v1alpha1
+kind: ValkeyCluster
+metadata:
+  name: %s
+spec:
+  shards: %d
+  replicas: 1
+  resources:
```
bjosv: Do we need resource restrictions here?

test/e2e/valkeycluster_test.go
```diff
@@ -310,6 +313,98 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+
+			By(fmt.Sprintf("waiting for the cluster to report %d ready shards", scaleOutShards))
+			verifyScaledOut := func(g Gomega) {
+				cr, err := utils.GetValkeyClusterStatus(valkeyClusterName)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(cr.Status.Shards).To(Equal(int32(scaleOutShards)))
+				g.Expect(cr.Status.ReadyShards).To(Equal(int32(scaleOutShards)))
+				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+			}
+			Eventually(verifyScaledOut, 10*time.Minute).Should(Succeed())
```
bjosv: Oh, would this step also need 10min max?
Wouldn't previous step (verifying all primaries receive slots after scale out) make sure this step is done?

test/e2e/valkeycluster_test.go
```diff
@@ -310,6 +313,98 @@ var _ = Describe("ValkeyCluster", Ordered, func() {
...
+    limits:
+      memory: "512Mi"
+      cpu: "500m"
+`, valkeyClusterName, baseShards)
+			manifestFile := filepath.Join(os.TempDir(), "valkeycluster-scaleout.yaml")
+			err := os.WriteFile(manifestFile, []byte(scaleOutManifest), 0644)
+			Expect(err).NotTo(HaveOccurred(), "Failed to write scale-out manifest file")
+			defer os.Remove(manifestFile)
+
+			cmd := exec.Command("kubectl", "delete", "valkeycluster", valkeyClusterName, "--ignore-not-found=true")
```
bjosv, line 342: Other tests uses `kubectl create`, which gives an error if already existing, and they are not deleting any old previous runs. I think its nice if we use the same setup sequence, we should either change this or the other tests.

test/e2e/valkeycluster_test.go
ysqyang: No it doesn't. Let me change it back to 2 minutes.

test/e2e/valkeycluster_test.go
ysqyang: No we don't:). Let me remove them.

PR #102, opened 2026-03-02
Setup container image publishing

SuperQ:
Add a GitHub Actions workflow to build and publish container images.
* Build but not publish on pull request.
* Publish `:main` on merge to main.
* Publish `:latest` on release tags.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,77 @@
+---
+name: Publish
+
+on:
+  pull_request:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - 'v*'
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    name: Publish container images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+
+      - name: Reclaim disk space
+        run: |
+          docker image prune --force --all
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+
+      - name: Install Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417  # v6.3.0
+        with:
+          go-version-file: go.mod
+          check-latest: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Login to ghcr.io
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
+        with:
+          images: ghcr.io/${{ github.repository }}
+          tags: |
+            # Tag with branch name
+            type=ref,event=branch
+            # Tag with PR number
+            type=ref,event=pr
+            # Tag with semver from git tag
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            # Tag with short SHA
+            type=sha,prefix=
+
+      - name: Get version info
+        id: version
+        run: echo "$(make print-version-ldflags)" | tee -a "$GITHUB_OUTPUT"
+
+      - name: Build and publish container
+        id: publish
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,56 @@
...
+      # - name: Login to ghcr.io
+      #   uses: docker/login-action@v3
+      #   with:
+      #     registry: ghcr.io
+      #     username: ${{ github.repository_owner }}
+      # - name: Extract metadata (tags, labels) for Docker
+      #   id: meta
+      #   uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+      #   with:
+      #     images: valkey/valkey-operator
```
jdheyburn: @stockholmux Is there somewhere we can publish the container images to?

SuperQ: I saw that Valkey has a Docker Hub account. So it should be possible to either add secrets access to this repo via GH org controls.

Or we can create a separate account that has push access to `valkey/valkey-operator` on Docker Hub.

I did a quick browse of the upstream valkey, I didn't see a GitHub action for publishing official builds there.

hieu2102: The repo handling build & uploads for core Valkey is [valkey-release-automation](https://github.com/valkey-io/valkey-release-automation).
You should take a look at the [valkey-bundle repo](https://github.com/valkey-io/valkey-bundle). It seems their steps are applicable to ours

SuperQ: I've updated the PR to publish to `ghcr.io`. This is built-in to GitHub so we can start testing with it. We can add a `docker.io` login later.

jdheyburn (dismissed):
LGTM, @SuperQ is there anything else we need to do to get images published? Would this get picked up automatically after we merge?

SuperQ:
Everything should be automatic after merge.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,72 @@
...
+            type=sha,prefix=
+
+      - name: Build and publish container
+        id: publish
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
+        with:
+          context: .
+          file: ./Dockerfile
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
```
bjosv, line 75: In the Makefile we set VERSION_LDFLAGS to set the built-in version information, but I guess this action don't use the Makefile.
Should we set VERSION_LDFLAGS as a build-args in this job?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,72 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
+        with:
+          images: ghcr.io/${{ github.repository }}
+          tags: |
```
bjosv, line 52: Do we get the `:latest` tag here somehow? I guess `:main` can come from `event=branch`

SuperQ: `:latest` will get set on tags.

.github/workflows/publish.yml
SuperQ: Oh yea, we need to plumb that in as well.

.github/workflows/publish.yml
bjosv: Ah ok, thanks. I now see that it was enabled by default in metadata-action

sandeepkunusoth (dismissed):

sandeepkunusoth:
i am not able to pull this image which is tagged ghcr.io/valkey-io/valkey-operator:pr-102
image

.github/workflows/publish.yml
```diff
@@ -0,0 +1,72 @@
+---
+name: Publish
+
+on:
+  pull_request:
+  push:
```
sandeepkunusoth, line 6: should we run publish on every pull_request? or only on merging to main

SuperQ: I don't think we should, this would pollute the public container registry a lot.

Maybe you missed this feature:
```
push: ${{ github.event_name != 'pull_request' }}
```

We run the build step, but don't actually push it on `pull_request`.

.github/workflows/publish.yml
SuperQ: Ok, I added a step to use `make` to extract the ldflags.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,77 @@
...
+      - 'v*'
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    name: Publish container images
+    runs-on: ubuntu-latest
```
bjosv, line 19: Do we want our own forks to run this job as well (most likely when its main is updated), or would it make sense to add a
`if: github.repository == 'valkey-io/valkey-operator'

bjosv (approved):

.github/workflows/publish.yml
SuperQ: Hmm, that's an option. I typically just disable Actions on my forks to avoid the noise.

PR #104, opened 2026-03-03
Add basic reconcile metrics

SuperQ:
Add basic metrics to monitor the reconcile loop.
* Histogram of reconcile timing.
* Counter for errors.
* Wrap the main reconcile function to make capturing the metrics easier.
* Future proof this to be able to handle add new control loops later.

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -37,6 +39,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
 	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
 	"valkey.io/valkey-operator/internal/valkey"
 )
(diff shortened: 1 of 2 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -52,6 +56,30 @@ const (
...
+		},
+		[]string{"controller"},
+	)
+	reconcileErrors = promauto.With(metrics.Registry).NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "valkey_operator_reconcile_errors_total",
+			Help: "Number of errors when performing reconcile",
+		},
+		[]string{"controller"},
+	)
```
sandeepkunusoth, line 77: are there any other open source operators doing this? can u share any reference.  i mean cant we use metrics endpoint exposed to get these metrics?

jdheyburn:
Does this also expose the prometheus metrics on the operator pod, or would that be covered by the Helm chart?

internal/controller/valkeycluster_controller.go
SuperQ: I'm not sure I understand the question. These are new metrics that are not part of the default set.

A good example of this is the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator/blob/b42636bc1037a9576527149d2bd3e40c93c8f106/pkg/operator/resource_reconciler.go#L664-L692). This operator has a bit of a different structure, but the metrics style is basically the same.

bjosv (approved):

internal/controller/valkeycluster_controller.go
```diff
@@ -98,6 +126,17 @@ var scripts embed.FS
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.1/pkg/reconcile
 func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	reconcileStart := time.Now()
```
bjosv, line 129: still a bit weird to me that we need this when its already done in the controller-runtime code..

SuperQ:
Oh, I didn't realize there was already a controller runtime metric for this. That's fine, we don't need this.

I'll dig into the controller-runtime some more and see if there are additional gaps we might want that aren't included there.

Issue #103, opened 2026-03-03
Cluster cannot recover from deleted deployment


utdrmac:
I deployed a new cluster using the operator code at 9264a50c. After everything stabilized, I had a cluster of 3 shards, 1 replica for each shard. This is the default deployment out of config/samples/
I then deleted one of the Deployments, expecting the operator to recreate the deployment and recover the cluster. I did not know which deployment I was deleting(ie: primary, or replica). I picked one at random. The operator was unable to recover this. A new deployment was created, and pod launched, but the operator cannot figure out how to rejoin this node into the cluster. I've attached the operator log from the moment of deployment deletion.
This situation should be added to the E2E test suite. The loss of a deployment could come at any time, notably a node failure.
Tagging a few people that might be involved: @ysqyang @sandeepkunusoth @bjosv

sandeepkunusoth:
https://github.com/valkey-io/valkey-operator/actions/runs/22506608716/job/65206553825. i observed sometimes e2e tests on main branch  also fails sometimes when ran repeatedly maybe related

jdheyburn:
We've had a number of stability improvements added in 0.2 - if this can't be replicated there then we can consider closing this ticket in the future.

jdheyburn:
Closing - if the issue resurfaces please raise another issue with repro steps for us to investigate.

PR #105, opened 2026-03-04
fix race condition between forget and failover

ysqyang:
## Summary

Fix a race condition between `forgetStaleNodes` and Valkey's auto-failover that can permanently prevent a replica from being promoted after its primary dies. See https://github.com/valkey-io/valkey-operator/issues/103 for more context.

### The bug

When a primary's deployment is deleted, the controller's `forgetStaleNodes` issues `CLUSTER FORGET` for the dead node from every surviving node. If this runs before Valkey's auto-failover election completes, it removes the dead primary from the other masters' node tables. Those masters can then no longer validate the replica's `FAILOVER_AUTH_REQUEST` (they don't recognize the dead node), so they never vote. The replica is permanently stuck as a slave, `findShardPrimary` never finds a primary for the shard, and the cluster enters an infinite loop of:

```
ERROR  command failed: CLUSTER FORGET  {"error": "Can't forget my master!"}
DEBUG  skipping replica; primary not ready yet
DEBUG  missing replicas, requeue..
```

This is a timing-dependent race. The window is roughly 0.5–1 second between the `fail` flag being set and the failover election completing. It was reported by a user who hit it when deleting a primary deployment.

### The fix

Before issuing `CLUSTER FORGET`, check whether any live node in the cluster still considers the failing node as its master (`HasReplicaOf`). If so, skip the FORGET — the replica needs the dead node in the other masters' node tables to complete the failover election. Once the failover completes and the replica is promoted, it no longer reports itself as a slave of the dead node, so the next reconcile will proceed with FORGET normally.

### Changes

- **`internal/valkey/clusterstate.go`**: Add `HasReplicaOf(nodeId)` method on `ClusterState` that checks if any node's `CLUSTER NODES` self-report shows it as a replica of the given node ID. Add `MasterIdFromSelf()` helper on `NodeState` that extracts `fields[3]` (master ID) from the `myself` line.
- **`internal/controller/valkeycluster_controller.go`**: Guard `forgetStaleNodes` with the `HasReplicaOf` check. When skipped, log `"skipping forget; failover pending for node"` at V(1).

### Why this is safe

- **Dead replica (not a master):** No node claims a replica as its master → `HasReplicaOf` returns false → FORGET proceeds immediately. No behavior change.
- **Both master and replica are dead:** The dead replica isn't in `state.Shards` (connection failed) → `HasReplicaOf` returns false → FORGET proceeds. Correct — no failover is possible anyway.
- **Scale-down stale nodes:** Drained masters have no replicas left → `HasReplicaOf` returns false → FORGET proceeds. No behavior change.
- **Failover permanently blocked for other reasons** (e.g., replica too far behind): `HasReplicaOf` returns true, FORGET is deferred. This is no worse than today where FORGET runs but the failover is also permanently blocked. With this fix, at least the failover has a chance if the blocking condition resolves.


internal/valkey/clusterstate.go
```diff
@@ -193,6 +193,38 @@ func (n *NodeState) IsReplicationInSync() bool {
...
+			}
+		}
+	}
+	return false
+}
+
+// MasterIdFromSelf returns the master node ID that this node reports as its
+// own master in CLUSTER NODES (fields[3] of the "myself" line). Returns "-"
+// for masters and the master's node ID for replicas.
+func (n *NodeState) MasterIdFromSelf() string {
```
bjosv:
```suggestion
func (n *NodeState) GetPrimaryId() string {
```

internal/controller/valkeycluster_controller.go
```diff
@@ -682,20 +682,28 @@ func (r *ValkeyClusterReconciler) forgetStaleNodes(ctx context.Context, cluster
...
-						log.Error(err, "command failed: CLUSTER FORGET")
-						r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "NodeForgetFailed", "ForgetNode", "Failed to forget node: %v", err)
-					} else {
-						r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "StaleNodeForgotten", "ForgetNode", "Forgot stale node %v", failing.Address)
-					}
+				if idx != -1 {
+					continue
+				}
+				// A live replica still considers this failing node its
+				// master. Forgetting it from the other masters now would
```
bjosv: Replace `master` with `primary`

internal/valkey/clusterstate.go
```diff
@@ -193,6 +193,38 @@ func (n *NodeState) IsReplicationInSync() bool {
 	return n.Info["master_link_status"] == "up"
 }
 
+// HasReplicaOf returns true if any live node in the cluster state reports
+// itself as a replica of the given node ID. This is used to prevent
+// CLUSTER FORGET from racing with auto-failover: forgetting a failed
+// primary from other masters removes it from their node tables, which
```
bjosv: Replace `masters` with e.g. `replica nodes`

bjosv (commented):
When running e2e tests in CI I hit the error in ~1 of 4 runs,
but with this fix I'm able to run 20 runs without any errors.
Seems to fix our problem!

jdheyburn (approved):
I rebased from main to pull in the ValkeyNode changes. We've discussed on a previous tech call that we can get this merged in once it had been rebased. Thank you @ysqyang!

PR #106, opened 2026-03-06
Add build info metric

SuperQ:
Add a `valkey_operator_build_info` metric. This allows tracking of rollouts via metrics. This uses the standard Prometheus version collector helper package and injects the build context via Go build ldflags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+VERSION              ?= latest
+IMG                  ?= controller
+CONTAINER_IMAGE_NAME ?= $(IMG):$(VERSION)
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -18,6 +20,16 @@ CONTAINER_TOOL ?= docker
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
+
+BRANCH    ?= $(shell git rev-parse --abbrev-ref HEAD)
+BUILDDATE ?= $(shell date -Iseconds)
+REVISION  ?= $(shell git rev-parse HEAD)
+
+VERSION_LDFLAGS := \
+  -X github.com/prometheus/common/version.Version=$(VERSION) \
+  -X github.com/prometheus/common/version.Revision=$(REVISION) \
+  -X github.com/prometheus/common/version.Branch=$(BRANCH) \
+  -X github.com/prometheus/common/version.BuildDate=$(BUILDDATE)
 
 .PHONY: all
 all: build
@@ -106,7 +118,7 @@ lint-config: golangci-lint ## Verify golangci-lint linter configuration
 
 .PHONY: build
 build: manifests generate fmt vet lint ## Build manager binary.
-	go build -o bin/manager cmd/main.go
+	go build -ldflags "$(VERSION_LDFLAGS)" -o bin/manager cmd/main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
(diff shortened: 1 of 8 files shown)


cmd/main.go
```diff
@@ -89,6 +91,8 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	metrics.Registry.MustRegister(versioncollector.NewCollector("valkey_operator"))
```
sandeepkunusoth, line 94: Can we update e2e test of valkey operator manager metrics endpoint on exactly what new metric is getting added here.

cmd/main.go
```diff
@@ -89,6 +91,8 @@ func main() {
...
+
```
bjosv, line 95: It would be nice if we also show the version in the log, but maybe that is a separate change?

It seems to exist an prometheus [Info()](https://github.com/prometheus/common/blob/v0.67.5/version/info.go#L73) like:
`setupLog.Info("Starting valkey-operator", "version", versioncollector.Info())`

or should we instead add extra JSON-fields using
`setupLog = setupLog.WithValues("version", versioncollector.Version, "commit", versioncollector.Commit)`.
In that case we store all version info in prometheus versioncollector, instead of in the (often used) main package, I guess thats ok?

Makefile
```diff
@@ -19,6 +19,17 @@ CONTAINER_TOOL ?= docker
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+VERSION   ?= 0.0.1
```
bjosv: We have `IMG ?= controller:latest` in the beginning of this makefile,
it might be nice to gather commonly used inputs.

The plan is that we make sure VERSION and IMG are aligned?
Like `make docker-build VERSION=0.2.4 IMG=valkey/valkey-operator:0.2.4`

bjosv (commented):
Nice! Some comments/questions

cmd/main.go
SuperQ: Yea, there's a version log string helper. Easy to add.

Good point about the structured fields. That's something we could add to the version helper package.

SuperQ: I created https://github.com/prometheus/common/pull/886, what do you think?

Makefile
SuperQ: Sure, actually, I can recompose this a bit.

cmd/main.go
SuperQ: Filed https://github.com/valkey-io/valkey-operator/issues/114

sandeepkunusoth: i still didn't get what this PR adds. there is no update in e2e tests

cmd/main.go
bjosv: Looks good, I guess we need to wait for it to be released before we can use it and emit a log entry.

test/e2e/valkeycluster_test.go
```diff
@@ -29,8 +29,8 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
```
bjosv, line 32: was this change from some auto-formatter?

bjosv (approved):

cmd/main.go
jdheyburn: There is some more documentation here: https://oneuptime.com/blog/post/2026-01-30-prometheus-info-metrics/view

We spoke on the weekly call that we will be adding e2e tests for metrics at a later stage once we start to onboard them.

test/e2e/valkeycluster_test.go
SuperQ: Yes, go fmt

Makefile
```diff
@@ -1,5 +1,7 @@
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+VERSION              ?= latest
+IMG                  ?= controller
+CONTAINER_IMAGE_NAME ?= $(IMG):$(VERSION)
```
sandeepkunusoth, line 4: Cant we just parse VERSION from IMG? here instead of changing definition of IMG and expecting users to pass VERSION in make deploy and make docker build. i think that would simplify this PR more.

jdheyburn: @SuperQ ICYMI

jdheyburn: Raised below to continue this thread:

- https://github.com/valkey-io/valkey-operator/pull/130

Issue #62, opened 2026-01-22
(feat) Make exporter sidecar more generic


utdrmac:
The operator currently has a metrics exporter sidecar, however, the implementation is tailored to a specific container image, and does not allow for any custom parameters, environment variables, etc. This feature request aims to "generic-ify" the exporter sidecar so that any container image can be used by the end-user.

Spec Changes Proposal
```
spec:
  metrics:
    enabled: true
    image: "oliver006/redis_exporter:v1.80.0"
    imagePullSecret: gitlab-cr-token
    command: ["some-exporter", "--flag1", "--flag2=foobar"]
    port: 9121
    resources:
      limits:
        cpu: 100m
        memory: 128Mi
    extraSecretsMounts:
      "some-secret": "/path/to/mount"
    extraEnvMounts:
      "some-configmap": "/path/config"
```

- If `spec.image` is not specified, then `command`, and `port` are ignored, and will use existing values.
  - Any other provided options will be merged into the existing container.
- If `spec.image` is not the same as default image, then `command` and `port` are required.
- If only spec.metrics.enabled=true is specified, the behavior is the same for backwards compatibility.

- [x] I'm willing to implement this feature

jdheyburn:
Thanks for the detail @utdrmac, that helps flesh out a design.

What I'm thinking is something similar to promtheus-operator which would solve both overriding any container (server, exporter), and adding additional containers.

```go
type ValkeyClusterSpec struct {
    // ... existing fields unchanged ...

    Exporter ExporterSpec `json:"exporter,omitempty"`

    // Containers allows overriding operator-managed containers via strategic
    // merge patch, matched by name. Known names ("server", "exporter") are
    // merged against operator defaults. Unknown names are appended as
    // additional sidecar containers, passed through verbatim.
    // +listType=map
    // +listMapKey=name
    // +optional
    Containers []corev1.Container `json:"containers,omitempty"`
}
```

For overriding any property of exporter container:

```yaml
spec:
  exporter:
    enabled: true
  containers:
    - name: exporter
      image: "my-custom-exporter:v2.0"
```

For additional containers and disabling exporter:

```yaml
spec:
  exporter:
    enabled: false
  containers:
    - name: my-monitoring-agent
      image: monitoring:latest
      ports:
        - containerPort: 8080
      env:
        - name: TARGET
          value: "localhost:6379"
```

There is a weekly tech call today, I'll bring this up then for discussion. Let me know if you have any additional comments.

PR #108, opened 2026-03-09
Implement strategic merge patch (#62)

hieu2102:
Implements #62 
- add field `Containers` to `ValkeyClusterSpec`
- add function `deployment.mergePatchContainers` 
- modify the function signatures of `deployment.generateContainersDef` and `deployment.createClusterDeployment` to return `error` in addition to the output object
- add tests


internal/controller/deployment.go
```diff
@@ -206,5 +216,54 @@ func createClusterDeployment(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex
...
+		if err := json.Unmarshal(jsonResult, &patchResult); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal merged container %s: %w", c.Name, err)
+		}
+
+		// Add the patch result and remove the corresponding key from the to do list.
+		output = append(output, patchResult)
+		delete(containersByName, c.Name)
+
+	}
+	for _, c := range patches {
```
bjosv, line 262: This is handling added containers?, i.e. those that were not patched and deleted from the map
Couldn't that be simplified to
```
for _, container := range containersByName {
    output = append(output, container)
}
```
but maybe we need to include added container in the same order as given?

test/e2e/valkeycluster_test.go
```diff
@@ -665,4 +665,109 @@ spec:
...
+			Expect(err).NotTo(HaveOccurred(), "Failed to write manifest file")
+			defer func() {
+				Expect(os.Remove(manifestFile)).To(Succeed())
+			}()
+
+			By("applying the ValkeyCluster CR")
+			cmd := exec.Command("kubectl", "apply", "-f", manifestFile)
+			output, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to apply ValkeyCluster: %s", output))
+
```
bjosv, line 706: The cluster needs to be deleted when test has finished (~"kubectl delete valkeycluster"+valkeyClusterName)

internal/controller/deployment.go
hieu2102: yeah, you are correct. I think it is better to add them in order, less surprise (if any) that way

PR #109, opened 2026-03-09
Add additional checks in CI

bjosv:
Verify that code formatting is correct, and that `go.mod` and generated code are up to date and committed.
Includes a correction of code formatting in an e2e test.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   lint:
-    name: Run on Ubuntu
+    name: Run linters
     runs-on: ubuntu-latest
     steps:
       - name: Clone the code


.github/workflows/lint.yml
```diff
@@ -24,3 +24,52 @@ jobs:
...
+          git diff --exit-code go.mod go.sum
+
+  generated:
+    name: Verify generated code
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - name: Setup Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417  # v6.3.0
```
sandeepkunusoth, line 68: can we add these steps as part of pre commit git hook instead of running on CICD. just thinking. if we added prehook it will not allow to commit at all.

jdheyburn: I'm in favour of adding the checks in both places. Users can skip pre-commit checks with `--no-verify`, so we would still need CI checks to validate on the server side.

bjosv: Isn't the pre-commit hook a local-only setup, i.e. a contributor need to perform additional steps to set it up?
..or is it possible to enforce it these days?

bjosv:
Triggered #112

PR #113, opened 2026-03-12
feat: ValkeyNode integration with ValkeyCluster

jdheyburn:
- ValkeyCluster creates ValkeyNode instances over managing deployments directly.
- Introduces StatefulSets as a workload Type
  - This is the new default, with an option to deploy Deployment too
- Splits out multiple config/samples of ValkeyCluster

**What this PR does not include (and will be fixed in subsequent PRs):**
- Updates to ValkeyNodes (i.e. image changes)
  - This mimics the current behaviour

**What this PR unblocks:**
- Live config reconciliation (avoiding pod rolls if necessary)
  - Will also address the temporary `ScriptsConfigMapName` and `UsersACLSecretName` fields on ValkeyNode
- Persistent volume management

**What can be improved after this PR is merged:**
- ValkeyNode ROLE reconciliation
  - Can report an incorrect role for up to 1m
- More e2e tests...?

**Below after applying config/samples/v1alpha1_valkeycluster.yaml**

```
$ kg valkeynode
NAME                 READY   ROLE      POD                           AGE
cluster-sample-0-0   true    primary   valkey-cluster-sample-0-0-0   3m49s
cluster-sample-0-1   true    replica   valkey-cluster-sample-0-1-0   3m49s
cluster-sample-1-0   true    primary   valkey-cluster-sample-1-0-0   3m49s
cluster-sample-1-1   true    replica   valkey-cluster-sample-1-1-0   3m49s
cluster-sample-2-0   true    primary   valkey-cluster-sample-2-0-0   3m49s
cluster-sample-2-1   true    replica   valkey-cluster-sample-2-1-0   19s
```

Deploying this yaml:

```yaml
kind: ValkeyCluster
metadata:
  name: cluster-sample-deployment
spec:
  shards: 3
  replicas: 1
  workloadType: Deployment
  resources:
    requests:
      memory: "256Mi"
      cpu: "100m"
    limits:
      memory: "512Mi"
      cpu: "500m"
  containers:
    - name: metrics-exporter
      image: oliver006/redis_exporter:v1.82.0
```

Deploys this:

```
$ kg valkeynode
NAME                            READY   ROLE      POD                                                     AGE
cluster-sample-deployment-0-0   true    primary   valkey-cluster-sample-deployment-0-0-7c94f6d589-5kp5t   20s
cluster-sample-deployment-0-1   true    replica   valkey-cluster-sample-deployment-0-1-784d8c999f-jqzjb   20s
cluster-sample-deployment-1-0   true    primary   valkey-cluster-sample-deployment-1-0-6c66c58f8c-xm74t   20s
cluster-sample-deployment-1-1   true    replica   valkey-cluster-sample-deployment-1-1-74c4dd5d57-fjdnc   20s
cluster-sample-deployment-2-0   true    primary   valkey-cluster-sample-deployment-2-0-75dc8dbcd9-7pbrz   20s
cluster-sample-deployment-2-1   true    replica   valkey-cluster-sample-deployment-2-1-6f8d757c59-9tv64   20s
```

And the container merge traverses to the pod (the default exporter version is 1.80.0):

```
$ kgpo valkey-cluster-sample-deployment-2-1-6f8d757c59-9tv64 -o yaml G redis_exporter
    image: oliver006/redis_exporter:v1.82.0
    image: docker.io/oliver006/redis_exporter:v1.82.0
    imageID: docker.io/oliver006/redis_exporter@sha256:6a97d4dd743b533e1f950c677b87d880e44df363c61af3f406fc9e53ed65ee03
```

### DIFF
--- a/config/crd/bases/valkey.io_valkeynodes.yaml
+++ b/config/crd/bases/valkey.io_valkeynodes.yaml
@@ -975,6 +975,1532 @@ spec:
                         x-kubernetes-list-type: atomic
                     type: object
                 type: object
+              containers:
+                description: |-
+                  Containers allows patching the default containers via a strategic merge
+                  patch. Existing containers (e.g. "server", "metrics-exporter") are merged
+                  by name; any container name not present in the defaults is appended.
+                items:
+                  description: A single application container that you want to run
+                    within a pod.
+                  properties:
+                    args:
+                      description: |-
+                        Arguments to the entrypoint.
+                        The container image's CMD is used if this is not provided.
+                        Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
+                        cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
+                        to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
+                        produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
+                        of whether the variable exists or not. Cannot be updated.
+                        More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
+                      items:
+                        type: string
+                      type: array
+                      x-kubernetes-list-type: atomic
+                    command:
+                      description: |-
+                        Entrypoint array. Not executed within a shell.
+                        The container image's ENTRYPOINT is used if this is not provided.
+                        Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
+                        cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
+                        to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
+                        produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
+                        of whether the variable exists or not. Cannot be updated.
+                        More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
+                      items:
+                        type: string
+                      type: array
+                      x-kubernetes-list-type: atomic
+                    env:
+                      description: |-
+                        List of environment variables to set in the container.
+                        Cannot be updated.
+                      items:
+                        description: EnvVar represents an environment variable present
+                          in a Container.
+                        properties:
+                          name:
+                            description: |-
+                              Name of the environment variable.
+                              May consist of any printable ASCII characters except '='.
+                            type: string
+                          value:
+                            description: |-
+                              Variable references $(VAR_NAME) are expanded
+                              using the previously defined environment variables in the container and
+                              any service environment variables. If a variable cannot be resolved,
+                              the reference in the input string will be unchanged. Double $$ are reduced
+                              to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e.
+                              "$$(VAR_NAME)" will produce the string literal "$(VAR_NAME)".
+                              Escaped references will never be expanded, regardless of whether the variable
+                              exists or not.
+                              Defaults to "".
+                            type: string
+                          valueFrom:
+                            description: Source for the environment variable's value.
+                              Cannot be used if value is not empty.
+                            properties:
+                              configMapKeyRef:
+                                description: Selects a key of a ConfigMap.
+                                properties:
+                                  key:
+                                    description: The key to select.
+                                    type: string
+                                  name:
+                                    default: ""
+                                    description: |-
+                                      Name of the referent.
+                                      This field is effectively required, but due to backwards compatibility is
+                                      allowed to be empty. Instances of this type with an empty value here are
+                                      almost certainly wrong.
+                                      More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                    type: string
+                                  optional:
+                                    description: Specify whether the ConfigMap or
+                                      its key must be defined
+                                    type: boolean
+                                required:
+                                - key
+                                type: object
+                                x-kubernetes-map-type: atomic
+                              fieldRef:
+                                description: |-
+                                  Selects a field of the pod: supports metadata.name, metadata.namespace, `metadata.labels['<KEY>']`, `metadata.annotations['<KEY>']`,
+                                  spec.nodeName, spec.serviceAccountName, status.hostIP, status.podIP, status.podIPs.
+                                properties:
+                                  apiVersion:
+                                    description: Version of the schema the FieldPath
+                                      is written in terms of, defaults to "v1".
+                                    type: string
+                                  fieldPath:
+                                    description: Path of the field to select in the
+                                      specified API version.
+                                    type: string
+                                required:
+                                - fieldPath
+                                type: object
+                                x-kubernetes-map-type: atomic
+                              fileKeyRef:
+                                description: |-
+                                  FileKeyRef selects a key of the env file.
+                                  Requires the EnvFiles feature gate to be enabled.
+                                properties:
+                                  key:
+                                    description: |-
+                                      The key within the env file. An invalid key will prevent the pod from starting.
+                                      The keys defined within a source may consist of any printable ASCII characters except '='.
+                                      During Alpha stage of the EnvFiles feature gate, the key size is limited to 128 characters.
+                                    type: string
+                                  optional:
+                                    default: false
+                                    description: |-
+                                      Specify whether the file or its key must be defined. If the file or key
+                                      does not exist, then the env var is not published.
+                                      If optional is set to true and the specified key does not exist,
+                                      the environment variable will not be set in the Pod's containers.
+
+                                      If optional is set to false and the specified key does not exist,
+                                      an error will be returned during Pod creation.
+                                    type: boolean
+                                  path:
+                                    description: |-
+                                      The path within the volume from which to select the file.
+                                      Must be relative and may not contain the '..' path or start with '..'.
+                                    type: string
+                                  volumeName:
+                                    description: The name of the volume mount containing
+                                      the env file.
+                                    type: string
+                                required:
+                                - key
+                                - path
+                                - volumeName
+                                type: object
+                                x-kubernetes-map-type: atomic
+                              resourceFieldRef:
+                                description: |-
+                                  Selects a resource of the container: only resources limits and requests
+                                  (limits.cpu, limits.memory, limits.ephemeral-storage, requests.cpu, requests.memory and requests.ephemeral-storage) are currently supported.
+                                properties:
+                                  containerName:
+                                    description: 'Container name: required for volumes,
+                                      optional for env vars'
+                                    type: string
+                                  divisor:
+                                    anyOf:
+                                    - type: integer
+                                    - type: string
+                                    description: Specifies the output format of the
+                                      exposed resources, defaults to "1"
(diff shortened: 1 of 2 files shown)


test/e2e/valkeycluster_test.go
```diff
@@ -586,20 +617,23 @@ spec:
...
-			getDeployment := func(g Gomega) {
-				var err error
-				deploymentToDelete, err = utils.GetReplicaDeployment(fmt.Sprintf("app.kubernetes.io/instance=%s", degradedClusterName))
-				g.Expect(err).NotTo(HaveOccurred(), "Failed to find a replica deployment")
-				g.Expect(deploymentToDelete).NotTo(BeEmpty())
+			By("getting a replica statefulset to delete")
+			var statefulsetToDelete string
+			getStatefulset := func(g Gomega) {
+				cmd := exec.Command("kubectl", "get", "statefulsets",
+					"-l", fmt.Sprintf("valkey.io/cluster=%s,valkey.io/node-index=1", degradedClusterName),
```
bjosv, line 617: When we previously used utils.GetReplicaDeployment we searched for an instance by probing the valkey instance and its role. Now the test needs to trust that the controller has done what it should correctly.
Maybe we should create a new util-function that can check that the found `statefulsetToDelete` has a pod with a valkey instance that  really is a replica, or is that overdoing it?

test/e2e/valkeycluster_test.go
```diff
@@ -497,7 +528,7 @@ spec:
 				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
 				g.Expect(cr.Status.ReadyShards).To(Equal(int32(scaleInShards)))
 			}
-			Eventually(verifyScaledIn).Should(Succeed())
+			Eventually(verifyScaledIn, 10*time.Minute, 2*time.Second).Should(Succeed())
```
bjosv, line 524: +-2 seconds shouldn't matter? Same above
```suggestion
			Eventually(verifyScaledIn, 10*time.Minute).Should(Succeed())
```

test/e2e/valkeycluster_test.go
```diff
@@ -414,6 +438,13 @@ metadata:
 spec:
   shards: %d
   replicas: 1
+  resources:
```
bjosv: needed for this test?

test/e2e/valkeycluster_test.go
```diff
@@ -363,7 +387,7 @@ spec:
 			By("verifying all primaries receive slots after scale out")
 			verifySlotRebalance := func(g Gomega) {
 				cmd := exec.Command("kubectl", "get", "pods",
-					"-l", fmt.Sprintf("app.kubernetes.io/instance=%s", valkeyClusterName),
```
bjosv, line 366: This should work even if we use `app.kubernetes.io/instance` right?
Or has it different meaning now in a ValkeyNode?

internal/controller/valkeycluster_controller.go
```diff
@@ -408,73 +406,96 @@ aclfile /config/users/users.acl`,
...
 	return nil
 }
 
-// ensureDeployment creates a single Deployment if it doesn't already exist.
-func (r *ValkeyClusterReconciler) ensureDeployment(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, shard int, nodeIndex int, expected int, created *int) error {
-	deployment := createClusterDeployment(cluster, shard, nodeIndex)
-	if err := controllerutil.SetControllerReference(cluster, deployment, r.Scheme); err != nil {
+// ensureValkeyNode creates a single ValkeyNode CR if it doesn't already exist.
+func (r *ValkeyClusterReconciler) ensureValkeyNode(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex int, nodeIndex int, expected int, created *int) error {
+	vkn := buildClusterValkeyNode(cluster, shardIndex, nodeIndex)
```
bjosv: not using `node :=` as in some other places in this PR? 
(I now see there are some use of vkn later on..)

internal/controller/valkeycluster_controller.go
```diff
@@ -408,73 +406,96 @@ aclfile /config/users/users.acl`,
...
 
-func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, pods *corev1.PodList) *valkey.ClusterState {
-	// Create a list of addresses to possible Valkey nodes
+// buildClusterValkeyNode constructs the ValkeyNode CR for a given (shard, node) position.
+func buildClusterValkeyNode(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex int, nodeIndex int) *valkeyiov1alpha1.ValkeyNode {
+	return &valkeyiov1alpha1.ValkeyNode{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      valkeyNodeName(cluster.Name, shardIndex, nodeIndex),
+			Namespace: cluster.Namespace,
+			Labels: map[string]string{
```
bjosv: We don't want to add the K8s recommended labels here?
The CR is created by the operator, like a deployment before which had the labels.
Not sure the benefit of doing it though..

internal/controller/valkeycluster_controller.go
bjosv: Heard that they don't prepend using vk in the valkey repo since vk = vkontakty.
Maybe we can just use vn if we cant use node?

test/e2e/valkeycluster_test.go
jdheyburn: This pattern was copied from an existing test [here](https://github.com/valkey-io/valkey-operator/blob/78139b06eef1829cbb5bbea44ed6d7b2b7a4383a/test/e2e/valkeycluster_test.go#L440). I believe this is because that is the requeue after time of the ValkeyCluster during reconciliation.

When unspecified it defaults to 1s, perhaps to keep things simple we can use that default. What do you think?

test/e2e/valkeycluster_test.go
jdheyburn: The way I saw it, `GetReplicaDeployment` was introduced before we had determinism on the pod role (`cluster-<shard-index>-<node-index>`), which I agree we are now assuming in here.

I would like to have some means of reliably determining this outside of a test utils area (perhaps on ValkeyNode.Status.Role?).

I think you're right, creating a new util-function that can definitely say "this is a replica" would be useful up until we have the above in place, since that's what we had before.

test/e2e/valkeycluster_test.go
jdheyburn: I'll remove this, thanks!

test/e2e/valkeycluster_test.go
jdheyburn: `app.kubernetes.io/instance` has a different meaning in ValkeyNode; it is the name of the ValkeyNode being deployed, so these selectors need to be updated too. 

I did some refactoring of the labels to be consistent across ValkeyCluster and ValkeyNode (and its child resources). 

[baseLabels](https://github.com/valkey-io/valkey-operator/pull/113/changes#diff-2e75daf7fbfd686a57f8e7bd14cb6e04c7ca3ffb3406fa0f702641c72ff12b3dR80)
[valkeyNodeLabels](https://github.com/valkey-io/valkey-operator/pull/113/changes#diff-4335bd43eee11330a0301e5799668b129821638703bbe7c310e6de4b04a67b4bL36)

My opinion is that `instance` is a bit vague, though I am all for including it if its recommended. The [docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) describe it as:

app.kubernetes.io/name	The name of the application	mysql	string
app.kubernetes.io/instance	A unique name identifying the instance of an application	mysql-abcxyz

The guidelines are a bit loose. I associate instance as "what is the instance of this component being deployed". If we're deploying a ValkeyNode, then this would be the name of that component.

internal/controller/valkeycluster_controller.go
jdheyburn: Hmm we're already using `vkc` as a [shortName](https://github.com/valkey-io/valkey-operator/blob/78139b06eef1829cbb5bbea44ed6d7b2b7a4383a/config/crd/bases/valkey.io_valkeyclusters.yaml#L15) for ValkeyCluster. `vkn` is an extension of that in its [CRD](https://github.com/valkey-io/valkey-operator/blob/78139b06eef1829cbb5bbea44ed6d7b2b7a4383a/config/crd/bases/valkey.io_valkeynodes.yaml#L15).

I've updated all references of `vkn` (outside of the shortName) to be `node`. But it would be good to align on the shortName for ValkeyCluster and ValkeyNode.

internal/controller/valkeycluster_controller.go
jdheyburn: Good point, I think I missed this being so focused on the valkey nodes child resources having the correct labels. I'll get this raised.

internal/controller/valkeycluster_controller.go
jdheyburn: I've done some more reading about this. Shortnames have to be globally unique across a cluster, so reducing it to `vn` for valkeynode or `vc` for valkeycluster might be a bit ambiguous and cause conflicts with another CRD the user might've deployed to the cluster. 

I think then we should:

- omit shortnames entirely 
- if we do want shortnames - then in the helm chart, allow the user to define their own shortname for the CRDs

We can then set a sensible default later on if we choose to.

internal/controller/valkeycluster_controller.go
```diff
@@ -335,7 +333,7 @@ func (r *ValkeyClusterReconciler) upsertService(ctx context.Context, cluster *va
 		Spec: corev1.ServiceSpec{
 			Type:      corev1.ServiceTypeClusterIP,
 			ClusterIP: "None",
-			Selector:  labels(cluster),
+			Selector:  map[string]string{LabelCluster: cluster.Name},
```
sandeepkunusoth, line 336: only cluster.Name label is sufficient here?

sandeepkunusoth (approved):

internal/controller/valkeycluster_controller.go
jdheyburn: Correct, `valkey.io/cluster` is added to all created pods under this ValkeyNode. I think this helps reduce cognitive load on what labels are being used for what.

test/e2e/valkeycluster_test.go
jdheyburn: Actually, the e2e tests appear to be passing for the time being. I think I would rather focus on getting ValkeyNode.Status.Role to be more reliable, then use that as a means of retrieving a replica to delete. Do you think that's ok?

If yes I'll cut a new issue to follow up on this after this PR is merged (I feel it could be solved with the better Config/ACL reconciliation. Otherwise I'll get a helper util added, no problem.

jdheyburn: We are using node-index=0 in another test too, which seems to work as well.

internal/controller/valkeycluster_controller.go
jdheyburn: I removed them in https://github.com/valkey-io/valkey-operator/pull/113/commits/3d60d550e12775c3b7ee86d0fbcdceab1dcf0e0d

bjosv: Good call, I like it

test/e2e/valkeycluster_test.go
bjosv: This is good, the cluster was just started so it should be ok. We will probably need to add some chaos testing later on where we need to keep checking roles after each event.

test/e2e/valkeycluster_test.go
bjosv: We can keep it as it is.

PR #116, opened 2026-03-17
feat: Sequentially update workloads one-by-one

jdheyburn:
- ValkeyCluster controller now reconciles spec changes onto ValkeyNode CRs
  one at a time (replicas before primaries, descending node-index order),
  gating each step on the previous node's workload being fully rolled out
- ValkeyNode controller stamps status.observedGeneration on every reconcile
  so ValkeyCluster can detect unprocessed spec changes; tracks rollout
  completion via isWorkloadRolledOut (StatefulSet revision equality /
  Deployment updatedReplicas gate) using APIReader to bypass cache
- Fix upsertService and upsertConfigMap to use controllerutil.CreateOrUpdate,
  preventing update failures on second reconcile after operator restart
- Extract conditionsChanged helper to deduplicate status comparison logic
- Add integration tests for isWorkloadRolledOut, buildClusterValkeyNode
  propagation, condition ObservedGeneration tracking, and rolling-update
  sequencing in the ValkeyCluster controller
- Add e2e tests: StatefulSet resourceVersion stability, ObservedGeneration
  tracking, rolling update readiness gate, workloadType immutability, and
  ValkeyCluster rolling update end-to-end
- Updated status-conditions documentation

Note: node-index is used to determine if it is a replica or not. A follow up PR will enhance this to use a live replica instead.

I think also the pods are being rolled too fast, so we might want to revisit the readiness checks, or introduce some additional cluster health checks when reconciling pods to ensure we're not adding fuel to the fire.
**Update:** I've got a follow up PR after this one that will stabilise this

### DIFF
--- a/docs/status-conditions.md
+++ b/docs/status-conditions.md
@@ -51,9 +51,11 @@ Indicates whether the cluster is fully functional and serving traffic.
 Common reasons when `Ready=False`:
 - `ServiceError` – failed to create/update headless service
 - `ConfigMapError` – failed to create/update configuration
-- `DeploymentError` – failed to create/update deployments
-- `PodListError` – failed to list pods
+- `UsersACLError` – failed to reconcile ACL users/secrets
+- `ValkeyNodeError` – failed to create/update ValkeyNode CRs
+- `ValkeyNodeListError` – failed to list ValkeyNodes
 - `Reconciling` – controller is making changes
+- `UpdatingNodes` – rolling update of ValkeyNode CRs in progress
 - `MissingShards` – waiting for all shards to be created
 - `MissingReplicas` – waiting for all replicas to be created
 
@@ -72,6 +74,7 @@ Common reasons:
 - `Initializing` – initial cluster creation
 - `Reconciling` – general reconciliation in progress
 - `AddingNodes` – adding nodes to the cluster
+- `UpdatingNodes` – rolling update of ValkeyNode CRs in progress
 - `RebalancingSlots` – rebalancing hash slots across primaries (scale-out and scale-in)
 - `ReconcileComplete` – reconciliation finished (typically with `status=False`)
 
@@ -85,8 +88,6 @@ Indicates whether the cluster is impaired but may still be partially functional.
 
 Common reasons:
 - `NodeAddFailed` – failed to add a node to the cluster
-- `PrimaryLost` – primary lost in one or more shards
-- `NoSlotsAvailable` – no unassigned slots available for new shard
 - `RebalanceFailed` – slot rebalancing failed (scale-out or scale-in)
 
 ---
@@ -126,17 +127,19 @@ The high-level `state` is derived from conditions (priority order):
 
 1. `Degraded=True` → `state=Degraded`
 2. `Ready=True` → `state=Ready`
-3. `Progressing=True` and cluster already has shards → `state=Reconciling`
-4. `Progressing=True` and new cluster (no shards yet) → `state=Initializing`
-5. `Ready=False` with no other stronger signal → `state=Failed`
+3. `Progressing=True` → `state=Reconciling`
+4. `Ready=False` with no other stronger signal → `state=Failed`
+
+> **Note:** `Initializing` is the kubebuilder default for `status.state` and is visible briefly on a brand-new cluster before the controller first updates the status to `Reconciling`.
 
 ### Visual flow
 
 ```mermaid
 stateDiagram-v2
-  [*] --> Initializing: New cluster / Progressing=True
-  Initializing --> Reconciling: Progressing=True (making changes)
+  [*] --> Initializing: CRD default (before first reconcile)
+  Initializing --> Reconciling: Controller first runs / Progressing=True
   Reconciling --> Ready: Ready=True
+  Ready --> Reconciling: Spec changed / Progressing=True
   Ready --> Degraded: Issues detected / Degraded=True
   Degraded --> Ready: Issues cleared / Degraded=False and Ready=True
   Reconciling --> Failed: Unrecoverable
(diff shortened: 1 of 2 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -375,91 +372,129 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
-	created := 0
-	expected := int(cluster.Spec.Shards) * nodesPerShard
+	totalCreated := 0
+
 	for shardIndex := range int(cluster.Spec.Shards) {
-		for nodeIndex := range nodesPerShard {
-			if err := r.ensureValkeyNode(ctx, cluster, shardIndex, nodeIndex, expected, &created); err != nil {
-				return err
+		// Iterate nodeIndex in reverse order (replicas before primary)
+		for nodeIndex := nodesPerShard - 1; nodeIndex >= 0; nodeIndex-- {
```
jdheyburn, line 425: A follow up PR will include logic to select a real replica, instead of via the nodeIndex

internal/controller/valkeycluster_controller.go
```diff
@@ -127,10 +128,15 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	if err := r.upsertValkeyNodes(ctx, cluster); err != nil {
+	if requeue, err := r.reconcileValkeyNodes(ctx, cluster); err != nil {
 		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonValkeyNodeError, err.Error(), metav1.ConditionFalse)
 		_ = r.updateStatus(ctx, cluster, nil)
 		return ctrl.Result{}, err
+	} else if requeue {
+		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonUpdatingNodes, "Updating ValkeyNodes", metav1.ConditionFalse)
+		setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonUpdatingNodes, "Updating ValkeyNodes", metav1.ConditionTrue)
+		_ = r.updateStatus(ctx, cluster, nil)
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
bjosv: We use 2 seconds RequeueAfter in other situations, but 5 is a better choice here?

internal/controller/valkeycluster_controller.go
```diff
@@ -375,91 +372,129 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
 		return err
 	}
-	if err := r.Create(ctx, cm); err != nil {
-		if apierrors.IsAlreadyExists(err) {
-			if err := r.Update(ctx, cm); err != nil {
-				r.Recorder.Eventf(cluster, cm, corev1.EventTypeWarning, "ConfigMapUpdateFailed", "UpdateConfigMap", "Failed to update ConfigMap: %v", err)
-				return err
-			}
-		} else {
-			r.Recorder.Eventf(cluster, cm, corev1.EventTypeWarning, "ConfigMapCreationFailed", "CreateConfigMap", "Failed to create ConfigMap: %v", err)
```
bjosv, line 400: Unless we can't trigger the event for `ConfigMapCreationFailed`, we should remove it as a status condition in docs/status-conditions.md and from test/e2e/valkeycluster_test.go.

internal/controller/valkeycluster_controller.go
```diff
@@ -375,91 +372,129 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
+		// Iterate nodeIndex in reverse order (replicas before primary)
+		for nodeIndex := nodesPerShard - 1; nodeIndex >= 0; nodeIndex-- {
+			requeue, nodeCreated, err := r.reconcileValkeyNode(ctx, cluster, shardIndex, nodeIndex)
+			if err != nil {
+				return false, err
+			}
+			if nodeCreated {
+				totalCreated++
+			}
+			if requeue {
```
bjosv, line 430: Check requeue (and return) before checking nodeCreated

internal/controller/valkeycluster_controller.go
```diff
@@ -375,91 +372,129 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
+		return true, false, nil
+	case controllerutil.OperationResultNone:
+		if node.Status.ObservedGeneration > 0 && node.Generation != node.Status.ObservedGeneration {
+			log.V(1).Info("ValkeyNode spec not yet observed by controller, waiting",
+				"name", node.Name,
+				"generation", node.Generation,
+				"observedGeneration", node.Status.ObservedGeneration)
+			return true, false, nil
 		}
-		r.Recorder.Eventf(cluster, node, corev1.EventTypeWarning, "ValkeyNodeCreationFailed", "CreateValkeyNode", "Failed to create ValkeyNode: %v", err)
```
bjosv, line 457: ValkeyNodeCreationFailed is checked for in test/e2e/valkeycluster_test.go (to not have be emitted which it never will now). We should remove that part in the test then.

internal/controller/valkeycluster_controller.go
```diff
@@ -375,91 +372,129 @@ func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *
...
-	if err := r.Create(ctx, node); err != nil {
-		if apierrors.IsAlreadyExists(err) {
-			return nil
+	result, err := controllerutil.CreateOrUpdate(ctx, r.Client, node, func() error {
+		node.Labels = desired.Labels
+		node.Spec = desired.Spec
+		return controllerutil.SetControllerReference(cluster, node, r.Scheme)
+	})
+	if err != nil {
+		r.Recorder.Eventf(cluster, node, corev1.EventTypeWarning, "ValkeyNodeFailed", "ReconcileValkeyNode", "Failed to reconcile ValkeyNode: %v", err)
```
bjosv, line 464: We have documentation for event in `docs/status-conditions.md` under the section `## Kubernetes Events`. These event types are new and should be added there I guess.

jdheyburn: Fair point, I've done an audit to see what needs to be removed/updated/added.

internal/controller/valkeycluster_controller.go
jdheyburn: Added an issue for this:

- https://github.com/valkey-io/valkey-operator/issues/123

PR #128, opened 2026-04-03
feat: proactive failovers before cluster rolls

jdheyburn:
Before a pod is removed, it is first checked if it is currently a primary of a shard. If it is, then it fails over to a healthy replica.

This should help improve the stability of the cluster as rolls are introduced.

### DIFF
--- a/internal/controller/failover.go
+++ b/internal/controller/failover.go
@@ -1,0 +1,137 @@
+/*
+Copyright 2025 Valkey Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/client-go/tools/events"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
+	"valkey.io/valkey-operator/internal/valkey"
+)
+
+const (
+	proactiveFailoverTimeout = 10 * time.Second
+	proactiveFailoverPoll    = 1 * time.Second
+
+	eventReasonFailoverInitiated = "FailoverInitiated"
+	eventReasonFailoverFailed    = "FailoverFailed"
+	eventReasonFailoverTimeout   = "FailoverTimeout"
+	eventReasonFailoverCompleted = "FailoverCompleted"
+	eventActionProactiveFailover = "ProactiveFailover"
+)
+
+// findFailoverShard returns the shard and its synced replicas if the node at
+// address is a primary that should be gracefully failed over before being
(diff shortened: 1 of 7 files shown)


internal/controller/failover.go
```diff
@@ -0,0 +1,137 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-timer.C:
+			recorder.Eventf(cluster, nil, corev1.EventTypeWarning, eventReasonFailoverTimeout, eventActionProactiveFailover, "Failover to %s in shard %s did not complete within %s", target.Address, shard.Id, proactiveFailoverTimeout)
+			return fmt.Errorf("failover to %s timed out after %s", target.Address, proactiveFailoverTimeout)
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
```
bjosv, line 97: It might be a bit unusual that we poll instead of using `ctrl.Result{RequeueAfter`, but I guess it wont be easy to keep state to know exactly what we wait for in each reconcile.. I guess we need to do this.

If we could use RequeueAfter somehow then we probably could trigger failovers on multiple shards in parallell.

internal/controller/valkeycluster_controller.go
```diff
@@ -426,16 +426,32 @@ aclfile /config/users/users.acl`,
...
+	// across the loop because nodes are iterated replicas-first (the primary
+	// is last in each shard), and after an update we requeue immediately,
+	// re-scraping fresh state.
+	var clusterState *valkey.ClusterState
+	if anyNodeRequiresRoll(cluster, nodes) {
+		operatorPassword, err := fetchSystemUserPassword(ctx, operatorUser, r.Client, cluster.Name, cluster.Namespace)
+		if err != nil {
+			return false, fmt.Errorf("failed to fetch operator password for proactive failover: %w", err)
+		}
+		clusterState = r.getValkeyClusterState(ctx, nodes, operatorUser, operatorPassword)
```
bjosv, line 447: We might scrap twice in a reconcile loop, but maybe reconcileValkeyNode always triggers a requeue before the second/legacy scrape?

internal/controller/failover.go
```diff
@@ -0,0 +1,137 @@
...
+
+	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
+	"valkey.io/valkey-operator/internal/valkey"
+)
+
+const (
+	proactiveFailoverTimeout = 10 * time.Second
+	proactiveFailoverPoll    = 1 * time.Second
+
+	eventReasonFailoverInitiated = "FailoverInitiated"
```
bjosv, line 37: We should add these in docs/status-conditions.md

bjosv (dismissed):
LGTM, just update docs/status-conditions.md (or so it in next PR depending on your plan)

internal/controller/failover.go
jdheyburn: Failovers are stateful, and I think it might be best to keep this failover logic in one reconcile loop so that we can track it accordingly. Otherwise as you mentioned, we may end up with multiple failovers when we don't want them. Let me know if there are any suggestions for improvements.

internal/controller/valkeycluster_controller.go
jdheyburn: I opted for a double scrape for now, given that pod rolls won't happen all that often. I did have an attempt at a single clusterState but it added a bit of complexity to the code and it wasn't a clean design.

I think we need to refactor how this Cluster State is passed around into somewhere that is more central.

internal/controller/failover.go
jdheyburn: I'll get it added and request a re-review. I wonder if its worth having a CI check it ensure that status conditions and events are documented.

internal/controller/valkeycluster_controller.go
bjosv: Sounds good, yes there we can refactor it a bit later

bjosv (approved):
LGTM!

Issue #110, opened 2026-03-10
feat: Add default system users for operator and exporter authentication


jdheyburn:
## Summary

Create two default system users (`_operator` and `_exporter`) that are automatically injected into every `ValkeyCluster`'s ACL configuration. This ensures the operator and metrics exporter can authenticate to Valkey independently of any user-defined ACLs, so that changes to `spec.users[]` cannot accidentally break operator or exporter connectivity.

## Motivation

Currently the operator and exporter sidecar connect to Valkey unauthenticated. If a cluster admin defines users and disables the `default` user, the operator loses the ability to manage the cluster. System users with auto-generated credentials solve this by giving the operator and exporter their own stable identities.

## Design decisions

- **`default` user is not touched.** Cluster admins manage it themselves via `spec.users[]`.
- **System users are always injected**, regardless of whether `spec.users[]` is empty or populated.
- **Passwords are auto-generated** (32 random bytes, hex-encoded) on first reconcile and stored in an operator-managed Secret. No admin setup required. Passwords are stored as plaintext hex strings in the Secret. At ACL generation time, they are SHA256-hashed and prefixed with `#` in the ACL line, consistent with existing user password handling in `buildUserAcl`.
- **Credentials are unique per `ValkeyCluster`** to limit blast radius if a Secret is compromised.
- **`_operator` has broad permissions (`+@all`)** for now since the operator's command surface is still evolving. Tightening permissions is a future issue.
- **`_exporter` has locked-down permissions** matching the [redis_exporter recommended ACL](https://github.com/oliver006/redis_exporter#authenticating-with-redis).
- **Names prefixed with `_`** are reserved for system users. `spec.users[]` entries starting with `_` are rejected at admission time via a CEL validation rule on the CRD.
- **`_exporter` user is only created when `exporter.enabled` == true

### `_operator`

```acl
user _operator on #<sha256-of-password> ~* &* +@all
```

Used by the operator's Valkey client to issue cluster management commands (`CLUSTER MEET`, `CLUSTER ADDSLOTSRANGE`, `CLUSTER REPLICATE`, `CLUSTER NODES`, `INFO`, etc.).

### `_exporter`

```acl
user _exporter on #<sha256-of-password> ~* &* +@connection +memory -readonly +strlen +config|get +xinfo +pfcount +zcard +type +xlen +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes +info +latency +scan +client
```

[...]

PR #129, opened 2026-04-07
Add default system users (#110)

hieu2102:
implements #110
- add defaults system users `_operator`  and `_exporter`
    - update `GetClusterState`, `GetNodeState` functions to connect to Valkey using `_operator` user
    - fall back to `default` (unauthenticated on WRONGPASS error) 
    - `metrics-exporter` container will connect using `_exporter` user (by setting `REDIS_USER`, `REDIS_PASSWORD` env variables)
- add CEL validation to reject users with named started with `_`
- add ACL hash annotation to pod template, when the internal ACL secret (`internal-<clusterName>-acl`) changed, the pods will be recreated automatically

### DIFF
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -173,6 +173,7 @@ const (
 	ReasonRebalanceFailed     = "RebalanceFailed"
 	ReasonUsersAclError       = "UsersACLError"
 	ReasonUpdatingNodes       = "UpdatingNodes"
+	ReasonSystemUsersAclError = "SystemUsersACLError"
 )
 
 // +kubebuilder:object:root=true
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -100,7 +100,7 @@ type ExporterSpec struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
 	// Enable or disable the exporter sidecar container
-	Enabled bool `json:"enabled"`
+	Enabled bool `json:"enabled,omitempty"`
 }
 
 // ValkeyClusterStatus defines the observed state of ValkeyCluster.
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -100,7 +100,7 @@ type ExporterSpec struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
 	// Enable or disable the exporter sidecar container
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled bool `json:"enabled"`
 }
 
 // ValkeyClusterStatus defines the observed state of ValkeyCluster.
--- a/internal/controller/metrics_exporter.go
+++ b/internal/controller/metrics_exporter.go
@@ -25,7 +25,7 @@ import (
 )
 
 // generateMetricsExporterContainerDef generates the container definition for the metrics exporter sidecar.
-func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec, labels map[string]string) corev1.Container {
+func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec, clusterName string) corev1.Container {
 	exporterImage := DefaultExporterImage
 	if exporter.Image != "" {
 		exporterImage = exporter.Image
@@ -40,7 +40,7 @@ func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec,
 				ValueFrom: &corev1.EnvVarSource{
 					SecretKeyRef: &corev1.SecretKeySelector{
 						LocalObjectReference: corev1.LocalObjectReference{
-							Name: getSystemPasswordSecretName(labels[LabelCluster]),
+							Name: getSystemPasswordSecretName(clusterName),
 						},
 						Key: exporterUser,
 					},
--- a/internal/controller/metrics_exporter.go
+++ b/internal/controller/metrics_exporter.go
@@ -25,7 +25,7 @@ import (
 )
 
 // generateMetricsExporterContainerDef generates the container definition for the metrics exporter sidecar.
-func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec) corev1.Container {
+func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec, labels map[string]string) corev1.Container {
 	exporterImage := DefaultExporterImage
 	if exporter.Image != "" {
 		exporterImage = exporter.Image
@@ -33,7 +33,21 @@ func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec)
 	return corev1.Container{
 		Name:  "metrics-exporter",
 		Image: exporterImage,
-		Args:  []string{fmt.Sprintf("--redis.addr=localhost:%d", DefaultPort)},
+		Env: []corev1.EnvVar{
+			{Name: "REDIS_USER", Value: exporterUser},
+			{
+				Name: "REDIS_PASSWORD",
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: getSystemPasswordSecretName(labels[LabelCluster]),
+						},
+						Key: exporterUser,
+					},
+				},
+			},
+		},
+		Args: []string{fmt.Sprintf("--redis.addr=localhost:%d", DefaultPort)},
 		Ports: []corev1.ContainerPort{
 			{
 				Name:          "metrics",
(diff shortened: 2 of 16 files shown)


internal/controller/metrics_exporter.go
```diff
@@ -25,15 +25,29 @@ import (
...
 		Image: exporterImage,
-		Args:  []string{fmt.Sprintf("--redis.addr=localhost:%d", DefaultPort)},
+		Env: []corev1.EnvVar{
+			{Name: "REDIS_USER", Value: exporterUser},
+			{
+				Name: "REDIS_PASSWORD",
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: getSystemPasswordSecretName(labels[LabelCluster]),
```
jdheyburn: nit: I wonder if we can pass through the clusterName instead of all the labels.

internal/controller/users.go
```diff
@@ -39,6 +40,16 @@ const (
 	aclFilename       = "users.acl"
 )
 
+var (
+	operatorUser    = "_operator"
+	exporterUser    = "_exporter"
+	systemUsers     = []string{operatorUser, exporterUser}
+	systemUsersAcls = map[string]string{
+		operatorUser: "+@all",
+		exporterUser: "-@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking",
```
jdheyburn, line 53: Perhaps we could leave a comment linking to the exporter docs where this is declared?

internal/controller/users.go
```diff
@@ -310,8 +304,126 @@ func fetchUserPasswords(ctx context.Context, user valkeyiov1alpha1.UserAclSpec,
...
 // If so, we assume this is a pre-hashed sha256 password.
 func isPreHashedPassword(password []byte) bool {
 	return password[0] == 35 && len(password) == 65
 }
+
+func generatePassword() string {
+	randstr := rand.Text()
+	// hasher := sha256.New()
+	// hasher.Write([]byte(randstr))
+	// hashBytes := hasher.Sum(nil)
```
jdheyburn: Is this commented code required?

internal/controller/users.go
```diff
@@ -310,8 +304,126 @@ func fetchUserPasswords(ctx context.Context, user valkeyiov1alpha1.UserAclSpec,
...
+}
+
 // Check if byte-string begins with # (byte 35) and is 65 total characters long.
 // If so, we assume this is a pre-hashed sha256 password.
 func isPreHashedPassword(password []byte) bool {
 	return password[0] == 35 && len(password) == 65
 }
+
+func generatePassword() string {
+	randstr := rand.Text()
```
jdheyburn: What does this generate for us? i.e. password length, characters, etc.

internal/controller/valkeycluster_controller.go
```diff
@@ -151,7 +151,14 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
-	state := r.getValkeyClusterState(ctx, nodes)
+	operatorPassword, err := fetchSystemUserPassword(ctx, operatorUser, r.Client, cluster.Name, cluster.Namespace)
+	if err != nil {
+		log.Error(err, "failed to retrieve system user password")
+		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonSystemUsersAclError, err.Error(), metav1.ConditionFalse)
+		_ = r.updateStatus(ctx, cluster, nil)
+		return ctrl.Result{}, nil
+	}
+	state := r.getValkeyClusterState(ctx, nodes, operatorUser, operatorPassword)
```
jdheyburn, line 161: If we had one operator password per operator, would that help simplify this?

internal/controller/users.go
```diff
@@ -310,8 +304,126 @@ func fetchUserPasswords(ctx context.Context, user valkeyiov1alpha1.UserAclSpec,
 	return passwords, nil
 }
 
+func fetchSystemUserPassword(ctx context.Context, username string, apiClient client.Client, clusterName, clusterNamespace string) (string, error) {
+	systemSecret := &corev1.Secret{}
+	err := apiClient.Get(ctx, types.NamespacedName{
+		Namespace: clusterNamespace,
+		Name:      getSystemPasswordSecretName(clusterName),
+	}, systemSecret)
```
jdheyburn, line 316: I wonder about all the kube API calls that would need to be made for it to retrieve the operator password on the reconcile loop. Is there a way we could cache it locally?

jdheyburn (commented):
Thanks for taking a look at this! I've not tested this myself yet, but left some comments.

internal/controller/metrics_exporter.go
hieu2102: Yeah, I was just trying to future proof things when passing all labels

internal/controller/users.go
hieu2102: the operator manager already cache it by default. I've tested with this audit policy, and it only `get` the secret once if it is not changed

```yaml
apiVersion: audit.k8s.io/v1
kind: Policy
rules:
  # Capture every interaction with Secret objects, including request and response bodies
  - level: RequestResponse
    resources:
      - group: ""
        resources: ["secrets"]

  # Optional: catch Secret subresources too, if any appear in your environment/tooling
  - level: RequestResponse
    resources:
      - group: ""
        resources: ["secrets/*"]

  # Do not log anything else
  - level: None
```

internal/controller/users.go
hieu2102: I've added a comment about this in the latest commit

internal/controller/valkeycluster_controller.go
hieu2102: Do you mean creating a separate secret for the operator user in the `valkey-operator-system` namespace? 
I think that would add to the administration burden if we have to remember/find the passwords for system users in multiple namespaces

internal/controller/users.go
```diff
@@ -45,7 +45,9 @@ var (
 	exporterUser    = "_exporter"
 	systemUsers     = []string{operatorUser, exporterUser}
 	systemUsersAcls = map[string]string{
+		// TODO: tighten the permission for the operator user
```
jdheyburn, line 50: Can you raise an issue for this please?

internal/controller/users.go
jdheyburn: I checked the generated value. Could we have it to be upper and lower case letters too? Today its just uppercase.

internal/controller/users.go
```diff
@@ -310,8 +306,127 @@ func fetchUserPasswords(ctx context.Context, user valkeyiov1alpha1.UserAclSpec,
...
+func (r *ValkeyClusterReconciler) upsertSystemUsersPasswordSecret(ctx context.Context, apiClient client.Client, cluster *valkeyiov1alpha1.ValkeyCluster) (*corev1.Secret, error) {
+	log := logf.FromContext(ctx)
+
+	systemUsersSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getSystemPasswordSecretName(cluster.Name),
+			Namespace: cluster.Namespace,
+			Labels:    labels(cluster),
+		},
+		Data: map[string][]byte{},
```
jdheyburn, line 353: Do we also need to specify the type here like what we did for the [ACLs](https://github.com/valkey-io/valkey-operator/pull/117)? @utdrmac what do you think?

jdheyburn (commented):
Looks good! I've some minor comments, but would also appreciate @utdrmac taking a look since they raised the original ACL feature.

internal/controller/valkeycluster_controller.go
jdheyburn: Yeah fair enough. If it's caching the per-cluster password then I'm sure that's fine.

internal/controller/users.go
hieu2102: #132

internal/controller/users.go
hieu2102: Good point, let me add it

api/v1alpha1/valkeycluster_types.go
```diff
@@ -100,7 +100,7 @@ type ExporterSpec struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
 	// Enable or disable the exporter sidecar container
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled bool `json:"enabled"`
```
jdheyburn: Is there a particular reason this is now required?

hieu2102: No, while working on this, my tests were failing because the `Enabled` field does not exist. It doesn't happen anymore, let me revert this one

jdheyburn (approved):
Thanks for working on this!

daanvinken:
I think we missed one here https://github.com/valkey-io/valkey-operator/pull/137

PR #130, opened 2026-04-08
chore: split IMG use for vars

jdheyburn:
Continuation of the feedback in below:

- https://github.com/valkey-io/valkey-operator/pull/106#discussion_r2970466964

cc @sandeepkunusoth - let me know if this was what you had in mind

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM golang:1.25 AS builder
 ARG TARGETOS
 ARG TARGETARCH
-ARG LDFLAGS
+ARG VERSION_LDFLAGS
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM golang:1.25 AS builder
 ARG TARGETOS
 ARG TARGETARCH
-ARG LDFLAGS
+ARG VERSION_LDFLAGS
```
sandeepkunusoth, line 5: Was change need in this variable version_?

sandeepkunusoth (approved):

Dockerfile
jdheyburn: Not necessarily, but we were inconsistent with using `LDFLAGS` and `VERSION_LDFLAGS`. For this I opted for the latter.

PR #133, opened 2026-04-12
feat: initial tls support for valkey cluster

sandeepkunusoth:
This PR Adds initial TLS/SSL Support for Valkey Cluster https://github.com/valkey-io/valkey-operator/issues/59.

### Summary

This PR introduces initial TLS support for Valkey clusters(SERVER SIDE).

---

### Changes

- Adds TLS support for Valkey cluster nodes
- Updates TLS configuration options in CRD for Certificate references.
- Updates Valkey configuration to include `tls-port`, `port 0`, `tls-cert-file`, `tls-key-file`, `tls-ca-cert-file`
-  fixed comments from old PR https://github.com/valkey-io/valkey-operator/pull/91

---

### Implementation

  - Added TLS configuration fields (TLSConfig and CertificateRef) types to ValkeyClusterSpec.
  - Mounts TLS secrets as volumes into Pod spec
  - Updates liveliness, readiness scripts with TLS flags

---

### Testing
  - Add e2e tests for TLS-enabled cluster creation

---

### Next steps
  - This PR only supports using exisiting certificates in the cluster. we can either use cert manager to create certificate. this will be done as part of seperate PR.
  - cert hot reload during cert renewal
---
### Checklist


api/v1alpha1/valkeycluster_types.go
```diff
@@ -88,6 +88,27 @@ type ValkeyClusterSpec struct {
...
+	// - `ca.crt`: The certificate authority.
+	// - `tls.crt`: The certificate (or a chain).
+	// - `tls.key`: The private key to the first certificate in the certificate chain.
+	Certificate CertificateRef `json:"certificate,omitempty"`
+}
+
+// CertificateRef defines the certificate reference for ValkeyCluster.
+type CertificateRef struct {
+	// SecretName is the name of the secret.
+	SecretName string `json:"secretName,omitempty"`
```
jdheyburn, line 111: I've never used TLS with Valkey. Is it common to use the same certificate across all the available configs?

daanvinken: Agreed it probably makes sense to have separate config for server, client and perhaps even replication / cluster bus?

I think we can iterate on this, but having the naming backwards compatible on those iterations may be useful.

daanvinken (commented):
Should we have a watcher or use an [annotation](https://github.com/valkey-io/valkey-operator/blob/5b9adec6e8277c11a4523031165c7258b216e47a/internal/controller/valkeynode_controller.go#L132) to make sure cert renewals roll the cluster?

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: are u asking?  @jdheyburn if certificates are shared across multiple diff clusters?
no each cluster will be assigned its own certificate and in our case, we have certificate subject with dns of instance and it will be unique.


@daanvinken the scope of current PR is only to add server side certs. are u saying that these server certs can be reused later to perform auth during replication? why do we need do handle client certs as part of valkey-operator?

sandeepkunusoth:
> Should we have a watcher or use an [annotation](https://github.com/valkey-io/valkey-operator/blob/5b9adec6e8277c11a4523031165c7258b216e47a/internal/controller/valkeynode_controller.go#L132) to make sure cert renewals roll the cluster?

we have added watcher for secrets managed by cert-manager and hash annotation to check changes in certs in our internal operator and performed config set on tls cert renewals. this can be added in next iterations as this will be seperate enhancement.


we can also use tls-auto-reload-interval 86400 to automatically reload certs https://valkey.io/topics/tls/#:~:text=tls%2Dauto%2Dreload%2Dinterval%2086400

api/v1alpha1/valkeycluster_types.go
daanvinken: Fair point. I was coming from the CNPG world where mTLS is the default auth mechanism, but Valkey relies more on ACL-based auth so client certs are less relevant here. And even if we'd ever need a separate CA for clients, which I now realize is rare for Valkey, we can easily add that on top of this.

The server-side TLS scope makes sense.

daanvinken (approved):

internal/controller/valkeynode_controller.go
```diff
@@ -372,11 +373,25 @@ func (r *ValkeyNodeReconciler) getPod(ctx context.Context, node *valkeyiov1alpha
...
-func getValkeyRole(ctx context.Context, podIP string, port int) string {
-	c, err := vclient.NewClient(vclient.ClientOption{
-		InitAddress:       []string{fmt.Sprintf("%s:%d", podIP, port)},
-		ForceSingleClient: true,
-	})
+func (r *ValkeyNodeReconciler) getValkeyRole(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) string {
+	var tlsConfig *tls.Config
+	if node.Spec.TLS != nil && node.Spec.TLS.Certificate.SecretName != "" {
+		cluster := &valkeyiov1alpha1.ValkeyCluster{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: node.Namespace, Name: node.Labels["valkey.io/cluster"]}, cluster); err == nil {
```
bjosv:
```suggestion
		if err := r.Get(ctx, client.ObjectKey{Namespace: node.Namespace, Name: node.Labels[LabelCluster]}, cluster); err == nil {
```

internal/controller/valkeynode_controller.go
```diff
@@ -372,11 +373,25 @@ func (r *ValkeyNodeReconciler) getPod(ctx context.Context, node *valkeyiov1alpha
...
 // ("primary" or "replica"). Returns an empty string if the role cannot be determined.
-func getValkeyRole(ctx context.Context, podIP string, port int) string {
-	c, err := vclient.NewClient(vclient.ClientOption{
-		InitAddress:       []string{fmt.Sprintf("%s:%d", podIP, port)},
-		ForceSingleClient: true,
-	})
+func (r *ValkeyNodeReconciler) getValkeyRole(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) string {
+	var tlsConfig *tls.Config
+	if node.Spec.TLS != nil && node.Spec.TLS.Certificate.SecretName != "" {
+		cluster := &valkeyiov1alpha1.ValkeyCluster{}
```
bjosv: This is creating a dependency in valkeynode_controller.go to the cluster code, so we won't able to start a standalone replication setup (since we use r.Get() to search for a ValkeyCluster CRD)

Could it be better to let GetTLSConfig take what is needed from cluster (secretName, serverName and namespace)?

Hum, maybe there still would be some dependency, but at least we wont need to do a lookup if we get the serverName like:
serverName := fmt.Sprintf("%s.%s.svc.cluster.local", node.Labels[LabelCluster], node.Namespace)

We could use a different servername when LabelCluster does not exist then

internal/controller/utils.go
```diff
@@ -227,3 +242,66 @@ func countSlots(ranges []valkey.SlotsRange) int {
...
+func generateValkeyConfig(cluster *valkeyv1.ValkeyCluster) string {
+	config := `cluster-enabled yes
+protected-mode no
+cluster-node-timeout 2000
+aclfile /config/users/users.acl`
+
+	if cluster.Spec.TLS != nil {
+		config += fmt.Sprintf(`
+tls-port %d
+port 0
```
bjosv, line 256: Enabling TLS will break the operator deployed metrics exporter I guess?

sandeepkunusoth: added e2e test to verify metrics is working with tls

internal/controller/valkeynode_controller.go
sandeepkunusoth: updated

bjosv (approved):
LGTM!

jdheyburn:
One thing missing from this PR is documentation, cc @sandeepkunusoth

PR #137, opened 2026-04-15
fix: authenticate getValkeyRole with _operator credentials

daanvinken:
This PR is a follow-up to #129 which added `_operator` system user authentication to the cluster controller but missed the `getValkeyRole` code path in the ValkeyNode controller.

### Summary

`getValkeyRole` connects to Valkey unauthenticated to check the replication role. Once the default user is restricted or disabled via `spec.users`, the role check fails silently and `ValkeyNode.status.role` is always empty.

### Implementation

Pass `_operator` credentials to `getValkeyRole`, matching how the cluster controller authenticates in `GetClusterState`. If the system password secret doesn't exist yet (fresh cluster before ACLs are set up), the password is empty and the connection falls back to unauthenticated.

### Limitations

None.

### Testing

Existing E2E tests cover this path (role is populated in ValkeyNode status). No behavioral change when the default user is open.

### Checklist

- [NA] Tests are added or updated.
- [NA] Documentation files are updated.

### DIFF
--- a/internal/controller/valkeynode_controller.go
+++ b/internal/controller/valkeynode_controller.go
@@ -388,10 +388,23 @@ func (r *ValkeyNodeReconciler) getValkeyRole(ctx context.Context, node *valkeyio
 		}
 	}
 
+	// Use _operator credentials when available (cluster-managed nodes).
+	// Standalone ValkeyNodes have no cluster label and no system password
+	// secret, so we skip the lookup entirely and connect unauthenticated.
+	var username, operatorPassword string
+	if clusterName, ok := node.Labels[LabelCluster]; ok {
+		operatorPassword, _ = fetchSystemUserPassword(ctx, operatorUser, r.Client, clusterName, node.Namespace)
+		if operatorPassword != "" {
+			username = operatorUser
+		}
+	}
+
 	opt := vclient.ClientOption{
 		InitAddress:       []string{fmt.Sprintf("%s:%d", node.Status.PodIP, DefaultPort)},
 		ForceSingleClient: true, // Don't connect to another cluster node.
 		TLSConfig:         tlsConfig,
+		Username:          username,
+		Password:          operatorPassword,
 	}
 
 	c, err := vclient.NewClient(opt)


jdheyburn:
Thanks for raising this PR! e2e tests are failing on:

```
Summarizing 2 Failures:
  [FAIL] ValkeyNode standalone StatefulSet [It] creates owned resources and populates status with role [valkeynode]
  /home/runner/work/valkey-operator/valkey-operator/test/e2e/valkeynode_test.go:129
  [FAIL] ValkeyNode standalone Deployment [It] creates owned resources and populates status with role [valkeynode]
  /home/runner/work/valkey-operator/valkey-operator/test/e2e/valkeynode_test.go:185

```

Which I find weird if this is supposed to fix this. Are you able to take a look @daanvinken ?

daanvinken:
Thanks for calling out. The standalone ValkeyNode e2e tests don't have a ValkeyCluster, so there's no system password secret. The original code always passed `_operator` as the username even with an empty password, which fails because that user doesn't exist on standalone nodes.

I considered wrapping it in a check for both the cluster label and a successful password fetch, but that added too much nesting (readability -1). 
```
      271 +      username := ""
      272 +      password := ""
      273 +      if clusterName := node.Labels[LabelCluster]; clusterName != "" {
      274 +        if pw, err := fetchSystemUserPassword(ctx, operatorUser, r.Client, clusterName, node.Namespace); err == nil && pw !=
          +"" {
      275 +          username = operatorUser
      276 +          password = pw
      277 +        }
      278 +      }
```

Simpler approach: only set the username when the password is non-empty, otherwise connect unauthenticated. Lmk if you feel differently

jdheyburn:
Thanks - when you rebase we'll get the CI checks ran

daanvinken:
@jdheyburn good to go!

internal/controller/valkeynode_controller.go
```diff
@@ -388,10 +388,22 @@ func (r *ValkeyNodeReconciler) getValkeyRole(ctx context.Context, node *valkeyio
 		}
 	}
 
+	// Use _operator credentials when available (cluster-managed nodes).
+	// Standalone ValkeyNodes have no system password secret, so we
+	// fall back to unauthenticated to avoid failing on AUTH.
+	clusterName := node.Labels[LabelCluster]
```
bjosv: We could also avoid the password fetching and do the fallback if we don't have any label, i.e like the TLS code above in the same function:
`if clusterName, ok := node.Labels[LabelCluster]; ok {`

daanvinken: Yeah this is cleaner. Updating

Thanks!

jdheyburn (approved):
LGTM, thank you!

daanvinken:
@jdheyburn ok to merge?

jdheyburn:
@daanvinken All good, thanks for the prompt!

PR #140, opened 2026-04-15
fix: surface reconcileUsersAcl errors in cluster status

### LINKED ISSUE #139: reconcileUsersAcl silently skips users with missing password secrets
This was brought in with https://github.com/valkey-io/valkey-operator/pull/82

When a user in `spec.users` references a password secret that doesn't exist, [`reconcileUsersAcl`](https://github.com/valkey-io/valkey-operator/blob/main/internal/controller/users.go#L169) logs an error but skips the user and returns nil. The ACL file is written without that user, and the cluster reports Ready.

Additionally, when `reconcileUsersAcl` does return an error (e.g. internal secret creation fails), the condition is set but `updateStatus` is never called here, so the error isn't persisted to the status either.

Reproduced with:
```yaml
spec:
  users:
    - name: testuser
      enabled: true
      passwordSecret:
        name: missing-secret
        keys: ["password"]
```

`kubectl get valkeycluster` shows Ready while the user doesn't exist in the ACL.

daanvinken:
This PR closes #139

### Summary

When a user in `spec.users` references a password secret that doesn't exist, [`reconcileUsersAcl`](https://github.com/valkey-io/valkey-operator/blob/main/internal/controller/users.go#L169) logs an error but skips the user and returns nil. The ACL file is written without that user, and the cluster reports Ready. I think it makes sense to fail hard.

Additionally, when `reconcileUsersAcl` does return an error (e.g. internal secret creation fails), the condition is set but `updateStatus` is never called here, so the error isn't persisted to the status either.

### Implementation

- `users.go`: replace `continue` with `return fmt.Errorf(...)` when `fetchUserPasswords` fails
- `valkeycluster_controller.go`: add missing `updateStatus` call on the `reconcileUsersAcl` error path

### Testing

Before the fix, a missing password secret is invisible:
```
NAME            STATE   REASON           AGE
acl-fail-test   Ready   ClusterHealthy   10m
```

After the fix:
```
 ༼ つ ▀_▀ ༽つ  ~  src  valkey-ope  k describe valkeycluster acl-fail-test | tail -n 15                        fix/acl-missing-secret-status  2✎  5+  ⎈ kind-valkey-repro  21:55:45
    Last Transition Time:  2026-04-15T19:51:06Z
    Message:               user testuser: no password or reference found
    Observed Generation:   1
    Reason:                UsersACLError
    Status:                False
    Type:                  Ready
  Message:                 user testuser: no password or reference found
  Ready Shards:            0
  Reason:                  UsersACLError
  Shards:                  0
  State:                   Failed
Events:
  Type    Reason          Age    From                      Message
  ----    ------          ----   ----                      -------
  Normal  ServiceCreated  4m46s  valkeycluster-controller  Created headless Service
 ༼ つ ▀_▀ ༽つ  ~  src  valkey-ope  k get valkeycluster                                                        fix/acl-missing-secret-status  2✎  5+  ⎈ kind-valkey-repro  21:55:53
NAME            STATE    REASON          AGE
acl-fail-test   Failed   UsersACLError   4m51s
```

### Limitations

None.

### Checklist

- [NA] Documentation files are updated.

### DIFF
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -165,8 +165,7 @@ func (r *ValkeyClusterReconciler) reconcileUsersAcl(ctx context.Context, cluster
 		// Get passwords from Secret
 		passwords, err := fetchUserPasswords(ctx, user, r.Client, cluster.Name, cluster.Namespace)
 		if err != nil {
-			log.Error(err, "failed to fetch password", "username", user.Name)
-			continue
+			return fmt.Errorf("user %s: %w", user.Name, err)
 		}
 
 		// Build ACL string for this user with found password(s)
@@ -248,8 +247,8 @@ func fetchUserPasswords(ctx context.Context, user valkeyiov1alpha1.UserAclSpec, 
 
 	log := logf.FromContext(ctx)
 
-	// If this user doesn't have a password, return empty
-	if user.NoPassword {
+	// If this user doesn't have a password or is disabled, return empty
+	if user.NoPassword || !user.Enabled {
 		return []string{}, nil
 	}
 	// Look for a Secret matching the user-provided name, or clusterName-users
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -124,6 +124,7 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	if err := r.reconcileUsersAcl(ctx, cluster); err != nil {
 		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonUsersAclError, err.Error(), metav1.ConditionFalse)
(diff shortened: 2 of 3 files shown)


daanvinken:
I considered failing soft (skip the user, set a `Degraded` condition and emit a warning event) so a missing secret doesn't block the whole cluster. 

However, the `Degraded` condition gets cleared when the cluster reaches `Ready` at the end of the reconcile, so the warning disappears from status. Checking for any warnings before marking the cluster as healthy feels messy, as we'd be adding semantics where `Degraded` means different things depending on who set it..

And the warning event fires every reconcile cycle unless you add dedup logic, but the above problem persists. The hard error is simpler and keeps the problem visible.

Any input here is welcome!

jdheyburn:
@daanvinken I ran CI checks, would you be able to take a look at those that have failed please?

daanvinken:
Yeah should be good now!

jdheyburn (approved):
Thank you!

PR #142, opened 2026-04-20
Enforce default shard anti-affinity

DharmendraChoudhary67:
## Summary
- add default required pod anti-affinity so nodes from the same shard cannot land on the same Kubernetes host
- merge the default shard rule with user-supplied affinity instead of replacing it
- extend unit and e2e coverage to prove same-shard primary/replica separation


internal/controller/valkeycluster_controller.go
```diff
@@ -543,6 +543,34 @@ func buildClusterValkeyNode(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex
...
+		affinity = cluster.Spec.Affinity.DeepCopy()
+	} else {
+		affinity = &corev1.Affinity{}
+	}
+
+	if affinity.PodAntiAffinity == nil {
+		affinity.PodAntiAffinity = &corev1.PodAntiAffinity{}
+	}
+
+	affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(
```
daanvinken: I don't think we should hardcode `RequiredDuringScheduling` here. If someone tries to run a 3 shard cluster on 2 nodes (or any other similar combination, e.g. testing clusters or CI), scheduling will fail with no obvious error. Maybe configurable but at least the default should be `Preferred`?

Where would this scheduling issue pop up? I think it's only an event on the pod itself and not even on the `Statefulset` (or `ValkeyCluster`). That may make it hard to debug.

daanvinken: In my book we should make this fully configurable and move the default to the spec examples. We can always add an opinionated default later on.

Note that we are using `append()` here, so your default can also not be overridden afaik.

jdheyburn: I agree we should have a soft default, with the option to enforce a strict mode.

One feature we would like for production is AZ-aware scheduling. i.e. scheduling primaries of a shard in one zone, and replicas in another. Perhaps that doesn't apply so much in a Cluster world, but it may exist in a Standalone one.

It might be worth opening up a [Design] issue so that we can hear from the community and come up with a good design for all.

DharmendraChoudhary67: Sure. I've made the changes for soft default and  will create a [design] issue.

daanvinken: All on board with a [design] issue, but commenting on the updated changes regardless if you don't mind.

- Shoudl we consider using Kubernetes' native `topologySpreadConstraints` instead of a custom `ShardAntiAffinitySpec`? The operator could expose `spec.topologySpreadConstraints` on the CRD (like it already does for `affinity`, `tolerations`, `nodeSelector`) and inject the shard label selector automatically. 
I feel conflicted about introducing this abstraction, using `topologySpreadConstraints` users get the full Kubernetes API (multiple topology keys, maxSkew, etc.).

- The `case ""` fallback to Preferred is a bit surprising. With `omitempty` on the Type field, `shardAntiAffinity: {}` silently gives you Preferred, while omitting `shardAntiAffinity` entirely gives you nothing. Might be cleaner to map empty string to Disabled, or remove the fallback since kubebuilder already validates the enum.

- `buildShardAntiAffinity()` is a pure function with no receiver, might fit better in `utils.go` or `valkeynode_resources.go` next to `buildClusterValkeyNode` where it's called from.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -64,11 +64,16 @@ type ValkeyClusterSpec struct {
...
-	// Some basic anti-affinity rules will be applied by default to spread pods across nodes and zones
+	// Affinity to apply to the pods, overrides NodeSelector if set.
 	// +optional
 	Affinity *corev1.Affinity `json:"affinity,omitempty"`
 
+	// ShardAntiAffinity controls how pods from the same shard are spread across
+	// topology domains. When omitted, the operator does not add shard-specific
+	// anti-affinity rules beyond any user-supplied affinity.
+	// +optional
+	ShardAntiAffinity *ShardAntiAffinitySpec `json:"shardAntiAffinity,omitempty"`
```
jdheyburn, line 75: I appreciate you raising this PR for discussion! Before we add any new attributes to CRDs, having a [Design] issue raised would be beneficial so that we can understand all use cases and solutions. This is so that we have a CRD that is well designed, because its hard to make changes to a CRD after the fact.

DharmendraChoudhary67:
Opened a design discussion in #146 so we can converge on the placement API before pushing the implementation further.

That issue captures the main questions raised here:
- native `topologySpreadConstraints` vs operator-specific abstraction
- soft/preferred vs strict/required behaviour
- shard-aware selector injection
- how to surface unschedulable placement failures clearly
- room for future hostname/zone/region-aware placement

Happy to continue the implementation work once we have consensus there.

DharmendraChoudhary67:
Closing this in favor of #163.

The design discussion in #146 converged on a different upstream direction than the original default anti-affinity approach here:
- native `topologySpreadConstraints`
- shard-aware selector injection
- no forced default placement policy
- explicit `PodUnschedulable` status surfacing when strict placement cannot be satisfied

I moved the implementation to a new PR so the review scope matches the final design and the history stays easier to follow.

Issue #121, opened 2026-03-23
[Design] Persistent Volumes for ValkeyNode


jdheyburn:
Previous discussions:

- https://github.com/valkey-io/valkey-operator/discussions/88
- https://github.com/valkey-io/valkey-operator/pull/85

The user should be able to allow for volumes to be attached to ValkeyNodes. This can be used for:

* persisting memory to disk
  * allow for a partial resync on pod roll
* persisting nodes.conf (when in Cluster mode) 
  * to avoid having to rediscover the cluster on pod rolls
* other data durability needs

When a volume is enabled on the ValkeyCluster, it should traverse this config to the ValkeyNode, which should create the volume prior to creating the workload (Deployment/StatefulSet). Then attach the volume to the workload (Deployment/StatefulSet) so that it can be read.

The volume should be independently managed by the ValkeyNode controller so that we can dynamically resize it (should the storage class support it) in either:

- a user request
- automatically increasing via a future feature

Note: cluster node reconciliation should continue to happen regardless of whether the volume is created or not. 

## Proposed specs:

```yaml
apiVersion: valkey.io/v1alpha1
kind: ValkeyCluster
metadata:
  name: my-cluster
spec:
  shards: 3
  replicas: 1
  persistence:
    size: 10Gi
    storageClassName: gp3
```

This then is traversed to ValkeyNodes:

```yaml
kind: ValkeyNode
metadata:
  name: my-cluster-0-0
spec:
  # ... other attributes
  persistence:
    size: 10Gi
    storageClassName: gp3
```

We can use presence-based enabling (if persistence is defined, create the PVC).

## Go structs and usage

```go
type PersistenceSpec struct {
    // Size of the PVC. Required when mode is enabled.
    // +optional
    Size *resource.Quantity `json:"size,omitempty"`

    // StorageClassName for the PVC. Uses cluster default if omitted.
    // +optional
    StorageClassName *string `json:"storageClassName,omitempty"`
}

type ValkeyNodeSpec struct {
    // ... existing fields ...

    // Persistent storage configuration.
    // When nil, no PVC is created (ephemeral storage).
    // +optional
    Persistence *PersistenceSpec `json:"persistence,omitempty"`
}

type ValkeyClusterSpec struct {
    // ... existing fields ...

    // Persistent storage configuration passed through to all ValkeyNodes.
    // +optional
    Persistence *PersistenceSpec `json:"persistence,omitempty"`
}
```

## Consumption

[...]

PR #85, opened 2026-02-16
feat: Add persistent volume support for ValkeyCluster

umar-riswan-ap:
- Add StorageSpec to ValkeyCluster API with enabled, size, storageClassName, and accessModes fields
- Add VolumePermissions field to enable init container for volume ownership
- Implement StatefulSet support for persistent storage with PVC templates
- Keep Deployment support for ephemeral storage (emptyDir)
- Update valkey.conf with dir /data and AOF persistence settings
- Add security contexts (non-root user 1001, drop ALL capabilities)
- Add init container to chown /data directory when volumePermissions=true
- Update RBAC with StatefulSet and PVC permissions
- Add two sample configurations: persistent and ephemeral
- Update README with storage features and examples
- Tested on EKS with gp3 storage class

PR #143, opened 2026-04-20
Add PVC-backed storage for StatefulSet nodes

DharmendraChoudhary67:
## Summary
- add an optional storage API for durable Valkey data volumes
- wire storage into StatefulSet volume claim templates and mount Valkey data at /data
- extend config generation, samples, and tests to cover PVC-backed storage


daanvinken:
You may wanna check out https://github.com/valkey-io/valkey-operator/issues/121 and https://github.com/valkey-io/valkey-operator/pull/85. There are some ongoing conversations around this, including a dicussion to not use `VolumeClaimTemplates` due to re-sizinglimitations.

DharmendraChoudhary67:
Thanks for the feedback here. I reworked the persistence implementation around the direction in #121 instead of continuing with the earlier `volumeClaimTemplates` approach.

I opened a fresh PR with the redesigned implementation here: #149.

That new PR switches to `ValkeyNode`-managed PVCs, keeps persistence `StatefulSet`-only, adds explicit reclaim policy and PVC readiness / resize-aware status, and tightens the persistence mutation contract so it stays expansion-only and upstream-friendly.

I thought a fresh PR would be easier to review than rewriting this thread in place since the architecture is now materially different from the original `volumeClaimTemplates` design proposed here. Happy to continue discussion in #149.

jdheyburn:
@DharmendraChoudhary67 Are we able to close out this PR in favour of #149 ?

jdheyburn:
Doh I hit the wrong button, but feel free to reopen if need be.

DharmendraChoudhary67:
lol.. NP either way we had to close it.

PR #145, opened 2026-04-22
fix: pass cluster arg to getValkeyClusterState in proactive failover

daanvinken:
## Description

The merge of #128 (proactive failovers) after #133 (TLS) introduced a build-breaking call. #133 added a `cluster` parameter to `getValkeyClusterState`, but #128's new call site at line 445 used the old signature without it. This causes a compile error on current main.

## Testing

`go build ./...` passes.


daanvinken:
Should we start using merge queues? GitHub has this built in ("Require merge queue" branch protection). A merge queue rebases each PR onto the latest main and runs CI *after* the rebase, before actually merging. That would have caught the compile error when #128 was rebased onto post-#133 main.

jdheyburn (approved):

jdheyburn:
@daanvinken This is worth bringing up at the weekly call - I won't be able to make this weeks one though. I'll put a message out on Slack on the day.

PR #149, opened 2026-04-22
Add ValkeyNode-managed persistence support

DharmendraChoudhary67:
## Summary
- add first-class persistence configuration to `ValkeyCluster` and `ValkeyNode`
- manage PVCs directly from `ValkeyNode` instead of using `volumeClaimTemplates`
- mount the managed PVC into StatefulSet workloads and persist `nodes.conf`
- add explicit `Retain` / `Delete` reclaim policy, PVC readiness and resize-aware status, and expansion-only persistence mutation rules

## Why
This follows the direction discussed in #121 and the earlier feedback on #85 / #143:
- persistence is modeled on `ValkeyCluster` and propagated to `ValkeyNode`
- `ValkeyNode` owns the PVC lifecycle
- the implementation avoids `StatefulSet.volumeClaimTemplates` as the storage control plane
- the contract is kept narrow and upstream-friendly: persistence is `StatefulSet`-only, cannot be added later, cannot be removed once set, and only supports size expansion

## What changed
- add shared persistence API types and CRD schema
- create/update named PVCs in the `ValkeyNode` reconciler
- wire workloads and generated config to use `/data`
- surface PVC readiness / resize state on `ValkeyNode` status
- support explicit reclaim policy with delete finalizer behavior
- add unit/envtest coverage and end-to-end persistence coverage

## Follow-up
This supersedes the earlier `volumeClaimTemplates`-based approach in #143.


jdheyburn:
Thanks for raising! This is definitely something we want, hope to get round to reviewing it this week. In the meantime would you mind rebasing against main?

DharmendraChoudhary67:
done.

internal/controller/valkeynode_controller.go
```diff
@@ -26,6 +26,7 @@ import (
 	vclient "github.com/valkey-io/valkey-go"
```
daanvinken, line 26: This file is getting big, would it make sense to split into valkeynode_persistence.go?

internal/controller/utils.go
```diff
@@ -269,3 +272,39 @@ func getTLSConfig(ctx context.Context, c client.Client, secretName, serverName,
 	}
 	return tlsCfg, nil
 }
+
+func generateValkeyNodeConfig(node *valkeyv1.ValkeyNode) string {
```
daanvinken: Just for my understanding, how does this relate/overlap with [`GetBaseConfig()`](https://github.com/valkey-io/valkey-operator/blob/bc78e5cc6b1339d8bebb7633943a070f455afc35/internal/controller/config.go#L58)?

For example TLS config here is a hardcodes string, but a map in the existing one. Should we/can we consolidate?

daanvinken (commented):
This looks great, thanks for plumbing through. Excited to test it out.

I realized the controller doesn't watch PVCs in `SetupWithManager` (it already watches ConfigMaps, StatefulSets, Deployments via `Owns()`). This means PVC state transitions like Pending -> Bound won't trigger a reconcile until the next 60s periodic requeue, which could make initial creation slower with persistence enabled.

Since we intentionally don't set ownerReferences on PVCs (discussed in #121), we can't use `Owns()`. But we could use `Watches()` with a custom handler that maps PVCs back to ValkeyNodes by label.

I don't think it's worth it to go down this path right now, but felt worth it to mention.

daanvinken (approved):

internal/controller/valkeynode_controller.go
DharmendraChoudhary67: Yepp. I’ll split the persistence-specific reconcile helpers into a separate `valkeynode_persistence.go` so the main controller flow stays easier to read.

internal/controller/utils.go
DharmendraChoudhary67: I agree there’s overlap, especially around TLS and persistence directives. Rather than fully merging the cluster and standalone-node builders, I’m thinking of extracting a shared helper for the common managed directives (ACL, persistence paths, TLS), then keeping buildServerConfig() responsible for cluster-specific defaults and user-config merge behavior while generateValkeyNodeConfig() reuses the same shared helper for the standalone ValkeyNode path. That should reduce drift without forcing both paths into the same shape

daanvinken:
FWIW, tested this locally on a kind cluster as well. Created a 2-shard 1-replica cluster with persistence:

```yaml
spec:
  shards: 2
  replicas: 1
  persistence:
    size: 1Gi
```

All 4 PVCs came up bound. Wrote a key (`valkey-cli -c SET testkey "persist-me"`), ran `BGSAVE`, confirmed `dump.rdb` and `nodes.conf` both on `/data/`. Deleted the pod, it came back with the same node ID and the key was still there.

Tested both reclaim policies. Default Retain: deleted the cluster, PVCs stuck around. Then created a fresh cluster with `reclaimPolicy: Delete`, verified the finalizer was on the ValkeyNode, deleted the cluster, PVCs were cleaned up.

daanvinken (approved):
LGTM but my approval has no functional meaning :)

Perhaps @jdheyburn wants to have a look as well and possibly involve others. I think the approach here looks great.

jdheyburn:
Thanks for raising this @DharmendraChoudhary67 and also to @daanvinken for reviewing - very helpful in what has been a busy week my side. I'll try to get round to testing this by EOD Monday.

internal/controller/valkeynode_controller.go
```diff
@@ -93,6 +105,9 @@ func (r *ValkeyNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 }
 
 func (r *ValkeyNodeReconciler) ensureWorkload(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) error {
+	if node.Spec.Persistence != nil && node.Spec.WorkloadType == valkeyiov1alpha1.WorkloadTypeDeployment {
+		return fmt.Errorf("persistence requires workloadType StatefulSet")
+	}
```
jdheyburn: Is this check necessary if the CRD validates it for us? Or is this an additional guard just to be safe?

internal/controller/valkeynode_persistence.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	return metav1.ConditionTrue,
+		valkeyiov1alpha1.ValkeyNodeReasonPersistentVolumeClaimBound,
+		fmt.Sprintf("PersistentVolumeClaim %s is bound", pvc.Name)
+}
+
+func pvcSizeStatusCondition(node *valkeyiov1alpha1.ValkeyNode, pvc *corev1.PersistentVolumeClaim) (metav1.ConditionStatus, string, string) {
+	if node.Spec.Persistence == nil {
+		return metav1.ConditionTrue, "", ""
+	}
```
jdheyburn: `pvcSizeStatusCondition` is called by one parent that already wraps the function call in this condition. Is it worth keeping this check here?

jdheyburn (commented):
I did a local test; the node reloaded its existing config and also performed a partial resync!

On the whole the PR looks great, and I'm looking forward to getting this tested in a live cloud environment to see how it behaves, but that will be after we've merged. 

I've a couple of low priority comments. On top of them, could you please update `docs/status-conditions.md` with the new status conditions being used?

DharmendraChoudhary67:
Good catch! @jdheyburn , Its safe to remove them as CRD validation already handles these cases. I'll update the `docs/status-conditions.md`.

ekarlso:
@jdheyburn chance to get this in ?

jdheyburn (approved):
Thank you for your help on this! 🎉

yuvalcohenrise:
Hi @jdheyburn @DharmendraChoudhary67,
My team and I built our own custom operator prior to this project rolling out, and we recently resolved a critical edge-case issue in production regarding pod rescheduling and gossip communication that I wanted to bring up here.

In Kubernetes environments with fast CNI IP recycling and frequent pod movement (for us, aggressive node consolidation to save compute costs), we encountered a scenario where two completely separate Valkey clusters merged into a single split-brain cluster:

1. A pod in Cluster A is terminated/rescheduled and releases its IP.
2. The CNI reassigns that exact IP to a newly scheduled pod in Cluster B.
3. Healthy pods in Cluster A still have the old IP in their gossip peer list and continue attempting heartbeat pings over the Cluster Bus port (16379).
4. The new Cluster B pod accepts the packet, triggers a gossip handshake, and the two clusters swap topologies and accidentally merge.

The Operator logic alone doesn't prevent this, even with persistence disabled or operator reconciliation active, there is a distinct race condition. The CNI re-allocates the IP faster than the operator reconciliation loop can remove the dead node.

Because Valkey gossip relies on raw IP addresses under the hood, the only foolproof fix we found was applying a fine-grained NetworkPolicy to strictly restrict traffic on port 16379 so pods only accept gossip traffic from sibling pods within their specific cluster instance (while leaving port 6379 open for application clients).

We are currently in the process of migrating from our in-house operator to this official project (great work on this repo, by the way!). Is this edge case something you've considered?

If relevant, I would be happy to contribute a PR adding built-in NetworkPolicy generation (or optional toggles) to isolate cluster bus traffic per instance!

jdheyburn:
Hey @yuvalcohenrise, thanks for raising. Would you mind raising a separate issue for this?

yuvalcohenrise:
@jdheyburn done - opened #356 to track this separately.

PR #150, opened 2026-04-23
Update Go build

SuperQ:
### Summary

Update building to Go 1.26.2.
* Swap using the `go.mod` for a `.go-version` file for CI versions since it doesn't support support properly using the latest release.
* Fix Go minimum version spec to 1.25.0 to follow Go conventions.
* Update Dockerfile to 1.26.x.
* Update golangci-lint to support Go 1.26.x.
* Update pre-commit toolchain to support testing with latest.
* Update devcontainer to 1.26.x.

### Checklist


.github/workflows/lint.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c  # v6.4.0
         with:
-          go-version-file: go.mod
+          go-version-file: .go-version
```
sandeepkunusoth, line 37: is using .go-version better than what was before?

sandeepkunusoth (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module valkey.io/valkey-operator
 
-go 1.25.3
+go 1.25.0
```
sandeepkunusoth, line 4: any reason for not using 1.26 here and in pre-commit-config.yaml

.github/workflows/lint.yml
SuperQ: Yes, the way the `setup-go` action works is that it reads the _exact_ version from go.mod, rather than [the way the go directive is designed](https://go.dev/ref/mod#go-mod-file-go).

Using the .go-version lets us keep the latest version we want for builds. Unforunately from what I can tell it is also an exact version spec and doesn't allow for something like 1.26.x to automatically specify the latest patch release.

go.mod
SuperQ: Yes, per the design of the `go` directive in Go modules, it is a minimum language spec for compile features. Currently Go 1.25.x is a supported Go version, so we want to maintain compatibility with it until Go 1.27 is the latest and Go 1.26 becomes the oldest supported version.

jdheyburn (approved):

go.mod
bjosv: go 1.25.3 came from kubebuilder, so we should remember this during the next scaffolding update, i.e. not change it back.

bjosv (approved):

go.mod
SuperQ: Sounds like a bug in kubebuilder, not understanding how Go modules work.

PR #159, opened 2026-05-04
feat: add operator-specific Prometheus metrics

daanvinken:
**Description**

Adds custom Prometheus metrics for ValkeyCluster observability. The default controller-runtime metrics (reconcile counts, workqueue depth) don't expose operator-specific information.

New metrics:

- `valkey_operator_cluster_state_info` - gauge with state label (Initializing, Ready, Reconciling, Degraded, Failed), value 1 for current state, 0 for all others
- `valkey_operator_cluster_shards` - total shard count per cluster
- `valkey_operator_cluster_shards_ready` - ready shard count per cluster
- `valkey_operator_failovers_total` - counter for proactive failover events
- `valkey_operator_slot_migration_batches_total` - counter for completed slot migration batches (scale-out and scale-in)

Gauge metrics are cleaned up when a ValkeyCluster is deleted. All metrics (gauges and counters) are cleaned up when a ValkeyCluster is deleted to prevent cardinality leaks.

Uses `valkey_cluster` label instead of `cluster` to avoid conflicts in multi-cluster Prometheus setups ([context](https://github.com/cloudnative-pg/cloudnative-pg/issues/2501)). Uses `ClusterState` constants from API types to avoid string mismatches.

**Testing**

Unit tests pass. Deployed to a local kind cluster with a 3-shard 1-replica cluster and scraped the metrics endpoint:

```
$ curl -s http://localhost:8443/metrics | grep valkey_operator_cluster

# HELP valkey_operator_cluster_state_info Information about a ValkeyCluster. Value is 1 for the current state, 0 for all others.
# TYPE valkey_operator_cluster_state_info gauge
valkey_operator_cluster_state_info{valkey_cluster="failover-test",namespace="default",state="Degraded"} 0
valkey_operator_cluster_state_info{valkey_cluster="failover-test",namespace="default",state="Failed"} 0
valkey_operator_cluster_state_info{valkey_cluster="failover-test",namespace="default",state="Initializing"} 0
valkey_operator_cluster_state_info{valkey_cluster="failover-test",namespace="default",state="Ready"} 1
valkey_operator_cluster_state_info{valkey_cluster="failover-test",namespace="default",state="Reconciling"} 0
# HELP valkey_operator_cluster_shards Total number of shards in a ValkeyCluster.
# TYPE valkey_operator_cluster_shards gauge
valkey_operator_cluster_shards{valkey_cluster="failover-test",namespace="default"} 3
# HELP valkey_operator_cluster_shards_ready Number of ready shards in a ValkeyCluster.
# TYPE valkey_operator_cluster_shards_ready gauge
valkey_operator_cluster_shards_ready{valkey_cluster="failover-test",namespace="default"} 3
```

### DIFF
--- a/internal/controller/metrics.go
+++ b/internal/controller/metrics.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2025 Valkey Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
+)
+
+var (
+	clusterInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "valkey_operator_cluster_info",
+			Help: "Information about a ValkeyCluster. Value is always 1, state is in the label.",
+		},
+		[]string{"cluster", "namespace", "state"},
+	)
+
+	clusterShards = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "valkey_operator_cluster_shards",
+			Help: "Total number of shards in a ValkeyCluster.",
+		},
+		[]string{"cluster", "namespace"},
+	)
+
+	clusterShardsReady = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "valkey_operator_cluster_shards_ready",
+			Help: "Number of ready shards in a ValkeyCluster.",
+		},
+		[]string{"cluster", "namespace"},
+	)
+
+	failoversTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "valkey_operator_failovers_total",
+			Help: "Total number of failover events.",
+		},
+		[]string{"cluster", "namespace", "type"},
+	)
+
+	slotMigrationsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "valkey_operator_slot_migrations_total",
+			Help: "Total number of slot migration batches completed.",
+		},
+		[]string{"cluster", "namespace"},
+	)
+)
+
+func init() {
+	metrics.Registry.MustRegister(
+		clusterInfo,
+		clusterShards,
+		clusterShardsReady,
+		failoversTotal,
+		slotMigrationsTotal,
+	)
+}
+
+// clusterStates lists all possible ValkeyCluster states for metric cleanup.
+var clusterStates = []string{"Ready", "Reconciling", "Degraded", "Failed"}
+
+// updateClusterMetrics sets the Prometheus gauges for a ValkeyCluster.
+func updateClusterMetrics(cluster *valkeyiov1alpha1.ValkeyCluster) {
+	name := cluster.Name
+	ns := cluster.Namespace
+
+	// Set info gauge: 1 for current state, 0 for all others
+	for _, s := range clusterStates {
+		val := float64(0)
+		if string(cluster.Status.State) == s {
+			val = 1
+		}
+		clusterInfo.WithLabelValues(name, ns, s).Set(val)
+	}
+
+	clusterShards.WithLabelValues(name, ns).Set(float64(cluster.Status.Shards))
+	clusterShardsReady.WithLabelValues(name, ns).Set(float64(cluster.Status.ReadyShards))
+}
--- a/internal/controller/metrics.go
+++ b/internal/controller/metrics.go
@@ -32,7 +32,7 @@ var (
 			Name: "valkey_operator_cluster_state_info",
 			Help: "Information about a ValkeyCluster. Value is 1 for the current state, 0 for all others.",
 		},
-		[]string{"valkey_cluster", "namespace", "state"},
+		[]string{"valkey_cluster", "target_namespace", "state"},
 	)
 
 	clusterShards = factory.NewGaugeVec(
@@ -40,7 +40,7 @@ var (
 			Name: "valkey_operator_cluster_shards",
 			Help: "Total number of shards in a ValkeyCluster.",
 		},
-		[]string{"valkey_cluster", "namespace"},
+		[]string{"valkey_cluster", "target_namespace"},
 	)
 
 	clusterShardsReady = factory.NewGaugeVec(
@@ -48,7 +48,7 @@ var (
 			Name: "valkey_operator_cluster_shards_ready",
 			Help: "Number of ready shards in a ValkeyCluster.",
 		},
-		[]string{"valkey_cluster", "namespace"},
+		[]string{"valkey_cluster", "target_namespace"},
 	)
 
 	failoversTotal = factory.NewCounterVec(
@@ -56,7 +56,7 @@ var (
 			Name: "valkey_operator_failovers_total",
 			Help: "Total number of failover events.",
 		},
-		[]string{"valkey_cluster", "namespace", "type"},
+		[]string{"valkey_cluster", "target_namespace", "type"},
 	)
 
 	slotMigrationBatchesTotal = factory.NewCounterVec(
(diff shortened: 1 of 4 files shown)


Preisschild:
It might make sense to avoid using `cluster` (instead something like `valkey_cluster`) as label, since the `cluster` label is often used (and hardcoded in various open source dashboards) to refer to a kubernetes cluster, which might lead to conflicts in case the valkey-operator runs in a multi-k8s-cluster - centralized prometheus setup.


There is a similar issue in CloudNativePG: 
https://github.com/cloudnative-pg/cloudnative-pg/issues/2501

internal/controller/metrics.go
```diff
@@ -0,0 +1,97 @@
...
+		clusterInfo,
+		clusterShards,
+		clusterShardsReady,
+		failoversTotal,
+		slotMigrationsTotal,
+	)
+}
+
+// clusterStates lists all possible ValkeyCluster states for metric cleanup.
+var clusterStates = []string{"Ready", "Reconciling", "Degraded", "Failed"}
```
bjosv: As a detail, we have these defined in `api/v1alpha1/valkeycluster_types.go`,
so we could have a `[]valkeyiov1alpha1.ClusterState` and then use that for checking to avoid a string mismatch.
Then we do `clusterInfo.WithLabelValues(name, ns, string(s)).Set(val)`  instead.

We will still have to make sure we don't forget to add a new state to the list though..

daanvinken:
Ah yes good call, we are relabelling for that exact use case as well. Returning from PTO 05/26 - will have a look

internal/controller/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+		prometheus.CounterOpts{
+			Name: "valkey_operator_failovers_total",
+			Help: "Total number of failover events.",
+		},
+		[]string{"valkey_cluster", "namespace", "type"},
+	)
+
+	slotMigrationsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "valkey_operator_slot_migrations_total",
```
bjosv: This is counting the number of slot migration calls, with currently max rebalanceSlotBatchSize = 400 slots in each batch.
Maybe the name should be `valkey_operator_slot_migration_batches_total` then. 

A `valkey_operator_slot_migrations_total`  should probably count the number of migrated slots  instead (increase with the size of move.Slots).

internal/controller/valkeycluster_controller.go
```diff
@@ -1028,6 +1018,7 @@ func (r *ValkeyClusterReconciler) rebalanceSlots(ctx context.Context, cluster *v
 		}
 		return false, err
 	}
+	slotMigrationsTotal.WithLabelValues(cluster.Name, cluster.Namespace).Inc()
```
bjosv: This is only counting the number of migrated batches on a scale-out, we also have the scale-in scenario in `drainExcessShards`.

internal/controller/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	clusterShardsReady.WithLabelValues(name, ns).Set(float64(cluster.Status.ReadyShards))
+}
+
+// deleteClusterMetrics removes all gauge metrics for a deleted ValkeyCluster.
+func deleteClusterMetrics(name, namespace string) {
+	for _, s := range clusterStates {
+		clusterInfo.DeleteLabelValues(name, namespace, string(s))
+	}
+	clusterShards.DeleteLabelValues(name, namespace)
+	clusterShardsReady.DeleteLabelValues(name, namespace)
```
bjosv, line 95: failoverTotal and slotMigrationsTotal not delete here, but maybe we shouldn't?

internal/controller/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
+)
+
+var (
+	clusterInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "valkey_operator_cluster_info",
```
bjosv: Alternative:
```suggestion
			Name: "valkey_operator_cluster_state_info",
```

bjosv (commented):
Thanks for putting time on this!
Added some comments, the main thing is to include scale-in migrations in `slotMigrationsTotal`, unless I missed a reason for not including that.

internal/controller/metrics.go
daanvinken: Yeap that was my understanding

daanvinken:
Thanks @bjosv ! All should be addressed

bjosv (dismissed):
LGTM!

internal/controller/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+var (
+	clusterInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "valkey_operator_cluster_state_info",
+			Help: "Information about a ValkeyCluster. Value is 1 for the current state, 0 for all others.",
+		},
+		[]string{"valkey_cluster", "namespace", "state"},
+	)
+
+	clusterShards = prometheus.NewGaugeVec(
```
SuperQ: You can eliminate the init() function with `promauto`.

```suggestion
	"github.com/prometheus/client_golang/prometheus"
	"github.com/prometheus/client_golang/prometheus/promauto"
	"sigs.k8s.io/controller-runtime/pkg/metrics"

	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
)

var (
	clusterInfo = promauto.NewGaugeVec(
		prometheus.GaugeOpts{
			Name: "valkey_operator_cluster_state_info",
			Help: "Information about a ValkeyCluster. Value is 1 for the current state, 0 for all others.",
		},
		[]string{"valkey_cluster", "namespace", "state"},
	)

	clusterShards = promauto.NewGaugeVec(
```

internal/controller/metrics.go
```diff
@@ -0,0 +1,112 @@
...
+		clusterInfo,
+		clusterShards,
+		clusterShardsReady,
+		failoversTotal,
+		slotMigrationBatchesTotal,
+	)
+}
+
+// clusterStates lists all possible ValkeyCluster states for metric cleanup.
+var clusterStates = []valkeyiov1alpha1.ClusterState{
```
SuperQ: I wonder if we should define this list in `valkey.io/valkey-operator/api/v1alpha1` so we don't have to keep it in sync here.

internal/controller/metrics.go
SuperQ: `failoversTotal` has a `valkey_cluster` label, which means we probably want to also delete the label combination here. Otherwise we will leak cardinality. Same with `slotMigrationBatchesTotal`.

SuperQ (commented):
One thing I recommend is to init the label values early in `Reconcile()` call

Something like this:
 ```go
func (r *ValkeyClusterReconciler) setupClusterMetrics(cluster *valkeyiov1alpha1.ValkeyCluster) {
	failoversTotal.WithLabelValues(cluster.Name, cluster.Namespace)
	slotMigrationBatchesTotal.WithLabelValues(cluster.Name, cluster.Namespace)
}
```

It's OK to call this function multiple times as it's thread safe and pretty cheap.

I don't actually remember if the controller reconcile object sticks around after each loop, otherwise we could cache the `WithLabelValues()` on `ValkeyClusterReconciler` struct with [MustCurryWith](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec.MustCurryWith).

File: internal/controller/metrics.go (hunk 6 above)
Comment (bjosv) [reply internal/controller/metrics.go]: Ah, that's nice

**Comment by SuperQ** [comment]
@bjosv If you want to merge this as-is, I can followup with my suggested improvements.

**Comment by daanvinken** [comment]
Thanks folks, should be addressed.

File: internal/controller/metrics.go [hunk 8]
```diff
@@ -0,0 +1,98 @@
...
+// deleteClusterMetrics removes all metrics for a deleted ValkeyCluster.
+func deleteClusterMetrics(name, namespace string) {
+	for _, s := range valkeyiov1alpha1.ClusterStates {
+		clusterStateInfo.DeleteLabelValues(name, namespace, string(s))
+	}
+	clusterShards.DeleteLabelValues(name, namespace)
+	clusterShardsReady.DeleteLabelValues(name, namespace)
+	failoversTotal.DeletePartialMatch(prometheus.Labels{"valkey_cluster": name, "namespace": namespace})
+	slotMigrationBatchesTotal.DeleteLabelValues(name, namespace)
+}
```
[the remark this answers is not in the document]
daanvinken: Updated.

internal/controller/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+
+var factory = promauto.With(metrics.Registry)
+
+var (
+	clusterStateInfo = factory.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "valkey_operator_cluster_state_info",
+			Help: "Information about a ValkeyCluster. Value is 1 for the current state, 0 for all others.",
+		},
+		[]string{"valkey_cluster", "namespace", "state"},
```
SuperQ: Similar to what @Preisschild said about `cluster`, `namespace` is used for the namespace the controller is running in.

I would probably call this `target_namespace` or `valkey_cluster_namespace`.

daanvinken: Yeah good call

bjosv (approved):
A good start, wdyt @SuperQ ?

SuperQ (approved):
Yup, let's give it a go.

PR #162, opened 2026-05-05
docs: Add valkeynode-design.md

jdheyburn:
<!--
Thanks for contributing to Valkey Operator!

Please make sure you are aware of our contributing guidelines [available
here](https://github.com/valkey-io/valkey-operator/blob/main/CONTRIBUTING.md)

-->

This PR closes <Issue #>

### Summary

Adds a document outlining the decision to implement ValkeyNode, with workloadTypes of StatefulSet and Deployment.

<!--
Add a summary describing the changes
-->

<!--
Outline the feature support or behaviour changes included in this PR
-->

### Implementation

<!--
Describe the implementation details. Highlight key code changes and call out any areas where you want reviewers to pay extra attention
-->

### Limitations

<!--
Describe any features or use cases that are not implemented or are only partially supported
-->

### Testing

<!--
Describe what tests have been conducted and any relevant test results
-->

### Checklist

I'm open to suggestions on a format of these docs we can adopt. I can expand on this with extra headings such as below if folks deem it helpful - but I didn't want to have _too_ much filler.

```
Reconciliation behavior — the doc describes what ValkeyNode manages but not what the controller actually does each loop. A short description of the reconcile sequence (create workload →  
  wait for pod → probe Valkey → update status) would help readers understand the implementation without reading the code.                                                                    
                                                                                                                                                                                             
  Status/state lifecycle — the doc lists status fields but not the state machine. What states can a ValkeyNode be in? What transitions are valid? What causes it to become not-ready, and    
  does it self-recover? This is useful for both implementers and operators debugging a stuck node.                                                                                           
                                                                                                                                                                                             
  Health checking — ready is mentioned as "whether the pod is running and Valkey is responding" but not how that's determined. Is it a TCP probe, a PING command, INFO replication? The      
  mechanism matters because it affects how quickly role changes are reflected in status.
                                                                                                                                                                                             
  Config application — the doc says ValkeyNode applies config "as-is without interpretation" but doesn't say how. Is it a ConfigMap mounted as a file, command-line flags, or something else?
   That detail is load-bearing for anyone extending the system.
                                                                                                                                                                                             
  RBAC / permissions — what does the ValkeyNode controller need access to? This is often left out of design docs and then surprises people at deploy time.                                   
   
```

### DIFF
--- a/docs/valkeynode-design.md
+++ b/docs/valkeynode-design.md
@@ -1,0 +1,105 @@
+# ValkeyNode Design
+
+ValkeyNode is an internal CRD that abstracts the infrastructure used to create a Valkey pod. ValkeyNodes are not intended to be created by users — they are managed exclusively by parent controllers such as ValkeyCluster.
+
+## Goals
+
+- Separation of concerns: ValkeyNode handles infrastructure; parent controllers handle topology
+- Configuration-driven behaviour: ValkeyNode behaviour is determined by spec fields, not coupled to any specific parent CRD
+- Reusable across future Valkey deployment modes (standalone, replication, etc.)
+
+## Non-Goals
+
+- Primary election or promotion decisions
+- Failover orchestration
+- Cluster slot management
+- Multi-pod workloads (ValkeyNode always manages a single pod)
+
+## Workload Type
+
+Users can define what workload type to use — either a StatefulSet or Deployment. The reason for this is that Valkey means different things to different users. It can be used as a cache or as a source-of-truth data store. If you're using Valkey as a cache and don't care about data durability, a Deployment is preferable. If you need stable storage and identity, a StatefulSet is more appropriate.
+
+Supporting both workload types is why the distinction is abstracted behind a ValkeyNode CRD.
+
+## Singletons
+
+ValkeyNodes map 1:1 to pods. A ValkeyNode deploys its pod as either a StatefulSet or Deployment, but these workloads always have exactly one replica. This enables precise pod scheduling — for example, placing a primary pod in a specific availability zone while avoiding nodes that already host other primaries.
+
+If an entire Valkey Cluster were deployed into a single StatefulSet, all pods in that StatefulSet would share the same spec, making advanced per-pod scheduling impossible.
+
+See also: [ClickHouse singleton StatefulSet pattern](https://youtu.be/QLLKzOEMkUk?si=kLvKw8AMuvgGCwbT)
+
+## Observability
+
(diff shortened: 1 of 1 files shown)


docs/valkeynode-design.md
```diff
@@ -0,0 +1,105 @@
...
+
+- `workloadType` — Deployment or StatefulSet
+- `podTemplate` — image, resources, scheduling constraints
+- `persistence` — storage mode and size
+- `config` — all Valkey configuration values (e.g. `cluster-enabled`, `maxmemory`); ValkeyNode applies these as-is without interpretation
+
+**Parents read from status:**
+
+- `ready` — whether the pod is running and Valkey is responding
+- `observedRole` / `observedReplicaOf` — replication topology observed from `INFO replication`
```
bjosv: There is no `observered...` in [status](https://github.com/valkey-io/valkey-operator/blob/main/api/v1alpha1/valkeynode_types.go#L100) yet, should we remove this line until we have added them?

docs/valkeynode-design.md
```diff
@@ -0,0 +1,105 @@
...
+That said, ValkeyNode's abstraction does not preclude direct pod management. A `Direct` WorkloadType could be added in the future to bypass StatefulSet and Deployment entirely — the design accommodates this without structural changes.
+
+## Interaction Contract with Parent Controllers
+
+ValkeyNode is deliberately "dumb" — it reconciles infrastructure without making topology decisions. All cluster-level awareness lives in the parent controller.
+
+**Parents write to spec:**
+
+- `workloadType` — Deployment or StatefulSet
+- `podTemplate` — image, resources, scheduling constraints
```
bjosv: There own fields for image, resource..., i.e. there is no podTemplate in the [spec](https://github.com/valkey-io/valkey-operator/blob/bbbb25e5b55669807f7dd9a4b99470ca32d02dc3/api/v1alpha1/valkeynode_types.go#L41)
Should we add them instead?

..and `config` below is soon to be added, so thats ok.

bjosv (commented):
Very nice, short and concise.
I think we can wait with adding the suggested section until we have settled the design.

docs/valkeynode-design.md
jdheyburn: `role` is captured, so I've updated that

docs/valkeynode-design.md
jdheyburn: Thanks, I've made a sub-heading which will group these together, indicating the sit at the root level

bjosv (approved):
LGTM!

Issue #146, opened 2026-04-22
[Design] Shard-aware placement policy for primaries and replicas


DharmendraChoudhary67:
Previous discussions:

- #74
- #142

We have a valid scheduling problem to solve: in Cluster mode, a primary and its replica for the same shard should ideally not land on the same Kubernetes node. At the same time, we do not want to back ourselves into an overly narrow API or a surprising default behaviour.

This issue is to discuss the scheduling/placement design before we continue implementation work.

## Problem

Today the operator exposes standard Kubernetes scheduling knobs such as `affinity`, `tolerations`, and `nodeSelector`, but it does not yet have a settled design for shard-aware placement.

The requirements we want to reason about are broader than just "anti-affinity":

- avoid same-shard primary/replica single-node SPOF
- support soft and hard placement behaviour
- leave room for zone-aware / region-aware placement in the future
- avoid an API that users cannot reason about or override
- surface scheduling failures clearly enough to debug

## Questions to resolve

1. Should we expose a native Kubernetes scheduling surface such as `topologySpreadConstraints` on the CRD, rather than a custom operator-specific abstraction?
2. If we use `topologySpreadConstraints`, should the operator inject shard-scoped selectors automatically (for example `valkey.io/cluster` + `valkey.io/shard-index`) so users do not need to handcraft selectors?
3. What should the default behaviour be when the user does not specify anything?
   - no default placement policy
   - soft/preferred spreading
   - strict/required separation
4. How should strict placement failures be surfaced?
   - pod events only
   - ValkeyNode / ValkeyCluster status condition
   - both
5. How should this interact with user-supplied `affinity` or future zone-aware policies?
6. Do we want the first design to solve only hostname-level separation, or should we intentionally design for hostname/zone/region/rack from day one?

### Option A: Native `topologySpreadConstraints`

Expose `topologySpreadConstraints` on the CRD and let the operator inject shard-aware selectors where appropriate.

Pros:
- reuses native Kubernetes API
- flexible enough for hostname and zone-aware placement
- avoids inventing a new scheduling DSL too early

Open questions:
- whether we should expose raw constraints directly, or only a curated subset
- how much automatic selector injection the operator should do
- how to document strict vs soft behaviour clearly

### Option B: Dedicated operator placement policy

Introduce an operator-level placement policy abstraction and translate that to native affinity/spread rules internally.

Pros:
- easier to document for common operator use cases
- we can encode shard-aware intent directly

Cons:
[...]

PR #163, opened 2026-05-06
feat: add shard-aware topology spread constraints

DharmendraChoudhary67:
## Summary
- expose native `topologySpreadConstraints` on `ValkeyCluster` and propagate them to `ValkeyNode`
- augment pod topology spread rules with shard-aware cluster/shard selectors so same-shard primary/replica placement can be spread across nodes
- surface real scheduler placement failures as `ValkeyCluster.status.reason=PodUnschedulable`
- add unit and e2e coverage, including a strict scheduler-failure scenario

## Context
- follows the native topology spread direction discussed in #146
- supersedes the earlier default anti-affinity approach from #142

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -389,6 +389,7 @@ func (r *ValkeyClusterReconciler) handlePodSchedulingIssues(ctx context.Context,
 	if issue.message != "" {
 		message = fmt.Sprintf("%s: %s", message, issue.message)
 	}
+	r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "PodUnschedulable", "SchedulePod", "%s", message)
 	setCondition(cluster, valkeyiov1alpha1.ConditionDegraded, valkeyiov1alpha1.ReasonPodUnschedulable, message, metav1.ConditionTrue)
 	setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonPodUnschedulable, message, metav1.ConditionFalse)
 	setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonReconciling, "Waiting for unschedulable pods to be scheduled", metav1.ConditionTrue)
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -381,6 +381,7 @@ func (r *ValkeyClusterReconciler) handlePodSchedulingIssues(ctx context.Context,
 	}
 	if issue == nil {
 		removeConditionIfReason(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionDegraded, valkeyiov1alpha1.ReasonPodUnschedulable)
+		removeConditionIfReason(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonPodUnschedulable)
 		return ctrl.Result{}, false, nil
 	}
 
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -160,11 +160,26 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		_ = r.updateStatus(ctx, cluster, nil)
 		return ctrl.Result{}, err
 	} else if requeue {
+		if result, handled, err := r.handlePodSchedulingIssues(ctx, cluster); err != nil {
+			setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonValkeyNodeError, err.Error(), metav1.ConditionFalse)
+			_ = r.updateStatus(ctx, cluster, nil)
+			return ctrl.Result{}, err
+		} else if handled {
+			return result, nil
+		}
 		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonUpdatingNodes, "Updating ValkeyNodes", metav1.ConditionFalse)
 		setCondition(cluster, valkeyiov1alpha1.ConditionProgressing, valkeyiov1alpha1.ReasonUpdatingNodes, "Updating ValkeyNodes", metav1.ConditionTrue)
 		_ = r.updateStatus(ctx, cluster, nil)
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
 	}
+	if result, handled, err := r.handlePodSchedulingIssues(ctx, cluster); err != nil {
+		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonValkeyNodeError, err.Error(), metav1.ConditionFalse)
+		_ = r.updateStatus(ctx, cluster, nil)
+		return ctrl.Result{}, err
+	} else if handled {
+		return result, nil
+	}
+
 	operatorPassword, err := fetchSystemUserPassword(ctx, operatorUser, r.Client, cluster.Name, cluster.Namespace)
 	if err != nil {
 		log.Error(err, "failed to retrieve system user password")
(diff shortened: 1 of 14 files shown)


jdheyburn:
Thanks for raising this @DharmendraChoudhary67!

So that I understand what the goal of this PR is, would you mind showing some examples of how it might be used, particularly with the augmentation it performs?

DharmendraChoudhary67:
The goal of this PR is to expose native Kubernetes topology spread on ValkeyCluster/ValkeyNode, while augmenting the constraints so they apply shard-aware placement automatically.

For example, a user can configure:

```yaml
spec:
  topologySpreadConstraints:
    - maxSkew: 1
      topologyKey: kubernetes.io/hostname
      whenUnsatisfiable: DoNotSchedule
```

When the operator builds each ValkeyNode workload, it augments that constraint to scope spreading to the current cluster and shard. Conceptually the resulting pod template looks like:

```yaml
topologySpreadConstraints:
  - maxSkew: 1
    topologyKey: kubernetes.io/hostname
    whenUnsatisfiable: DoNotSchedule
    labelSelector:
      matchLabels:
        valkey.io/cluster: <cluster-name>
    matchLabelKeys:
      - valkey.io/shard-index
```

That means:
- spreading is only evaluated against pods from the same ValkeyCluster
- the scheduler further scopes it to pods with the same shard label as the incoming pod
- in practice, a primary/replica pair for shard `0` is spread independently from shard `1`, shard `2`, etc.

So for a `3 shard / 1 replica` cluster, this allows users to express host/zone spreading with the native Kubernetes API, while the operator makes it shard-aware without requiring them to hand-write per-shard selectors.

The PR also surfaces the failure mode for strict placement. If a user chooses `DoNotSchedule` and the topology cannot satisfy the spread (for example only one eligible node is available for a shard that requires two domains), the scheduler leaves the pod pending and the operator reports `ValkeyCluster.status.reason=PodUnschedulable`.

jdheyburn:
@DharmendraChoudhary67 SGTM - I will get round to testing this on Monday. I think this is a good first start for adding TSCs. Any long-term design work we will discuss in #146.

yuvalcohenrise:
Hey all, first comment here, we've built an in-house Valkey operator and are evaluating moving to a community-backed one, so really glad to see this project moving forward.

@DharmendraChoudhary67 nice work on the shard-aware augmentation, but I'm not sure TSC is the right tool here.

The thing Valkey actually needs is "try not to put this shard's primary and replica on the same node." That's a colocation-avoidance problem, and `PodAntiAffinity` is built exactly for that. A `preferredDuringSchedulingIgnoredDuringExecution` rule scoped to the shard label gives you a soft but meaningful HA default without blocking scheduling when topology is limited.

TSC is really about balancing pod counts evenly across domains. With `maxSkew=1` it looks at the global pod distribution and tries to keep things even, but "even" doesn't mean "safe." In a 3-shard/1-replica cluster on 2 nodes, TSC could legally land a shard's primary and replica on the same node if that satisfies the skew math. That's a valid schedule but a bad day for your data.

The `matchLabelKeys` augmentation narrows this per-shard which helps, but now the operator is carrying logic that anti-affinity handles for free in a single `matchLabels` block.

I'd rather see the operator inject `preferredDuringSchedulingIgnoredDuringExecution` anti-affinity scoped to shard labels by default, users get sensible HA placement out of the box, scheduling still works on small/constrained clusters.

DharmendraChoudhary67:
@yuvalcohenrise  One reason we chose TSC here is that from a longer-term/upstream perspective, If we later want to support topology-aware scheduling across zones, racks, or other custom topology domains, TSC provides a more extensible foundation for shard-aware spreading semantics.

Also, the scheduler itself does not inherently understand leader/follower relationships between pods. Regardless of whether we use TSC or podAntiAffinity, we still need shard-aware labels/selectors so the scheduler can distinguish “same shard” vs “different shard” pods.

The intention behind the matchLabelKeys augmentation was to introduce that shard-awareness into the scheduling decision, so spread evaluation happens independently per shard rather than globally across all Valkey pods.

With the current configuration:

`whenUnsatisfiable: DoNotSchedule
maxSkew: 1
topologyKey: kubernetes.io/hostname
matchLabelKeys:
    - valkey.io/shard-index`

if the leader pod of a shard is already present on a node, scheduling the follower onto the same hostname domain would violate the skew constraint (2-0 > 1), so the pod remains unschedulable instead of being colocated on the same node.

I do agree that  ` podAntiAffinity ` expresses the HA intent more directly from a semantic/readability perspective, but the intention here was to build shard-aware topology semantics that can also scale naturally to future topology domains like zones/racks.

jdheyburn:
Sorry for the delay getting to this, it's not been forgotten. Hope to find some capacity to go through this and the design proposal soon.

DharmendraChoudhary67:
@jdheyburn   resolved conflicts. u can test it now

jdheyburn:
Sorry for the delay here @DharmendraChoudhary67, I've tried to get the branch updated with the latest changes too.

There are multiple changes happening here besides just adding TSCs, such as refactoring (adding `reconcileNodesAndScheduling` amongst others). That is making the PR difficult to understand exactly the changes you want to add here.

Can you help me by pointing me to the places where I should be focusing on?

DharmendraChoudhary67:
I cleaned up the branch so the PR is now based on the latest upstream main with a single feature commit, and removed the extra merge/refactor noise, including the `reconcileNodesAndScheduling` refactor.

The places to focus now are:
- API surface: `ValkeyCluster.spec.topologySpreadConstraints` and `ValkeyNode.spec.topologySpreadConstraints`
- Propagation: `buildClusterValkeyNode` copies the cluster constraints down to each ValkeyNode
- Scheduling behavior: `buildShardTopologySpreadConstraints` augments user-provided constraints with cluster/shard scoping
- Failure reporting: `PodUnschedulable` status handling surfaces strict scheduler failures on the ValkeyCluster
- Tests/sample: focused unit tests, topology-spread e2e coverage, and a sample CR

jdheyburn:
@DharmendraChoudhary67 Thank you, this is cleaner! Would you mind adding some docs to the below on this new feature?

- https://github.com/valkey-io/valkey-operator/blob/main/docs/valkeycluster.md#scheduling

I'm happy to get this in after that, then we can iterate on the design for the remaining scheduling.

DharmendraChoudhary67:
Added docs under the ValkeyCluster Scheduling section for shard-aware topology spread constraints. The docs now cover the default behavior when the field is omitted, the required Kubernetes fields, and the impact of `DoNotSchedule` vs `ScheduleAnyway`.

internal/controller/valkeycluster_controller.go
```diff
@@ -354,6 +369,66 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if issue == nil {
+		removeConditionIfReason(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionDegraded, valkeyiov1alpha1.ReasonPodUnschedulable)
+		removeConditionIfReason(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonPodUnschedulable)
+		return ctrl.Result{}, false, nil
+	}
+
+	message := fmt.Sprintf("Pod %s is unschedulable", issue.podName)
+	if issue.message != "" {
+		message = fmt.Sprintf("%s: %s", message, issue.message)
+	}
```
bjosv, line 391: Would it help users if we emit an event via `r.Recorder.Eventf` here as well?

DharmendraChoudhary67: @bjosv Sure, we can add an event here.

One thing I wanted to confirm: should this warning event be emitted on every reconcile while the pod remains `unschedulable`, or only when the cluster first transitions into `PodUnschedulable`?

We could also emit a normal `PodSchedulingResolved` event when scheduling recovers, will it be a over kill, let me know if it would be useful.

bjosv: I was thinking just in the transition, but when checking the other events we have no guards there, but they might not be called multiple times.
But, if we emit the same event in every reconcile it will just shown once anyway but having a count, so maybe that is simpler.
We also don't have any events when things are resolved, that would be a new pattern so I'm ok skipping `PodSchedulingResolved`.

jdheyburn (dismissed):

DharmendraChoudhary67:
Added the warning event for unschedulable pods only. It emits a `PodUnschedulable` warning event with the same message used for the status condition, and I skipped a resolved event to keep the existing event pattern unchanged.

: @jdheyburn sorry. . .have to approve again :-|

jdheyburn:
Closes #224

jdheyburn (approved):
LGTM, I'll defer to @bjosv for final approval and merge if his comment is resolved

bjosv (approved):
Looks great, thanks!

jdheyburn:
@DharmendraChoudhary67 Can you expand a bit on the refactoring that the PR was doing? Do you think its worth adding it in another PR?

DharmendraChoudhary67:
@jdheyburn No I don't think so. It was mostly a cleanup attempt to reduce the size/complexity of `Reconcile` by extracting the ValkeyNode reconciliation + scheduling check path into a helper. After cleaning up the branch, I removed it to keep this PR focused only on topology spread constraints.

jdheyburn:
@DharmendraChoudhary67 Sounds good -- we want to reduce the complexity of the Reconcile function at some point in #185, although I imagine we'll get it prioritised for maybe v1.1.

PR #164, opened 2026-05-06
[feat] Roll ValkeyNodes on changes to ValkeyCluster.Spec.Config

### LINKED ISSUE #50: TODO: Updates to valkeyConfig do not trigger rolling restart
This issue documents a TODO, where user updates to the valkeyConfig (configMap) do not trigger a rolling restart of the cluster.

jdheyburn:
This PR closes #50 

### Summary

Will make ValkeyNodes roll sequentially if ValkeyCluster.Spec.Config is updated.

This avoids the need for the user to manually roll pods for them to pick up config changes.

Nodes now pick up config changes made on ValkeyCluster.Spec.Config.

### Implementation

Added a `ServerConfigHash` field to ValkeyNode.Spec. When the valkey cluster controller updates this, it will cause a generation bump and trigger the underlying pod to be rolled.

Note that I fully intend on this being a temporary solution. When we refactor configs as part of #141 - this can be removed. Given we are in alpha we can continue to make these kind of changes.

### Limitations

N/A

### Testing

I ran a local test and saw the change trickle down.

As part of #141 we should introduce some e2e tests for configs, but we'll revisit that after its been implemented.

### Checklist

### DIFF
--- a/api/v1alpha1/valkeynode_types.go
+++ b/api/v1alpha1/valkeynode_types.go
@@ -79,6 +79,13 @@ type ValkeyNodeSpec struct {
 	// +optional
 	ServerConfigMapName string `json:"serverConfigMapName,omitempty"`
 
+	// ServerConfigHash is a hash of the server ConfigMap contents. When set by
+	// the ValkeyCluster controller it bumps the spec Generation, which triggers
+	// the ValkeyNode controller to reconcile and update the pod template
+	// annotation — causing a rolling restart when the cluster config changes.
+	// +optional
+	ServerConfigHash string `json:"serverConfigHash,omitempty"`
+
 	// UsersACLSecretName is the name of the Secret containing the ACL user
 	// file. When set, mounts a users-acl volume from this Secret so the
 	// container can load aclfile /config/users/users.acl.
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -428,6 +428,19 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNodes(ctx context.Context, clus
 	return false, nil
 }
 
+// getServerConfigHash returns the config hash annotation from the cluster's
+// ConfigMap. Propagating this to ValkeyNode annotations triggers the ValkeyNode
+// controller to reconcile and update the pod template, causing a rolling restart
+// when the cluster config changes.
+func (r *ValkeyClusterReconciler) getServerConfigHash(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) string {
+	cm := &corev1.ConfigMap{}
+	if err := r.Get(ctx, client.ObjectKey{Name: GetServerConfigMapName(cluster.Name), Namespace: cluster.Namespace}, cm); err != nil {
+		return ""
+	}
+	return cm.Annotations[configHashKey]
+}
+
+
 // reconcileValkeyNode reconciles a single ValkeyNode for (shardIndex, nodeIndex).
 // Returns (requeue, nodeCreated, err). requeue signals the caller should stop
 // iterating and wait before processing the next node.
@@ -465,6 +478,7 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNode(ctx context.Context, clust
 	result, err := controllerutil.CreateOrUpdate(ctx, r.Client, node, func() error {
 		node.Labels = desired.Labels
 		node.Spec = desired.Spec
+		node.Spec.ServerConfigHash = r.getServerConfigHash(ctx, cluster)
 		return controllerutil.SetControllerReference(cluster, node, r.Scheme)
 	})
 	if err != nil {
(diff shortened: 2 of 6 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -130,6 +130,13 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		_ = r.updateStatus(ctx, cluster, nil)
 		return ctrl.Result{}, err
 	}
+	configHash := func() string {
+		cm := &corev1.ConfigMap{}
+		if err := r.Get(ctx, client.ObjectKey{Name: GetServerConfigMapName(cluster.Name), Namespace: cluster.Namespace}, cm); err != nil {
```
bjosv, line 135: `upsertConfigMap` above could also return newServerConfigHash using `(string, error)` to avoid this `Get`, and then we cant get transient errors.
But if this is a temporary change and we don't want to change config.go I'm ok with the current impl.

bjosv (approved):

api/v1alpha1/valkeynode_types.go
```diff
@@ -79,6 +79,13 @@ type ValkeyNodeSpec struct {
 	// +optional
 	ServerConfigMapName string `json:"serverConfigMapName,omitempty"`
 
+	// ServerConfigHash is a hash of the server ConfigMap contents. When set by
+	// the ValkeyCluster controller it bumps the spec Generation, which triggers
+	// the ValkeyNode controller to reconcile and update the pod template
+	// annotation — causing a rolling restart when the cluster config changes.
+	// +optional
+	ServerConfigHash string `json:"serverConfigHash,omitempty"`
```
sandeepkunusoth, line 87: why do we need hash on actual spec. cant we store this in Valkey CR annotations as its auto generated from config map?

jdheyburn: Generation only updates on a Spec change, which is used to determine if the pods need to be rolled.

ValkeyCluster manages the config map, and if the valkeynode controller saw that configmaps had changed, then all pods would be rolled at once.

We want the roll to happen sequentially, so in this PR each node would get updated one at a one with the config hash.

In a future PR, when config is managed in ValkeyNode directly, that would cause a generation update and cause pods to roll sequentially - then we can remove ServerConfigHash.

internal/controller/valkeynode_controller.go
```diff
@@ -115,6 +115,19 @@ func (r *ValkeyNodeReconciler) ensureWorkload(ctx context.Context, node *valkeyi
...
 
+// buildPodTemplateAnnotations assembles the annotations that must be present on
+// the pod template spec to trigger rolling updates when the ACL secret or the
+// server config changes.
+func buildPodTemplateAnnotations(node *valkeyiov1alpha1.ValkeyNode, aclSecret *corev1.Secret) map[string]string {
+	annotations := map[string]string{
+		hashAnnotationKey: aclSecret.Annotations[hashAnnotationKey],
+	}
+	if node.Spec.ServerConfigHash != "" {
+		annotations[configHashKey] = node.Spec.ServerConfigHash
```
bjosv, line 126: Just wanted to highlight that we have a bit mixed naming for these labels:
	hashAnnotationKey = "**valkey.io/internal-acl-hash**"
	configHashKey  = "**valkey.io/config-hash**"
	scriptsHashKey = "valkey.io/script-hash"

Something to handle in another PR maybe

internal/controller/valkeycluster_controller.go
bjosv: This also _might_ be the problem in the failing testrun.
There seems to be an extra podroll in "should deploy ValkeyCluster with metrics exporter sidecar by default",
and that could have been due a timing issue on writing the config and Get:ing the config hash, so we might get an "" here... but its hard to tell from the logs really

internal/controller/valkeynode_controller.go
jdheyburn: Agree. I anticipate config-hash and script-hash to go away once we refactor the configs. We can then decide on the naming of internal-acl-hash going fwd.

internal/controller/valkeycluster_controller.go
jdheyburn: @bjosv I tried to replicate the failing e2e locally, it failed for another unrelated test. I re-ran the e2e tests on this PR and it passed. I think there might be some flakey e2e tests which could warrant an investigation. Let me know your thoughts - I am hoping to prioritise this PR so that we have rolling pods for 0.1, and iterate on it after the fact.

jdheyburn:
@bjosv Just checking despite your comments, this PR still gets your stamp of approval?

internal/controller/valkeycluster_controller.go
bjosv: Ok! We can keep an eye on it, it might be something only seen on Github at busy hours

bjosv (approved):
Looks good, was only hesitating due to the failing test.
We'll investigate the flaky tests.

Issue #134, opened 2026-04-13
Allow creating users with resetpass flag


hieu2102:
mentioned in the comments of #82; If we don't allow creating users without a password, then it will make modules that override the authentication flow (e.g valkey-ldap) unusable

sandeepkunusoth:
We need to add these users with resetpass by default

jdheyburn:
We can create users without passwords via the `NoPassword` field.  I'm not familiar with the valkey-ldap module. How does that relate to user ACLs @hieu2102?

> We need to add these users with resetpass by default

@sandeepkunusoth Why would resetpass help here?

sandeepkunusoth:
@jdheyburn if we create users with nopass anybody can login without password unless we restrict these within modules. so by default where we dont want users to login with password, we should set it as resetpass instead of nopass for custom auth like OAuth token based auth or ldap or mtls.  https://valkey.io/topics/acl/#:~:text=Note%3A%20if%20a%20user%20is%20not%20flagged%20with%20nopass%20and%20has%20no%20list%20of%20valid%20passwords%2C%20that%20user%20is%20impossible%20to%20authenticate%20via%20password.%20Such%20users%20can%20still%20be%20authenticated%20through%20mTLS%20certificate%20fields%20if%20tls%2Dauth%2Dclients%2Duser%20is%20enabled.

jdheyburn:
Okay thanks for the context. So `resetpass` is required to be set on a user for valkey-ldap to work? Do we need to support `ResetPassword` on the UserACLs config then? Do we expect the human to add in users to the UsersACLs to allow them to work with ldap? I'm not familiar with that flow.

@hieu2102 Does the name of the issue need to be changed? Since "Create user even if no password is provided" can be achieved today via `NoPassword`.

jdheyburn:
ACLs have a `permissions` field, so perhaps this could be achieved with the use of `permissions: "resetpass"`?

Is the ask to have resetpass as a supported field in the CRD?

sandeepkunusoth:
permission list is different from resetpass. i think by default user should have resetpass whenever user doesn't have nopass or specific password assigned.

bjosv:
Fixed by #166

PR #166, opened 2026-05-07
add resetpass to UserAclSpec

hieu2102:
This PR closes <Issue #134>

### Summary

This PR adds the `ResetPass` flag to `v1alpha1.UserAclSpec`

If `ResetPass` is true, it will skip the other authentication-related flag (`PasswordSecret`, `NoPassword`)

Since when setting that flag, Valkey will also remove the other two

```
127.0.0.1:6379> acl setuser test on >hello +@admin
OK

127.0.0.1:6379> acl list 
2) "user test on sanitize-payload #2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824 resetchannels alldbs -@all +@admin"

127.0.0.1:6379> acl setuser test resetpass
OK

127.0.0.1:6379> acl list
2) "user test on sanitize-payload resetchannels alldbs -@all +@admin"
```

```
127.0.0.1:6379> acl setuser test nopass
OK

127.0.0.1:6379> acl list 
2) "user test on nopass sanitize-payload resetchannels alldbs -@all +@admin"

127.0.0.1:6379> acl setuser test resetpass
OK

127.0.0.1:6379> acl list 
2) "user test on sanitize-payload resetchannels alldbs -@all +@admin"

```

### Implementation
Add the field `ResetPass` to `v1alpha1.UserAclSpec`, default value is false

### Limitations
N/A

### Testing
- add unit test to verify if `ResetPass` is added, the ACL entry is created with no password or `nopass` flag
- add additional checks to e2e-test to verify the flag is passed correctly from yaml config 

### Checklist


config/samples/v1alpha1_valkeycluster-with-user.yaml
```diff
@@ -50,6 +50,11 @@ spec:
           - davidnew
       commands:
         allow: ["@admin"]
+    - name: edward
+      enabled: true
+      resetpass: true
+      nopass: true
```
bjosv: Should we remove `nopass` in this example/sample since its ignored now?
`If ResetPass is true, it will skip the other authentication-related flag (PasswordSecret, NoPassword)`

..and don't we need to change the check in `fetchUserPasswords` too?
Maybe I have missed something.

hieu2102: >Should we remove nopass in this example/sample since its ignored now?

about this, I leave it in to test the value is ignored if we set `resetpass`, will add a comment about it to be clearer

internal/controller/users.go
```diff
@@ -226,11 +226,14 @@ func buildUserAcl(user valkeyiov1alpha1.UserAclSpec, passwords []string) string
...
-	} else {
-		appendAcl(&acl, passwords, "#")
+	// If resetpass flag is false, then add password(s)/nopass flag to the ACL
+	if !user.ResetPass {
+		// If enabled, append password(s), which should already be prefix-hashed
+		if user.NoPassword {
+			fmt.Fprintf(&acl, " nopass")
+		} else {
+			appendAcl(&acl, passwords, "#")
+		}
```
sandeepkunusoth, line 236: is resetpass added when its enabled?

```
if user.ResetPass {
    fmt.Fprintf(&acl, " resetpass")
}
```

hieu2102: no, when `ACL LIST`ing after setting the `resetpass` flag, valkey wont show it
```
127.0.0.1:6379> acl setuser test resetpass
OK

127.0.0.1:6379> acl list
2) "user test on sanitize-payload resetchannels alldbs -@all +@admin"
```

So I just skip auth-related flags if `resetpass` is enabled

config/samples/v1alpha1_valkeycluster-with-user.yaml
bjosv: Ah, ok. I was thinking that this file is placed in the samples directory, and might be used as a reference.
So in that case it would be nice to have it as correct as possible.
Optionally this specific scenario could be checked by a specific testcase instead.

hieu2102: Make sense, let me remove it then

hieu2102:
the CICD e2e-tests failed on the metrics reporter tests, should I wait for #173 to be merged first?

bjosv:
> the CICD e2e-tests failed on the metrics reporter tests, should I wait for #173 to be merged first?

Exactly, it hit the issue fixed by #173, I re-ran the test to see if we hit anything else

test/e2e/valkeycluster_test.go
```diff
@@ -476,13 +476,104 @@ spec:
...
 					ContainSubstring("user david on"),
+					ContainSubstring("user edward on"),
 					ContainSubstring("user _exporter on"),
 					ContainSubstring("user _operator on"),
 				))
 			}
 			Eventually(verifyCreatedUsers).Should(Succeed())
 		})
 
+		It("skips authentication configs if user is created with resetpass flag", Label("authentication", "resetpass"), func() {
```
sandeepkunusoth: nit: instead of adding new e2e test can we update exisitng e2e test It("creates a cluster with custom users"

we can assert here nopass not to be present for resetpass users and nopass present for resetpass disabled users. as its specific to users this can be convered in exisitng tests. i feel creating seperate cluster for this is not neccessary

hieu2102: So update the existing test to use hard-coded YAML instead of reading from the sample folder? @bjosv's comment that we should keep the sample files as correct as possible, since they are more likely to be used as starting points for users

sandeepkunusoth: i dont think any changes required in sample users  yaml as it will anyway contain all types of users examples

test/e2e/valkeycluster_test.go
```diff
@@ -476,13 +476,104 @@ spec:
...
+				g.Expect(err).NotTo(HaveOccurred())
+
+				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+				g.Expect(cr.Status.Shards).To(Equal(int32(1)))
+				g.Expect(cr.Status.ReadyShards).To(Equal(int32(1)))
+
+			}
+			Eventually(verifyCrStatus, 5*time.Minute, 2*time.Second).Should(Succeed())
+
+			By("verifying created usuers")
```
bjosv:
```suggestion
			By("verifying created users")
```

test/e2e/valkeycluster_test.go
```diff
@@ -476,13 +476,104 @@ spec:
...
+			verifyCrStatus := func(g Gomega) {
+				cr, err := utils.GetValkeyClusterStatus(valkeyClusterName)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+				g.Expect(cr.Status.Shards).To(Equal(int32(1)))
+				g.Expect(cr.Status.ReadyShards).To(Equal(int32(1)))
+
+			}
+			Eventually(verifyCrStatus, 5*time.Minute, 2*time.Second).Should(Succeed())
```
bjosv: 1 shard and 5 minutes seems long, but maybe its needed? or can default 2 min be enough?
```suggestion
			Eventually(verifyCrStatus).Should(Succeed())
```

test/e2e/valkeycluster_test.go
```diff
@@ -476,13 +476,104 @@ spec:
...
+				output, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				cmd = exec.Command("kubectl", "delete", "pod", "client",
+					"--wait=true", "--timeout=30s")
+				_, err = utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				g.Expect(output).To(SatisfyAll(
+					ContainSubstring("user alice on"),
```
bjosv: Since we only check a substring; cant the string also contain `nopass` or `#...` after `on`, and we pass the test anyway?

test/e2e/valkeycluster_test.go
bjosv: I guess this PR needed a test to make sure a `resetpass` skipped generating a `nopass`, even when configured so.
To me the added unittest in TestBuildAclFileContents could be good enough. It makes sure there is no `nopass`.
@sandeepkunusoth so you only wanted to add new assert to the existing e2e test "creates a cluster with custom users" and remove this then?

bjosv (approved):
Nice!

jdheyburn (approved):
LGTM too. I don't have the greatest context on how this is then used in LDAP, but happy to get this in so that we can iterate on it 👍🏻

PR #171, opened 2026-05-11
Use status patch to eliminate resource version conflicts

bjosv:
### Summary
`updateStatus()` called `Status().Update()` on the in-memory cluster object fetched at the start of `Reconcile`. Between consecutive reconciles, the informer cache may not yet reflect the `resourceVersion` written by the
previous reconcile's status update, causing the next reconcile to read a stale version and fail with `the object has been modified`.

Switch to `Status().Patch()` with `client.MergeFrom` which sends only the status diff and does not conflict with concurrent updates to other fields.

Replace `statusChanged`, `nodeStatusChanged`, and `conditionsChanged` helpers with a single `reflect.DeepEqual` guard that skips the patch call entirely when nothing changed, avoiding unnecessary API requests.

This fixes intermittent e2e test timeouts where clusters failed to reach Ready state due to accumulated retry delays from status conflicts.

### Testing

Have run all e2e tests multiple times to get an understanding of the intermittent issues.
Example cases:
- "creates a cluster with custom users" - cluster stuck in Reconciling
- "should deploy ValkeyCluster with metrics exporter sidecar by default" - pod recreated with new IP due to cascading conflicts 
- "should deploy ValkeyCluster without metrics exporter when disabled" - client pod AlreadyExists from delayed cluster formation

All share the same root cause: status update conflicts delaying reconciliation.

### Checklist


jdheyburn (approved):
Thank you!

PR #175, opened 2026-05-12
feat: add watch-namespace flag to filter namespaces to watch for resources

### LINKED ISSUE #178: Filter namespaces for the operator to watch
If managed resources are going to only ever be deployed to select namespaces, we can introduce a list of namespaces that the operator needs to watch. This can help to lower memory use by having the operator only watch for resources it needs to.

jdheyburn:
Closes #178 

### Summary

Adds a new flag `--watch-namespace` that will limit the resources watched by the managers. This will help to reduce memory usage in larger clusters. The flag can be specified multiple times to limit to multiple namespaces.

Lower memory use

### Testing

Tested locally by building the image and deploying to local kind cluster.

```yaml
- args:
  # ...
  - --watch-namespace=valkey-operator-system
  - --zap-log-level=4
```

```
2026-05-12T10:06:58Z    INFO    setup    Restricting cache to namespaces    {"namespaces": ["valkey-operator-system"[]}
```

### Checklist


cmd/main.go
```diff
@@ -82,6 +87,21 @@ func main() {
 	flag.StringVar(&metricsCertKey, "metrics-cert-key", "tls.key", "The name of the metrics server key file.")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
+	seenNamespaces := make(map[string]struct{})
+	flag.Func("watch-namespace", "Namespace to watch (repeatable; omit for cluster-wide)", func(s string) error {
```
sandeepkunusoth, line 96: how can users set this watch-namespace envvar from helm or operator env? i think we need to document that part

jdheyburn: I've got a draft PR for the helm chart here: https://github.com/valkey-io/valkey-helm/pull/172

Is that what you had in mind?

For env var, can any of these arguments today be set via env var? I hadn't had that in scope.

bjosv (dismissed):
Some operators seem to only, or also, use the env. variable `WATCH_NAMESPACE`, like [dragonfly-opeator](https://github.com/dragonflydb/dragonfly-operator/blob/4de29bb4386a386a8dfb9e544eaff63a146a3c48/cmd/main.go#L198), [cnpg](https://github.com/cloudnative-pg/cloudnative-pg/blob/6b885c8ea3fb7b1bca08aa1b076c628a17ba66cf/internal/configuration/configuration.go#L94), [OT Redis operator](https://github.com/OT-CONTAINER-KIT/redis-operator/blob/33dce92ac2ae6ce7ec4613cbabfc76f369051720/internal/envs/envs.go#L31) and [cockroach-operator](https://github.com/cockroachdb/cockroach-operator/blob/14a70d3d73f860e9b7318cfbe9f0f52111280290/cmd/cockroach-operator/main.go#L42).
It might be a convention from the Operator SDK, and some state that its simpler with env. than injecting extra args.
An env var would need to be comma separated.

Looking at the helm chart PR which uses your args; your solution seems cleaner.

jdheyburn:
@bjosv I wonder if there is a way to have our args defined as env vars too.

cmd/main.go
```diff
@@ -114,6 +133,33 @@ func main() {
 		tlsOpts = append(tlsOpts, disableHTTP2)
 	}
 
+	// Only cache Kubernetes resources managed by this operator to reduce memory
```
bjosv: Less diff if we don't move it, but maybe it fits better here?

bjosv (dismissed):
LGTM
The test failed due to an ValkeyNodeFailed event, and this seem to be a pre-existing race condition during initial cluster creation

`Failed to reconcile ValkeyNode: Operation cannot be fulfilled on valkeynodes.valkey.io "cluster-sample-0-1": the object has been modified; please apply your changes to the latest version and try again` 
Then we try again and it works then. In the future, maybe we should change this event to be skipped when its a conflict.

cmd/main.go
jdheyburn: Merge conflict went AWOL... then I got distracted with something else before checking the diff - I've fixed it now!

jdheyburn:
@bjosv should be ready for final review now.

For that race condition, I think we should try to replace whatever Apply is causing that with a Patch if it makes sense to.

bjosv (approved):

bjosv:
> For that race condition, I think we should try to replace whatever Apply is causing that with a Patch if it makes sense to.

Ah, yes that might be a better solution. We'll see if it pops up again

jdheyburn:
Helm chart PR to support this value:

- https://github.com/valkey-io/valkey-helm/pull/172/

PR #182, opened 2026-05-14
[feat] Add minimal pod disruption budget

jdheyburn:
### Summary

Adds a pod disruption budget to ensure that voluntary pod cycles are limited to 1 at a time.

Adds a new field `podDisruptionBudget` which takes an enum, defaulting to `Managed` to indicate that a PDB should be created.

Currently this defaults to `maxUnavailable: 1`, with selectors against `valkey.io/cluster=<CLUSTER_NAME>`

### Implementation

```
podDisruptionBudget: Managed # default

# disable it with:
podDisruptionBudget: Disabled
```

Enum chosen over boolean enabled/disabled in case additional modes are added later on.

For future use, we can switch it to be:

```
podDisruptionBudget:
  policy: Managed
  # ... more fields here
```

### Limitations

No customisation yet, we can add this later on while we're still in alpha.

### Testing

e2e tests passed locally

### Checklist

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -107,7 +107,7 @@ type ValkeyClusterReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.3/pkg/reconcile
-func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) { //nolint:gocyclo
 	log := logf.FromContext(ctx)
 	log.V(1).Info("reconcile...")
 
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -105,7 +107,7 @@ type ValkeyClusterReconciler struct {
 //
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.23.3/pkg/reconcile
-func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) { //nolint:gocyclo
```
jdheyburn, line 110: This change tips the complexity over at 31 (we fail on >30). Adding this nolint temporarily, will raise a follow up PR to split the function up.

internal/controller/valkeycluster_pdb.go
```diff
@@ -0,0 +1,82 @@
...
+	maxUnavailable := intstr.FromInt32(1)
+	existing := &policyv1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pdbName(cluster),
+			Namespace: cluster.Namespace,
+		},
+	}
+	result, err := controllerutil.CreateOrUpdate(ctx, r.Client, existing, func() error {
+		existing.Labels = labels(cluster)
+		existing.Spec.MaxUnavailable = &maxUnavailable
```
bjosv, line 64: Should we reset other values in case they were changed by anyone, like
`existing.Spec.MinAvailable = nil`

Maybe not needed?

internal/controller/valkeycluster_pdb.go
```diff
@@ -0,0 +1,82 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
+)
+
+func pdbName(cluster *valkeyiov1alpha1.ValkeyCluster) string {
+	return "valkey-" + cluster.Name
```
bjosv, line 35: Configmaps uses "valkey-" + cluster.Name + "-config", but I think its fine to skip -pdb here since its a PodDisruptionBudget resource, but maybe we should remove `-config` instead in another PR?

bjosv (commented):
LGTM

internal/controller/valkeycluster_pdb.go
jdheyburn: I'm not really a fan of including the Kind in the name; i.e. we're already querying for ConfigMaps, do we need `-config` appended?

I agree with you. I've raised an issue:

- https://github.com/valkey-io/valkey-operator/issues/184

internal/controller/valkeycluster_pdb.go
jdheyburn: That is a great shout. I've added a commit for that.

bjosv (approved):

PR #194, opened 2026-05-22
Filter informer cache to operator-managed resources only

### LINKED ISSUE #186: Filter resources the operator watches/caches to its own managed resources
From #176 

Currently the operator is not limited to the resources it watches and caches. This unbounded watching is causing high memory. The Operator should only be concerned about the resources it creates, which would always be labelled with `app.kubernetes.io/managed-by: valkey-operator`.

Introduce a filter for this label.

bjosv:
This PR closes #186

### Summary
Add label selector filtering to the manager cache so only resources with `app.kubernetes.io/managed-by=valkey-operator` are cached. This reduces memory usage in large clusters where many unrelated resources exist.

Filtered types: Service, StatefulSet, Deployment, Pod, ConfigMap, PersistentVolumeClaim, PodDisruptionBudget.

Secrets are left unfiltered because the operator reads user-provided password Secrets that don't carry the managed-by label.

### Testing

In a test cluster (kind) with 5000 ConfigMaps, 1000 Services, and 200 Deployments from other workloads, operator RSS dropped from 76MB to 57MB (~25% reduction).

### Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -27,10 +27,16 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
 	versioncollector "github.com/prometheus/client_golang/prometheus/collectors/version"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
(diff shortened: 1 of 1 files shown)


jdheyburn (approved):
Nice one!

PR #196, opened 2026-05-24
fix: handle fragmented slot ranges in assignSlotsToPendingPrimaries

bjosv:
### Summary

When both primary and replica pods are lost for a shard, the operator must reassign that shard's slots to a new primary. The previous implementation calculated slots-per-node as TotalSlots / total_shards and distributed the remainder based on a global shard index. During recovery the shard index of the replacement didn't qualify for the extra slot, leaving 1 slot permanently unassigned.

Additionally, it assumed unassigned slots formed a single contiguous range, which fails when slots are fragmented after rebalancing.

### Features / Behaviour Changes
- Calculate slots-per-node from actual unassigned count divided by pending primaries (equivalent as for initial creation)
- Consume from multiple fragmented ranges per primary
- Issue a single CLUSTER ADDSLOTSRANGE call per node with all ranges
- Add SlotsRange.String() and FormatSlotsRanges for readable event messages

### Testing

The issue can be reproduced using:
```
# Create a cluster (v0.1.0)
kind create cluster
helm repo update
helm install valkey-operator valkey/valkey-operator -n valkey-operator-system --create-namespace

cat <<EOF | kubectl apply -f -
  apiVersion: valkey.io/v1alpha1
  kind: ValkeyCluster
  metadata:
    name: my-cluster
  spec:
    shards: 3
    replicas: 1
EOF
kubectl get valkeycluster

# Scale out, and in again
kubectl patch valkeycluster my-cluster --type=merge -p '{"spec":{"shards":5}}'
kubectl get valkeycluster
kubectl patch valkeycluster my-cluster --type=merge -p '{"spec":{"shards":3}}'
kubectl get valkeycluster

# Verify that slot ranges are fragmented now
kubectl exec pod/valkey-my-cluster-0-0-0 -c server -- valkey-cli CLUSTER NODES

# Delete both pods of a shard
kubectl delete pod valkey-my-cluster-0-0-0 valkey-my-cluster-0-1-0

# The cluster is now forever degraded.
kubectl logs -n valkey-operator-system deployment.apps/valkey-operator-controller-manager

# Log snippet
..  DEBUG	events	Failed to assign slots 6661-12121 to 10.244.0.17: Slot 7646 is already busy
```
### Checklist

Before submitting the PR make sure the following are checked:

- [ ] This Pull Request is related to one issue.
- [X] Commit message explains what changed and why
- [ ] Tests are added or updated.
- [ ] Documentation files are updated.
[...]

PR #197, opened 2026-05-26
Set config epoch on new nodes before MEET to prevent slot loss

bjosv:
### Summary
When new nodes join a cluster they start with config epoch 0. After `MEET`, gossip from existing members might carry slot ownership from dead nodes at higher epochs. The new node accepts this gossip, overwriting its own `ADDSLOTS` claims. When the operator later forgets the dead node, those slots become permanently unassigned.

This leads to a cluster stuck in Reconciling state.

Call `CLUSTER SETCONFIGEPOCH` on each isolated node before `MEET`, setting it above the cluster's current epoch. This ensures the new node's epoch is higher than any dead node it replaces, so gossip cannot override its slot claims. The command fails gracefully on nodes that already have a non-zero epoch.

### Testing

To reproduce:
1. Deploy a ValkeyCluster (e.g., 3 shards, 1 replica each)
2. Delete both the primary and replica Pods of a shard simultaneously
3. Wait for the operator to replace the nodes
4. Occasionally the cluster gets stuck in Reconciling state with permanently unassigned slots (depends on gossip timing)

The Valkey server log on the new node will show:
```
Processing UPDATE message received from <meet-target> about node <dead-node>. old configEpoch 0, new configEpoch <higher>
Slot range [X, Y] is migrated from node <new-node> to node <dead-node>
```
With this correction the UPDATE message will be ignored because the new node's epoch is already higher than the dead node's.

### Checklist

### DIFF
--- a/internal/valkey/clusterstate.go
+++ b/internal/valkey/clusterstate.go
@@ -209,8 +209,8 @@ func (n *NodeState) IsPrimary() bool {
 // CurrentEpoch returns the cluster's current epoch as reported by this node's
 // CLUSTER INFO output.
 func (n *NodeState) CurrentEpoch() int64 {
-	if v, err := strconv.ParseInt(n.ClusterInfo["cluster_current_epoch"], 10, 64); err == nil {
-		return v
+	if epoch, err := strconv.ParseInt(n.ClusterInfo["cluster_current_epoch"], 10, 64); err == nil {
+		return epoch
 	}
 	return 0
 }
--- a/internal/valkey/clusterstate.go
+++ b/internal/valkey/clusterstate.go
@@ -206,6 +206,15 @@ func (n *NodeState) IsPrimary() bool {
 	return slices.Contains(n.Flags, "master")
 }
 
+// CurrentEpoch returns the cluster's current epoch as reported by this node's
+// CLUSTER INFO output.
+func (n *NodeState) CurrentEpoch() int64 {
+	if v, err := strconv.ParseInt(n.ClusterInfo["cluster_current_epoch"], 10, 64); err == nil {
+		return v
+	}
+	return 0
+}
+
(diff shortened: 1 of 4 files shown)


internal/valkey/clusterstate.go
```diff
@@ -206,6 +206,15 @@ func (n *NodeState) IsPrimary() bool {
 	return slices.Contains(n.Flags, "master")
 }
 
+// CurrentEpoch returns the cluster's current epoch as reported by this node's
+// CLUSTER INFO output.
+func (n *NodeState) CurrentEpoch() int64 {
+	if v, err := strconv.ParseInt(n.ClusterInfo["cluster_current_epoch"], 10, 64); err == nil {
+		return v
```
jdheyburn: Could we use `epoch` or `currentEpoch` instead of `v`?

jdheyburn:
I wasn't able to replicate this on main, but this is after I merged in the below PR, so that might have helped reduce this happen?

- https://github.com/valkey-io/valkey-operator/pull/196

Also you might need a dummy commit to kick off CI run.

bjosv:
> I wasn't able to replicate this on main, but this is after I merged in the below PR, so that might have helped reduce this happen?

Running with the fix for fragmented slot I needed to run ~60-102~ 10-20 loop-iterations of only killing both pods before hitting it. So very seldom..but it's a race condition dependent on gossip timing.
I have created a testcase that runs different scenarios in a loop, and this is how I found these issues.
I will push a (draft) PR with it.

internal/valkey/clusterstate.go
bjosv: Ah yes, done

jdheyburn (approved):

bjosv:
Rerun the tests on main after hitting another timing issue, this time with Deployments and shardExistsInTopology().
Investigating.

PR #207, opened 2026-05-29
fix: check primary node first when resolving shard index

bjosv:
### Summary

During scale-in, `shardIndexFromState` could match a stale replica from a drained shard that temporarily appears in a remaining shard via gossip before `CLUSTER FORGET` propagates. This returns the wrong shard index, causing the controller to drain the wrong shards, e.g. draining 3 shards out of 4 total, instead of just draining 1 shard, leading to an unrecoverable Reconciling state.

Fixed by checking the primary node first. The primary is the authoritative slot owner and its ValkeyNode CR always has the correct shard-index label. Stale replicas from drained shards are never primaries of remaining shards.

### Testing

This has been found using #203 running:
`CHAOS_SCENARIOS=scale-shards CHAOS_MIN_SHARDS=3 CHAOS_MAX_SHARDS=9 make test-chaos`

This repeatedly scales the cluster to a random shard count (between 3–9) and verifies it recovers correctly each time, running until failure.

Previously failed within 3–20 iterations; now passes 1000+ without failure.

### Checklist

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -1207,9 +1207,16 @@ func (r *ValkeyClusterReconciler) deleteExcessValkeyNodes(ctx context.Context, c
 	return deleted, nil
 }
 
-// shardIndexFromState determines the pod shard-index for a given Valkey shard
-// by matching any of its nodes' addresses to ValkeyNode labels.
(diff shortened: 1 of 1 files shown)


jdheyburn (approved):
Makes sense, thank you! The chaos testing suite really was worth the time you put into it!

PR #208, opened 2026-05-29
fix: rolling update loses all keys when replicas are not synced

bjosv:
### Summary

This PR fixes two issues that together make rolling updates safe with and without PVCs.

1. Without persistence: restarting a replica destroys its cluster membership and data. The operator considered it Ready and immediately rolled the primary. With no synced replica to fail over to, all shard data was lost.

3. With persistence: the replica stays synced and proactive failover is attempted, but the `_operator` user ACL was missing `cluster|failover`. Every failover failed with NOPERM and the primary was rolled without a graceful handoff.

### Testing

Found using #203 running the rolling-update scenario. It patches io-threads on the ValkeyCluster to trigger a rolling restart of all pods, then verifies the cluster recovers and all keys are preserved.

```
# Without persistence
CHAOS_SCENARIOS=rolling-update make test-chaos

# With persistence
CHAOS_SCENARIOS=rolling-update CHAOS_PERSISTENCE=true make test-chaos
```

### Checklist

### DIFF
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -57,6 +57,7 @@ var (
 			"+cluster|addslotsrange",     // assign slots
 			"+cluster|replicate",         // set up replication
 			"+cluster|forget",            // remove stale nodes
+			"+cluster|failover",          // proactive failover before rolling primary
 			"+cluster|getslotmigrations", // check migration status
 			"+cluster|migrateslots",      // migrate slots between shards
 			"+cluster|set-config-epoch",  // set epoch on new nodes
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -472,12 +472,28 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNode(ctx context.Context, clust
 					"name", node.Name, "err", err)
 			}
 		} else if nodeRequiresRoll(current, desired) {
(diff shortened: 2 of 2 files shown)


internal/controller/users.go
```diff
@@ -57,6 +57,7 @@ var (
 			"+cluster|addslotsrange",     // assign slots
 			"+cluster|replicate",         // set up replication
 			"+cluster|forget",            // remove stale nodes
+			"+cluster|failover",          // proactive failover before rolling primary
```
jdheyburn, line 60: Oh man we really need to get some CI check to make sure the operator has everything it needs... This particular line was on me - so my bad

jdheyburn:
Something I've been wanting to introduce at some point is better readiness probes - perhaps gated on cluster_state:ok. I tried to do that in #120 but there were some other instabilities which prevented reconcile from progressing. I think that would help prevent the operator from rolling pods when the cluster is unhealthy, and also prevent k8s from descheduling pods when it is reconciling.

internal/controller/users.go
bjosv: I think we merged ACL stuff PR right after the proactive PR, so we missed it.
Also, I guess a test of failover pass without the proactive feature since there is a fallback (except it lost data).
But there is probably a way to check for NOPERM or some other way

jdheyburn (approved):

PR #209, opened 2026-05-29
feat: Live configuration apply framework with MVP configs

### LINKED ISSUE #210: [feat] Apply minimum set of configs live without rolls
We require a minimum framework in place to allow applying of configurations live onto nodes without the need of rolling them.

jdheyburn:
This PR closes #210 

### Summary

Initial working MVP of applying live-settable configurations on a Valkey Node, avoiding the need to introduce pod rolls for configs that don't require rolling.

Supports these configs today, any others will be evaluated on user requests.

```
maxclients
maxmemory
maxmemory-policy
```

### Implementation

I decided against the per-node ConfigMap as I mentioned in the design #141. There wasn't an immediate need for it for the keys that we want to add. Future iteration to crack down on the limitations outlined below may reintroduce them.

### Limitations

- If a previously managed config was removed, the operator does not unset it back to the default
  - Follow up issue: #211 
- No per-node configuration supported
  - I don't anticipate users will want to specify per-node configs -- we designed the CRD with that in mind
  - In the future we may want to support a different configs per node that are managed by the operator, such as `replica-priority` being 50 for node index 0, 75 for node index 1, and 100 for node index 2.
- Value validation is performed on CONFIG SET, operator does not reject invalid configs (awaiting [CONFIG INFO](https://github.com/valkey-io/valkey/pull/3050))
- Adding to allowlist in the operator causes pods to be rolled
  - I'll look to iterate on this, I am hoping to remove `serverConfigRollHash` from the ValkeyNode spec - since I don't like exposing an implementation detail to the public CRD
- ACL changes are not applied live, will be implemented in #212 

### Testing

Validated locally

### Checklist


internal/controller/valkeynode_controller.go
```diff
@@ -45,12 +47,50 @@ const (
...
+func (rc *realValkeyConfigClient) Close() { rc.client.Close() }
+
+// realConfigClient opens a real Valkey connection to the node's pod.
+func realConfigClient(ctx context.Context, r *ValkeyNodeReconciler, node *valkeyiov1alpha1.ValkeyNode) (valkeyConfigClient, error) {
+	c, err := vclient.NewClient(r.buildNodeClientOption(ctx, node))
+	if err != nil {
+		return nil, err
+	}
+	return &realValkeyConfigClient{client: c}, nil
+}
```
jdheyburn, line 83: In a future PR I would like to tidy up how the operator connects to the nodes.

Rather than each consumer creating its own client, they should all share the same connection to nodes - this would help tidy things up.

jdheyburn:
@greptile-apps

jdheyburn:
@greptile-apps

internal/controller/valkeynode_controller.go
```diff
@@ -100,11 +140,75 @@ func (r *ValkeyNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	applied, err := r.applyLiveConfig(ctx, node)
```
bjosv, line 144: This will apply the configs in every reconciliation, i.e every 60s.

I guess we could check the nodes ValkeyNodeConditionLiveConfigApplied condition to see if its ObservedGeneration is the same as the nodes Generation,
i.e. if we are already synced (Generation) and last CONFIG SET succeeded (True) we skip sending configs again and again.

internal/controller/valkeynode_controller.go
```diff
@@ -483,6 +588,32 @@ func (r *ValkeyNodeReconciler) getValkeyRole(ctx context.Context, node *valkeyio
...
+// when there is nothing to apply, and (false, err) on failure.
+func (r *ValkeyNodeReconciler) applyLiveConfig(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) (bool, error) {
+	params := liveConfigToApply(node.Spec.Config)
+	if len(params) == 0 {
+		return false, nil
+	}
+
+	factory := r.newConfigClient
+	if factory == nil {
+		factory = realConfigClient
```
bjosv: We set `ValkeyNodeReconciler.APIReader` in SetupWithManager, but here we only use ValkeyNodeReconciler.newConfigClient for mocking.

 I believe the tests would would work if set (new)ConfigClient to realConfigClient SetupWithManager(?). Maybe I missed some reason for doing this manuever.

bjosv (commented):
Nice work, added some comments

jdheyburn:
@greptile-apps

internal/controller/valkeynode_controller.go
jdheyburn: True, although if we were to gate on the condition and generation then we would not be correcting any drift for managed configs - which I would expect an operator to do IMO.

There is connection churn in this PR which I would say is more expensive, so I'm working on a client registry that keeps a connection open to every node for the operator.

What do you think?

internal/controller/valkeynode_controller.go
jdheyburn: No real reason - I've implemented your suggestion, thanks for the prompt, let me know how it looks now!

jdheyburn:
@greptile-apps

melancholictheory:
nice, rendering the roll hash from the config minus the live keys (buildRollServerConfig) is the clean way to do it, a change confined to allowlisted keys genuinely can't move the hash. batching every param into one CONFIG SET is also right since multi-param CONFIG SET is atomic, so you never get half-applied state. one real gap from when we built this though:

the live keys still get written into the shared ConfigMap at their new value (buildServerConfig renders everything), but the CONFIG SET that actually validates them runs later in the node controller. so if a value is bad, say maxmemory set to something CONFIG SET rejects, the live apply fails and you correctly surface LiveConfigApplied=False, but the ConfigMap already has the bad value. nothing rolls, the node keeps running on the old value, and it looks handled. then later that node restarts for some unrelated reason, reads valkey.conf out of the ConfigMap, and boots with the bad value. if it's a parse-level reject that's a CrashLoopBackOff out of nowhere; if it's just semantically off it silently takes effect. the config you "rejected" becomes the restart-time config.

this is independent of the CONFIG INFO pre-validation you're waiting on, it's an ordering thing. what worked for us: don't persist a live-settable key into the file until CONFIG SET has accepted it on the node, so the file only ever holds values the server has proven it accepts and restart-time always matches live-time. it's a bit awkward with one shared ConfigMap (the file is cluster-wide but the apply is per-node), but even gating the ConfigMap write on "all nodes applied clean" would close it.

smaller heads-up on maxmemory specifically: lowering it live takes effect the instant CONFIG SET lands, so if used_memory is already above the new limit the eviction (or with noeviction, the write rejection) hits immediately, on every node at once. usually that's what you want from a live knob, just worth knowing it's instant and uncoordinated compared to the roll path.

happy to go deeper on any of this.

jdheyburn:
@melancholictheory  I think that's a good call out on the ordering issue. I think any issues with applying the config live would become a restart one naturally, so we would need a way to surface this to the user. I agree that it should be via one that does not surprise them on a reboot. I'll look to iterate on this design after we've merged.

jdheyburn:
I have a conflict with the merge from main, the fix for preventing rolls after bootstrapping caused a conflict. I'm looking into fixing it now, will put this back into draft for now.

melancholictheory:
sounds good, and agree it's fine to iterate post-merge. one concrete shape for "make it a restart one naturally" when you get to it: on a CONFIG SET that fails, fold that key back into the roll hash for the node instead of leaving it live-only. that flips it into an ordinary pending roll, so it shows up in whatever rollout surfacing you already have, and the restart actually re-reads the value from the file, so live and on-disk converge instead of drifting. it's basically the inverse of #227 (a key moving the other way, live -> needs-restart), so the same classification plumbing covers both directions.

internal/controller/valkeynode_controller.go
bjosv: Sounds fair, keeping as-is is fine with me

bjosv (approved):
LGTM!

nitaicaro:
@jdheyburn FYI - valkey-io/valkey#3050 has been merged, so the pre-validation limitation you mentioned here can be addressed now.

PR #222, opened 2026-06-04
fix: disable replica migration and validity factor in cluster config

### LINKED ISSUE #216: [feat] Disable `cluster-allow-replica-migration` by default for clusters
We should consider setting `cluster-allow-replica-migration no` in the operator's default [Valkey configuration](https://github.com/valkey-io/valkey/blob/12061032d0099c4033f29db0123a1c6d6de4888d/valkey.conf#L1911).

### What this setting does

When `cluster-allow-replica-migration` is enabled (the default), Valkey automatically migrates replicas between shards to maintain replica balance. Specifically:

1. **Replica migration to orphaned primaries**
When a primary in another shard has no replicas, a replica from a shard with "extra" replicas can automatically migrate to become a replica of that orphaned primary.

2. **Cross-shard auto-demotion**
When a primary loses ALL its slots to a different shard, it automatically becomes a replica of that shard.

This behavior is designed for manually-managed clusters where automatic healing is desirable.

### Why disable it for operator-managed clusters

In an operator-managed environment, the operator should have full control over cluster topology. Automatic replica migration can cause:

- **Unexpected topology changes** during scaling operations
- **Race conditions** where the operator's view of cluster state becomes inconsistent with reality
- **Interference with planned operations** like scale-down, where the operator expects specific nodes to remain in specific shards

I have observed this during chaos testing where nodes were automatically migrated between shards during scale-down operations, causing the operator's shard classification to become incorrect.

### Questions for discussion

1. Are there scenarios where automatic replica migration is beneficial?
2. Are there any failure recovery scenarios where disabling this could be problematic?

bjosv:
This PR closes #216 

### Summary

Add `cluster-allow-replica-migration=no` to prevent Valkey from moving replicas between shards autonomously, which conflicts with the operator's topology management.

Add `cluster-replica-validity-factor=0` so replicas always attempt failover regardless of disconnection time. With cluster-node-timeout at 2s, the default factor of 10 gives only a 30s window before replicas refuse to failover, causing stuck clusters under disruption.

Reorder `PlanDrainMove` to check slot count before primary existence, preventing a spurious error on already-drained shards. Returning an error due to the primary would cause the caller (drainExcessShards) to propagate it up, halting scale-down and leaving stale ValkeyNodes around forever.
This problem solved itself previously when `cluster-allow-replica-migration=yes`.

### Testing

This problem was found by using #203 and running on a machine with moderate load:
`CHAOS_MAX_SHARDS=15 CHAOS_SCENARIOS="scale-shards" make test-chaos`

### Checklist

### DIFF
--- a/internal/controller/config.go
+++ b/internal/controller/config.go
@@ -103,9 +103,11 @@ func generateValkeyNodeConfig(node *valkeyiov1alpha1.ValkeyNode) string {
 func getBaseConfig(cluster *valkeyiov1alpha1.ValkeyCluster) map[string]string {
 	baseConfig := buildManagedConfig(true, cluster.Spec.Persistence, cluster.Spec.TLS)
 	maps.Copy(baseConfig, map[string]string{
-		"cluster-enabled":      "yes",
-		"protected-mode":       "no",
-		"cluster-node-timeout": "2000",
+		"cluster-enabled":                 "yes",
+		"protected-mode":                  "no",
+		"cluster-node-timeout":            "2000",
+		"cluster-allow-replica-migration": "no",
+		"cluster-replica-validity-factor": "0",
 	})
 
 	return baseConfig
--- a/internal/valkey/cluster_rebalance.go
+++ b/internal/valkey/cluster_rebalance.go
@@ -98,16 +98,16 @@ func PlanDrainMove(src *ShardState, dsts []*ShardState, maxSlots int) (*SlotMove
 	if maxSlots <= 0 || len(dsts) == 0 {
 		return nil, nil
 	}
-	srcPrimary := src.GetPrimaryNode()
-	if srcPrimary == nil {
-		return nil, fmt.Errorf("primary missing for draining shard %s", src.Id)
-	}
 	srcCount := 0
 	for _, slot := range src.Slots {
 		srcCount += slot.End - slot.Start + 1
 	}
 	if srcCount == 0 {
-		return nil, nil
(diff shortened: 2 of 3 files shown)


jdheyburn (approved):

melancholictheory:
Nice fix — the stuck-cluster failure mode from a too-strict validity window is real and easy to hit under load.

One thing worth flagging on `cluster-replica-validity-factor 0`: it's an availability-vs-data-freshness tradeoff, and setting it unconditionally cluster-wide may surprise consistency-sensitive users. With the validity gate off, a replica is always eligible to fail over regardless of how far it has fallen behind — so if a primary dies while its only reachable replica is very stale, that replica gets promoted and silently drops the writes it never received. For availability-first / cache-like clusters that is exactly what you want (keep the shard serving; stale data is acceptable). For durable / consistency-first clusters it is arguably the wrong default — you might prefer the shard to stay down over promoting stale data.

(The offset-ranking still applies: among *eligible* replicas the best-offset one wins, so the factor only changes behavior when *every* replica is past the window — i.e. the choice is "promote the freshest-of-the-stale" vs "no election at all".)

In our operator we ended up scoping this per workload class — `cluster-replica-validity-factor 0` for availability-first clusters, and leaving the default gate in place for durable ones. Might be worth either gating the directive behind a similar availability/durability signal here, or at least documenting the freshness tradeoff alongside it.

Either way, thanks for chasing this down — the chaos-suite repro is a great catch.

PR #225, opened 2026-06-06
enhancement: Use Recreate deployment strategy

jdheyburn:
### Summary

ValkeyNode is a 1:1 mapping of node to pod. By default the Deployment's updateStrategy is set to RollingUpdate which will add a new pod first before removing the old one. This can create 2 pods for a ValkeyNode at once. Recreate will remove the old pod first before creating the new one.

### Checklist

### DIFF
--- a/internal/controller/valkeynode_resources.go
+++ b/internal/controller/valkeynode_resources.go
@@ -381,6 +381,9 @@ func buildValkeyNodeDeployment(node *valkeyiov1alpha1.ValkeyNode) (*appsv1.Deplo
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: func(i int32) *int32 { return &i }(1),
+			Strategy: appsv1.DeploymentStrategy{
+				Type: appsv1.RecreateDeploymentStrategyType,
+			},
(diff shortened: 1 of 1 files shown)


jdheyburn:
@greptile-apps

jdheyburn:
@bjosv I've lost the chaos testing thread where I mentioned this before - do you think this is worth adding?

edit: found it! #215 

- https://github.com/valkey-io/valkey-operator/pull/215#issuecomment-4605564763

sandeepkunusoth (approved):

bjosv:
> @bjosv I've lost the chaos testing thread where I mentioned this before - do you think this is worth adding?

Ah, yes, this is probably better! Will just testrun this and then close PR #215.
I now understand that .spec.strategy.type should only affect how rollouts are done, so the deployment/replicaset should still be able to start a new pod to replace an unreachable pod.

jdheyburn:
@bjosv Maybe #215 is still needed if it makes it more stable. It was a smell that Deployment would have 2 pods alive at the same time.

bjosv (approved):

PR #230, opened 2026-06-07
fix: New cluster rolls immediately after creation

### LINKED ISSUE #229: [bug] New clusters are immediately rolled after being created
When creating a new cluster via the below:

```
```

The cluster is then immediately rolled after. Given it is a new cluster, it shouldn't require a roll - it should be bootstrapped ready to go without the need to roll.

jdheyburn:
This PR closes #229 

### Summary

Stops newly provisioned clusters from rolling immediately after creation and bootstrapping.

### Implementation

The issue is because of ServerConfigHash being empty on bootstrap. It is empty because after the ConfigMap is upserted, it does not exist in the cache for the Hash to be calculated. ServerConfigHash is then set to be empty on bootstrap. After bootstrap, the Hash can be calculated after the ConfigMap is available in the cache, so this diff then causes the pods to be rolled.

The fix here is to have upsertConfigMap return the hash so that it can be immediately used to set in ServerConfigHash.

After we've implemented live config sets in #209 we can revisit the use of ServerConfigHash to see if there is some refactoring that can be done.

### Testing

Tested locally with a roll.

### Checklist

### DIFF
--- a/internal/controller/config.go
+++ b/internal/controller/config.go
@@ -143,20 +143,27 @@ func buildServerConfig(cluster *valkeyiov1alpha1.ValkeyCluster) string {
 
 // Create or update a default valkey.conf
 // If additional config is provided, append to the default map
-func (r *ValkeyClusterReconciler) upsertConfigMap(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+//
+// Returns the hash of the rendered server config. Callers MUST use this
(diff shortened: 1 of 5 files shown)


jdheyburn:
@greptile-apps

bjosv (approved):
Great fix. I suddenly see this issue affecting a lot of local tests now.

PR #235, opened 2026-06-09
(bug) fixed Liveness and Readiness Probes to use operator managed user instead of default user

### LINKED ISSUE #232: [bug] Probes fail if the default user has a password
When a password is set for the default user in Valkey, probes fail because they cannot pass the password to authenticate.

sandeepkunusoth:
This PR closes https://github.com/valkey-io/valkey-operator/issues/232

### Summary

fixed Authenticatation of Liveness and Readiness Probes with Operator User

### Changes

-  added helper returning the system-passwords Secret selector (key `_operator`).
- when the node has a cluster label, inject `VALKEY_OPERATOR_USER=_operator` plus `REDISCLI_AUTH` 
- pass `--user "$VALKEY_OPERATOR_USER"` and `--no-auth-warning` when set.

### Testing

- `make test` (envtest unit suite) — passed successfully
- `./bin/golangci-lint run` — 0 issues
- `e2e` compiles (`go vet -tags e2e ./test/e2e/...`)
- Live cluster verification:
  - A `ValkeyCluster` whose `default` user has a password reaches state `Ready` (`readyShards=3`).
  - `PING` without credentials returns `NOAUTH`.

### Checklist


internal/controller/scripts/liveness-check.sh
```diff
@@ -42,10 +42,21 @@ if [ -n "${VALKEY_TLS_ARGS:-}" ]; then
...
 fi
 
+# Authenticate as the operator-managed user when configured. valkey-cli has
+# no env var for the username, so it is passed explicitly; the password is read
+# from REDISCLI_AUTH.
+auth_args=""
+if [ -n "${VALKEY_OPERATOR_USER:-}" ]; then
+    auth_args="--user $VALKEY_OPERATOR_USER"
+fi
+if [ -n "${REDISCLI_AUTH:-}" ]; then
```
hieu2102: maybe use VALKEYCLI_AUTH instead. I think it's better to avoid reference to redis if we can

internal/controller/valkeynode_resources.go
```diff
@@ -278,6 +278,18 @@ func buildContainersDef(node *valkeyiov1alpha1.ValkeyNode) ([]corev1.Container,
 		)
 	}
 
+	// Authenticate the probes as the operator-managed "_operator" system user.
+	// The operator generates this user's password and stores it in the cluster's
+	// system-passwords Secret. The username is passed to the probe scripts via VALKEY_OPERATOR_USER;
+	// the password is read natively by valkey-cli from VALKEYCLI_AUTH.
+	if clusterName := node.Labels[LabelCluster]; clusterName != "" {
+		probeUserSecret := operatorUserPasswordSecret(clusterName)
```
sandeepkunusoth: i think there are some cases in e2e tests where internal-{name}-system-passwords secret is not getting created automatically by operator. will check on this.

internal/controller/scripts/liveness-check.sh
```diff
@@ -42,10 +42,18 @@ if [ -n "${VALKEY_TLS_ARGS:-}" ]; then
     tls_args="$VALKEY_TLS_ARGS"
 fi
 
+# Authenticate as the operator-managed user when configured. valkey-cli has
+# no env var for the username, so it is passed explicitly; the password is read
+# from VALKEYCLI_AUTH.
+auth_args=""
+if [ -n "${VALKEY_OPERATOR_USER:-}" ]; then
```
jdheyburn: Could we rename this to something more generic like `VALKEY_USER`? If we decide we don't want to use the operator user for this in the future then there are fewer lines to change. I saw that there are some comments we'd need to update too.

I wonder if its worth contributing a `VALKEYCLI_USER` env var upstream to Valkey.

internal/controller/valkeynode_resources.go
```diff
@@ -278,6 +278,18 @@ func buildContainersDef(node *valkeyiov1alpha1.ValkeyNode) ([]corev1.Container,
...
+		containers[0].Env = append(containers[0].Env,
+			corev1.EnvVar{Name: "VALKEY_OPERATOR_USER", Value: operatorUser},
+			corev1.EnvVar{Name: "VALKEYCLI_AUTH", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: probeUserSecret}},
+		)
+	}
```
jdheyburn, line 289: Why would we wrap this in a condition around `clusterName != ""`? Would this ever be null?

jdheyburn: Oh just seen your comment above.

internal/controller/valkeynode_resources.go
jdheyburn: Hmm I think this might be where we do ValkeyNode creation outside of ValkeyCluster.

Could ValkeyCluster pass in these env vars to ValkeyNode?

sandeepkunusoth: in e2e tests we are creating ValkeyNode independently today. so updated e2e tests to create operator managed secret manually for ValkeyNode cases.

sandeepkunusoth:
@greptile-apps

internal/controller/scripts/liveness-check.sh
```diff
@@ -42,10 +42,16 @@ if [ -n "${VALKEY_TLS_ARGS:-}" ]; then
     tls_args="$VALKEY_TLS_ARGS"
 fi
 
+# Authenticate as the custom user when user and password are set.
+auth_args=""
+if [ -n "${VALKEY_USER:-}" ] && [ -n "${VALKEYCLI_AUTH:-}" ]; then
+    auth_args="--user $VALKEY_USER --no-auth-warning"
```
jdheyburn: Is --no-auth-warning needed here? I think it's only needed if we use `-a` or `--auth` on the CLI.

sandeepkunusoth: you are right its not needed. removed

sandeepkunusoth:
@greptile-apps

sandeepkunusoth:
@greptile-apps

jdheyburn (approved):
LGTM - thank you!

PR #237, opened 2026-06-09
add dedicated replication system user

### LINKED ISSUE #236: Add dedicated system user for replication
As mentioned in #232, if the default user has ACL configured, then replication will break unless `primaryauth` and `primaryuser` are configured. 

To avoid this dependency on the default user configuration, we should create and use a dedicated replication user.

hieu2102:
This PR closes #236 

### Summary

Add a dedicated Valkey user for replication. 

- Add the system user `_replication`.
- Add `primaryauth`, `primaryuser` to Valkey's pod startup args

### Checklist

### DIFF
--- a/internal/controller/config_test.go
+++ b/internal/controller/config_test.go
@@ -49,6 +49,8 @@ var _ = Describe("When creating a cluster", Label("userconfig"), func() {
 
 		// Primary fails over to a replica on SIGTERM (graceful shutdown)
 		Expect(testConfigString).To(ContainSubstring("shutdown-on-sigterm failover"))
+		
+		Expect(testConfigString).To(ContainSubstring("primaryauth sup3rSecr3tP4ssw0rd"))
 	})
 })
 
--- a/internal/controller/config_test.go
+++ b/internal/controller/config_test.go
@@ -49,7 +49,7 @@ var _ = Describe("When creating a cluster", Label("userconfig"), func() {
 
 		// Primary fails over to a replica on SIGTERM (graceful shutdown)
 		Expect(testConfigString).To(ContainSubstring("shutdown-on-sigterm failover"))
-		
+
 		Expect(testConfigString).To(ContainSubstring("primaryauth sup3rSecr3tP4ssw0rd"))
 	})
 })
--- a/internal/controller/config_test.go
+++ b/internal/controller/config_test.go
@@ -50,7 +50,6 @@ var _ = Describe("When creating a cluster", Label("userconfig"), func() {
 		// Primary fails over to a replica on SIGTERM (graceful shutdown)
 		Expect(testConfigString).To(ContainSubstring("shutdown-on-sigterm failover"))
 
-		Expect(testConfigString).To(ContainSubstring("primaryauth sup3rSecr3tP4ssw0rd"))
 	})
 })
 
--- a/internal/controller/config_test.go
+++ b/internal/controller/config_test.go
@@ -49,7 +49,6 @@ var _ = Describe("When creating a cluster", Label("userconfig"), func() {
 
 		// Primary fails over to a replica on SIGTERM (graceful shutdown)
 		Expect(testConfigString).To(ContainSubstring("shutdown-on-sigterm failover"))
-
 	})
 })
 
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -44,7 +44,8 @@ const (
 var (
 	operatorUser    = "_operator"
 	exporterUser    = "_exporter"
-	systemUsers     = []string{operatorUser, exporterUser}
+	replicationUser = "_replication"
+	systemUsers     = []string{operatorUser, exporterUser, replicationUser}
 	systemUsersAcls = map[string]string{
 		operatorUser: strings.Join([]string{
 			"+@connection",               // client handshake (CLIENT TRACKING, SETINFO, AUTH, PING)
@@ -65,6 +66,8 @@ var (
 		}, " "),
 		// the ACL rawstring for exporter is taken from the redis_exporter documentation: https://github.com/oliver006/redis_exporter#authenticating-with-redis
 		exporterUser: "-@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking",
+		// the ACL rawstring for replication is taken from Valkey documentation: https://valkey.io/topics/acl/#acl-rules-for-sentinel-and-replicas
+		replicationUser: "-@all +psync +replconf +ping",
 	}
 )
 
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -136,7 +136,8 @@ func (r *ValkeyClusterReconciler) createSystemUsersAcl(ctx context.Context, clus
 				return "", err
 			}
 		} else {
-			return "", err
+			// error is unknown system user, either by manual modification, or removal from between valkey-operator versions
+			log.Error(err, "error validating system user secret")
 		}
 	}
 
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -68,7 +68,7 @@ var (
 		// the ACL rawstring for exporter is taken from the redis_exporter documentation: https://github.com/oliver006/redis_exporter#authenticating-with-redis
 		exporterUser: "-@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking",
 		// the ACL rawstring for replication is taken from Valkey documentation: https://valkey.io/topics/acl/#acl-rules-for-sentinel-and-replicas
-		replicationUser: "-@all +psync +replconf +ping",
+		replicationUser: "-@all +psync +replconf +ping +cluster|syncslots",
 	}
 )
 
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -67,8 +67,11 @@ var (
 		}, " "),
 		// the ACL rawstring for exporter is taken from the redis_exporter documentation: https://github.com/oliver006/redis_exporter#authenticating-with-redis
 		exporterUser: "-@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking",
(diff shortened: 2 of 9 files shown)


internal/controller/users.go
```diff
@@ -65,6 +66,8 @@ var (
 		}, " "),
 		// the ACL rawstring for exporter is taken from the redis_exporter documentation: https://github.com/oliver006/redis_exporter#authenticating-with-redis
 		exporterUser: "-@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking",
+		// the ACL rawstring for replication is taken from Valkey documentation: https://valkey.io/topics/acl/#acl-rules-for-sentinel-and-replicas
+		replicationUser: "-@all +psync +replconf +ping",
```
jdheyburn: Nice

internal/controller/config.go
```diff
@@ -100,22 +100,24 @@ func generateValkeyNodeConfig(node *valkeyiov1alpha1.ValkeyNode) string {
...
+func getBaseConfig(cluster *valkeyiov1alpha1.ValkeyCluster, replicationPassword string) map[string]string {
 	baseConfig := buildManagedConfig(true, cluster.Spec.Persistence, cluster.Spec.TLS)
 	maps.Copy(baseConfig, map[string]string{
 		"cluster-enabled":                 "yes",
 		"protected-mode":                  "no",
 		"cluster-node-timeout":            "2000",
 		"cluster-allow-replica-migration": "no",
 		"cluster-replica-validity-factor": "0",
+		"primaryuser":                     replicationUser,
+		"primaryauth":                     replicationPassword,
```
jdheyburn: Is there a way we can avoid having the password in cleartext in the ConfigMap?

Perhaps making `replicationPassword` available as an env var, then passing this as an argument to the container (untested):

```
			Command: []string{
				"valkey-server",
				"/config/valkey.conf",
				"--cluster-announce-ip",
				"$(POD_IP)",
				"--primaryauth",
				"$(PRIMARY_AUTH)",
			},
```

hieu2102:
the e2e tests failed because the system users password secret is not created (same as #235 )

internal/controller/valkeynode_resources.go
```diff
@@ -182,6 +186,17 @@ func buildContainersDef(node *valkeyiov1alpha1.ValkeyNode) ([]corev1.Container,
...
+					Name: "PRIMARY_AUTH",
+					ValueFrom: &corev1.EnvVarSource{
+						SecretKeyRef: &corev1.SecretKeySelector{
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: getSystemPasswordSecretName(node.Labels[LabelCluster]),
+							},
+							Key: replicationUser,
+						},
+					},
+				},
```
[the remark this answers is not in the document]
hieu2102: I think this can be ignore, since we only create standalone ValkeyNode for testing anyway

internal/controller/users.go
```diff
@@ -357,39 +381,81 @@ func generatePassword(length int) ([]byte, error) {
...
 	}
 
-	err := apiClient.Create(ctx, &systemUsersSecret)
-	return &systemUsersSecret, err
+	var err error
+	if createSecret {
+		err = apiClient.Create(ctx, systemUsersSecret)
+	} else {
+		err = apiClient.Update(ctx, systemUsersSecret)
+	}
```
jdheyburn, line 437: Could we use `controllerutil.CreateOrUpdate` here? We're using that in a few places already.

internal/controller/users_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"strings"
 	"testing"
 
+	. "github.com/onsi/gomega"
```
jdheyburn: Is this used in this file?

internal/controller/users.go
```diff
@@ -357,39 +381,81 @@ func generatePassword(length int) ([]byte, error) {
 	return ret, nil
 }
 
+func validateSystemUserPasswordSecret(data map[string][]byte, cluster *valkeyiov1alpha1.ValkeyCluster) error {
```
jdheyburn, line 360: Can you comment what we are validating for here please?

internal/controller/users.go
```diff
@@ -357,39 +381,81 @@ func generatePassword(length int) ([]byte, error) {
...
 
-	err := apiClient.Create(ctx, &systemUsersSecret)
-	return &systemUsersSecret, err
+	var err error
+	if createSecret {
+		err = apiClient.Create(ctx, systemUsersSecret)
+	} else {
+		err = apiClient.Update(ctx, systemUsersSecret)
+	}
+	return systemUsersSecret, err
```
jdheyburn, line 438: Should this be `&systemUsersSecret` if the return type is `*corev1.Secret`?

hieu2102: `systemUsersSecret` is created as a pointer at the beginning of the function (L408)
```
	systemUsersSecret := &corev1.Secret{}
```

internal/controller/users_test.go
hieu2102: I use it for `TestValidateSystemUserPasswordSecret`. Can remove if you think it's not needed

internal/controller/users.go
hieu2102: then that would rotate the password of `_operator` too, should we do that?

internal/controller/users.go
jdheyburn: Oh great, my bad

internal/controller/users.go
jdheyburn: Why would it? Is `apiClient.Update` any different here?

hieu2102: Since the passwords are randomly generated, each time the function run it will create a different one for the system users. 
If I use `controllerutil.CreateOrUpdate` here, I'll have to check whether the secret exists/the secret is missing entry first if we don't want to change the passwords for existing users. Kind of defeat the purpose of using that function IMO

internal/controller/valkeynode_resources.go
```diff
@@ -172,6 +172,10 @@ func buildContainersDef(node *valkeyiov1alpha1.ValkeyNode) ([]corev1.Container,
 				"/config/valkey.conf",
 				"--cluster-announce-ip",
 				"$(POD_IP)",
+				"--primaryuser",
+				replicationUser,
```
sandeepkunusoth, line 176: dont we need to add replication only when replication is enabled and not always? same applies to replication user?

hieu2102: We currently only support cluster so I think it's okay if we just add replication configs for simplicity. 

Make sense to remove it when deploying standalone instances though. 

@jdheyburn WDYT?

sandeepkunusoth: what if we have cluster only master shards without replica. do we need replication there?

internal/controller/users.go
```diff
@@ -357,39 +384,84 @@ func generatePassword(length int) ([]byte, error) {
...
+
+	if err := r.Get(ctx, types.NamespacedName{
+		Name:      secretName,
+		Namespace: cluster.Namespace,
+	}, systemUsersSecret); err != nil {
+		if !apierrors.IsNotFound(err) {
+			log.Error(err, "failed to fetch system users secret")
+			return nil, err
+		}
+		log.V(2).Info("creating internal secret", "secretName", systemUsersSecret)
```
sandeepkunusoth: can we change systemUsersSecret to secretName. we dont need to log secret.

internal/controller/config_test.go
```diff
@@ -44,8 +44,6 @@ var _ = Describe("When creating a cluster", Label("userconfig"), func() {
 		// Check user-added parameter
 		Expect(testConfigString).To(ContainSubstring("maxmemory-policy"))
 
-		// Base, non-overridable parameter
-		Expect(testConfigString).To(ContainSubstring("cluster-enabled"))
```
sandeepkunusoth, line 48: any reason this is removed?

internal/controller/valkeynode_resources.go
hieu2102: then when admins add replica, the primary nodes will have to be recreated too, unless failover never happens

internal/controller/config_test.go
hieu2102: removed because my first implementation is adding replica configs to valkey.conf; forgot to add it back.

internal/controller/users_test.go
```diff
@@ -115,3 +117,20 @@ func TestBuildAclFileContents(t *testing.T) {
...
+	}
+	cluster := &valkeyiov1alpha1.ValkeyCluster{
+		Spec: valkeyiov1alpha1.ValkeyClusterSpec{
+			Exporter: valkeyiov1alpha1.ExporterSpec{
+				Enabled: true,
+			},
+		},
+	}
+	err := validateSystemUserPasswordSecret(userSecret, cluster)
+	g.Expect(err).To(MatchError(errMissingSystemUser))
```
bjosv: Using `testing` as in the other testcase works as well here, like the prod. code.
if !errors.Is(err, errMissingSystemUser) {
   t.Errorf(..)
}

bjosv (dismissed):
LGTM, just a test nit

internal/controller/users.go
```diff
@@ -357,39 +384,84 @@ func generatePassword(length int) ([]byte, error) {
...
+// and ensures all required system users are present.
+func validateSystemUserPasswordSecret(data map[string][]byte, cluster *valkeyiov1alpha1.ValkeyCluster) error {
+	// list of users need to exists in the secret (with _exporter being optional)
+	u := make([]string, len(systemUsers))
+	copy(u, systemUsers)
+	for user := range data {
+		i := slices.Index(u, user)
+		if i == -1 {
+			return fmt.Errorf("%w: %s", errUnknownSystemUser, user)
+		}
```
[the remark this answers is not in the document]
jdheyburn: I think this seems like a valid comment. Could we have it log about an unknown key @hieu2102 ?

internal/controller/valkeynode_resources.go
jdheyburn: I think leave this in, and raise another issue for an enhancement to only create replication user when it is required.

jdheyburn:
@hieu2102 Are you able to rebase please? It would be great to get this in 0.3.0 next week

internal/controller/users_test.go
```diff
@@ -17,10 +17,13 @@ limitations under the License.
 package controller
 
 import (
+	"errors"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	//	corev1 "k8s.io/api/core/v1"
```
bjosv: Seems like its not needed, remove it?

internal/controller/config_test.go
```diff
@@ -49,6 +49,7 @@ var _ = Describe("When creating a cluster", Label("userconfig"), func() {
 
 		// Primary fails over to a replica on SIGTERM (graceful shutdown)
 		Expect(testConfigString).To(ContainSubstring("shutdown-on-sigterm failover"))
+
```
bjosv: nit: remove

bjosv (commented):
2 quick nits

bjosv (approved):

jdheyburn (approved):
Thank you!

PR #238, opened 2026-06-09
Filter Secrets by managed-by label to reduce cache memory usage

### LINKED ISSUE #187: Filter out Secrets and ConfigMaps from Watches
From #176 

CloudNative-PG disables caching for sensitive data like secrets:

- https://github.com/cloudnative-pg/cloudnative-pg/blob/c9ea3aac03a951ee510849b9e333f8b560692714/internal/cmd/manager/instance/run/cmd.go#L207

We can help to reduce memory use by removing caching for Secrets and ConfigMaps.

bjosv:
This PR closes #187

### Summary

Only cache operator-owned Secrets (those with the`app.kubernetes.io/managed-by=valkey-operator` label).
User-provided Secrets (passwords, TLS certificates) are now read directly from the API server via `APIReader`.

Remove the `Watches()` for user-provided Secrets. The watch predicate filtered on `type=valkey.io/acl` which only matched operator-created Secrets (already covered by `Owns()`), so it had no effect on user-provided Secrets.

Widen `getTLSConfig` and `fetchUserPasswords` to accept `client.Reader` so they can be called with `APIReader` instead of the cached client.

### Checklist

### DIFF
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -32,7 +32,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	valkeyiov1alpha1 "valkey.io/valkey-operator/api/v1alpha1"
 )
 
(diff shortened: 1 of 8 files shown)


internal/controller/users.go
```diff
@@ -81,49 +80,6 @@ func getSystemPasswordSecretName(clusterName string) string {
...
-						Namespace: cluster.Namespace,
-					},
-				})
-			}
-		}
-	}
-
-	return requests
-}
-
```
jdheyburn, line 126: In my ignorance, why does this get removed here?

bjosv: This function is only used by Watches(), which I now removed.
This hasn't worked since adding the predicate in https://github.com/valkey-io/valkey-operator/commit/6b6047f0.
A change in a PasswordSecret does not trigger a reconcile for the cluster using it.

If we want to fix this instead then we need the Watch on Secrets, and they will be cached..

With this PR we wont immediately trigger a reconcile when a user-provided secret is modified (current faulty? behavior anyway), but first at next reconcile triggered by something else, or max 60s.

jdheyburn: Ok thanks for explaining. I _think_ we can wait until the next reconciliation as a suitable trade-off for this.

jdheyburn (approved):

PR #242, opened 2026-06-10
(feat) support Mutual TLS (mTLS) certificate-based ACL authentication

sandeepkunusoth:
This PR closes (https://github.com/valkey-io/valkey-operator/issues/243)

### Summary

This PR extends the `spec.networking.tls` configuration options for `ValkeyCluster` resources to support client certificate enforcement (mTLS) and automatic user mapping based on the client certificate's Common Name (CN).

### Features 
 - added mTLS client auth support

### Implementation

- Added two optional fields to `TLSConfig` in `api/v1alpha1/valkeycluster_types.go` backed by validated string enums:
  - `authClients`: Optional / Required / Disabled (Defaults to Optional). Maps to Valkey directive tls-auth-clients.
  - `authClientsUser`: `CN`/ `DNS`/ `Disabled` (Defaults to `Disabled`). Maps to Valkey directive `tls-auth-clients-user`.

### Testing

- Added an e2e test case inside `test/e2e/valkeycluster_tls_test.go` running a scenario that validates successful mTLS connections and tests using client certificate.

### Checklist


melancholictheory:
nice, the two-enum shape (authClients + authClientsUser) maps cleanly onto the valkey directives. one thing worth calling out in the design, since it bit us: turning on authClients: Required with authClientsUser: CN authenticates every client connection, including the operator's own and a couple of internal ones, so each of those needs a cert whose CN maps to an ACL user with the right grants. concretely:

- the operator's control-plane connections (CLUSTER NODES/SLOTS, failover, INFO via valkey-go) now need a client cert mapped to a user with the operator ACL (the _operator user from #136), otherwise the operator locks itself out of the cluster it's managing.
- the readiness/liveness probes shell out to valkey-cli, so they need --cert/--key/--cacert and a CN that maps to a user allowed to PING / CLUSTER INFO. this overlaps with #232/#235 (probes failing under auth), mTLS Required is another path into the same hole.
- replica to primary replication links are clients too. with CN mapping the replica's cert CN has to map to a user with replication grants, which lines up with the dedicated replication user in #236.

so the feature is really "every internal connection becomes authenticated", and the CRD probably wants the operator/probe/replication certs provisioned and CN-mapped before Required can be safely set, otherwise enabling it bricks the cluster it's managing.

two smaller notes:
- tls-auth-clients-user is valkey 9.0+, so authClientsUser: CN should be version-gated (rendered only on >= 9.0), since older servers reject the unknown directive on boot.
- CN to user mapping is sensitive to cert rotation: cert-manager renewals have to preserve the CN, or the mapping breaks and the client drops to the default user on renew. worth a line in docs/tls.md.

happy to share how we wired the operator's own client cert if useful.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -144,6 +174,22 @@ type TLSConfig struct {
...
+	AuthClients TLSAuthClients `json:"authClients,omitempty"`
+
+	// AuthClientsUser configures how Valkey maps an authenticated client
+	// certificate to an ACL user. Set to `CN` to authenticate clients as
+	// the ACL user matching the certificate's Common Name. Defaults to
+	// `Off` (no automatic mapping). Requires Valkey >= 9.0.0.
+	// +kubebuilder:default=Off
+	// +optional
+	AuthClientsUser TLSAuthClientsUser `json:"authClientsUser,omitempty"`
 }
```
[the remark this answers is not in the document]
sandeepkunusoth: fixed

sandeepkunusoth:
Sorry for delay in response.

> tls-auth-clients-user is valkey 9.0+, so authClientsUser: CN should be version-gated (rendered only on >= 9.0), since older servers reject the unknown directive on boot.

The operator currently only supports Valkey **9.0+**, so `authClientsUser` isn't introducing a new compatibility issue today. That said, we will be adding explicit version gating as part of a separate follow-up for the upcoming **9.1.0+ features**.

> **coming to operator users, health checks, and replication**

i fixed this by reusing valkey tls server certs  on clients side for operator user, redis exporter. i think we need to check more on if we want to move operator user to use mtls client cert as part of seperate Issue.  Health checks already pass client certs. updated e2e tests to test scenario with TLSAuthClients `yes` scenario and `tls-auth-clients-user: yes`. added point to docs/tls.md.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -151,6 +151,41 @@ type ValkeyClusterSpec struct {
...
+	// ACL username. Pair with `AuthClients: Required` to enforce mTLS.
+	TLSAuthClientsUserCN TLSAuthClientsUser = "CN"
+	// TLSAuthClientsUserURI maps the first URI from the certificate's
+	// Subject Alternative Name (SAN) that natches valkey ACL username.
+	TLSAuthClientsUserURI TLSAuthClientsUser = "URI"
+	// TLSAuthClientsUserOff disables certificate-to-user mapping (default).
+	TLSAuthClientsUserOff TLSAuthClientsUser = "off"
+)
+
+// +kubebuilder:validation:XValidation:rule="!(self.authClients == 'no' && self.authClientsUser != 'off')",message="authClientsUser=CN/URI has no effect when authClients=no (Valkey ignores client certificates)"
```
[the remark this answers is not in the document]
sandeepkunusoth: this is default behaviour from valkey. users are expected not to set nopass for only mtls client certificate based authentication. updated docs.

docs/mtls.md
```diff
@@ -0,0 +1,138 @@
+# Mutual TLS (mTLS) certificate-based ACL authentication
```
bjosv, line 1: We have TLS section in the [valkeycluster.md](https://github.com/valkey-io/valkey-operator/blob/main/docs/valkeycluster.md#tls) docs, we should add a link to this new file.
Maybe have this file a mTLS only doc and rename it?

docs/tls.md
```diff
@@ -0,0 +1,138 @@
+# Mutual TLS (mTLS) certificate-based ACL authentication
+
+`spec.tls.authClients` and `spec.tls.authClientsUser` extend the existing `spec.tls` feature so that:
+
+1. Clients can be required to present a TLS certificate (mTLS), and
+2. Authenticated clients can be automatically logged in as a Valkey ACL user matching the certificate's Common Name (CN) or URI SAN.
+
+> **Requires Valkey >= 9.0.0** for `authClientsUser: CN`/`URI`. The `tls-auth-clients-user` directive landed in Valkey 9.0.0. The default operator image (`valkey/valkey:9.0.0`) is sufficient.
```
bjosv: remove `valkey/valkey:9.0.0` if/when its changed?

docs/tls.md
```diff
@@ -0,0 +1,138 @@
...
+  users:
+    - name: alice
+      enabled: true
+      resetpass: true
+      permissions: "+@all ~app:* &events:*"
+```
+
+With `authClients: yes` + `authClientsUser: CN`, any TLS client whose certificate has `CN=alice` is automatically authenticated as the ACL user `alice` -- no `AUTH` command required. Pass `resetpass: true` with this configuration so authentication relies exclusively on the client certificate.
+
+When `authClients` is left at its default `yes`, Valkey clients requires any valid client certificate at the TLS handshake, but that does not disable password-based ACL authentication. Clients can still authenticate with `AUTH` as long as they present a client certificate signed by the configured CA. Today operator user, health check probes, redis exporter uses server certs to validate clients when `authClients: yes` is set.
```
bjosv: nit: I guess its `authClients: "yes"` so the yaml parser wont fail? In `authClients: yes` I think it's seen as a bool

bjosv (commented):
Just a quick look as we will have this under `spec.networking.tls`, but avoiding yes/no and using Required/Optional/Disabled as discussed in the issue has the additional benefit of minimizing the yaml-parser annoyance (bool vs string).

docs/mtls.md
```diff
@@ -0,0 +1,144 @@
...
+2. Authenticated clients can be automatically logged in as a Valkey ACL user matching the certificate's Common Name (CN) or URI SAN.
+
+> **Requires Valkey >= 9.0.0** for `authClientsUser: CN`/`URI`.
+
+## Valkey defaults vs operator defaults for mTLS
+
+By default, Valkey uses mutual TLS and requires clients to present a valid certificate verified against trusted root CAs configured via `tls-ca-cert-file` or `tls-ca-cert-dir`. You may use `tls-auth-clients no` to disable client authentication.
+
+When `spec.tls.authClients` is omitted, the operator defaults it to `Optional` and renders `tls-auth-clients optional` so TLS clients can connect without presenting a client certificate. Set `authClients: Required` to enforce mTLS (`tls-auth-clients yes`), or `authClients: Disabled` to turn client certificate processing off (`tls-auth-clients no`).
+
```
sandeepkunusoth, line 15: should we set default value of "tls-auth-clients"] to "Required" inside operator or stick with current "Optional"? when spec.tlsAuthClients is empty in spec as by default valkey 9.* uses tls-auth-clients to yes

jdheyburn:
@sandeepkunusoth Can you help me understand how this ties in with the wider TLS API design to be able to support additional TLS configurations like split-TLS configurations (see this [comment](https://github.com/valkey-io/valkey-operator/issues/318#issuecomment-5097232164) too).

docs/mtls.md
```diff
@@ -0,0 +1,144 @@
...
+```yaml
+apiVersion: valkey.io/v1alpha1
+kind: ValkeyCluster
+metadata:
+  name: valkeycluster-mtls
+spec:
+  shards: 3
+  replicas: 1
+  tls:
```
sandeepkunusoth: will update docs to reflect new spec

docs/valkeycluster.md
```diff
@@ -386,7 +386,7 @@ Cluster-owned StatefulSets use the **cluster headless Service** as `spec.service
...
 
-`networking.clusterDomain` must match the kubelet cluster domain (`--cluster-domain`). The field is a DNS subdomain (a trailing dot on input is trimmed). Hostname announce uses an FQDN **without** a trailing dot, so it is a valid TLS SNI name (RFC 6066) and matches typical cert SANs and the default TLS `serverName`. The per-pod announce name has at least five dots, so under the pod default `ndots:5` resolvers try it as absolute first (one query); only short custom cluster domains fall back to the search list, which is accepted over breaking SNI.
+`networking.clusterDomain` must match the kubelet cluster domain (`--cluster-domain`). The field is a DNS subdomain (optional trailing dot). Hostname announce uses an absolute FQDN (trailing dot) so resolvers do not append search domains. The default TLS `serverName` is that Service DNS name without the trailing dot.
```
daanvinken: Seems your rebase is reverting #425

internal/controller/metrics_exporter.go
```diff
@@ -54,10 +54,20 @@ func generateMetricsExporterContainerDef(exporter valkeyiov1alpha1.ExporterSpec,
 	var volumeMounts []corev1.VolumeMount
 
 	if tlsSpec != nil {
-		env = append(env, corev1.EnvVar{
-			Name:  "REDIS_EXPORTER_TLS_CA_CERT_FILE",
-			Value: fmt.Sprintf("%s/%s", tlsCertMountPath, tlsSecretKeyCA),
-		})
+		env = append(env,
```
daanvinken: These exporter env vars always point at the server cert/key whenever TLS is on.

`authClients: Optional` still verifies a client cert if one is presented. Always sending the server identity is not free: with `authClientsUser: CN` Valkey will try to map that server CN to an ACL user.

Same in `getTLSConfig`: every operator dial sets `tls.Config.Certificates` from the server secret, including when `authClients` is `Optional`.

IIUC we should present a client cert only when we mean to (Required, or an explicit operator/exporter identity).. May belong on `certificates.operator` from #360, not on the server slot.

api/v1alpha1/valkeynode_types.go
```diff
@@ -172,6 +173,24 @@ type NodeTLSSpec struct {
...
+	AuthClients TLSAuthClients `json:"authClients,omitempty"`
+
+	// AuthClientsUser configures how Valkey maps an authenticated client
+	// certificate to an ACL user. Set to `CN` to use the certificate's Common
+	// Name (requires Valkey >= 9.0), or `URI` to use the first matching URI from the certificate's Subject
+	// Alternative Name (SAN) (requires Valkey >= 9.1).
+	// Defaults to `Disabled`, which leaves the directive unset.
+	// +kubebuilder:default=Disabled
+	// +optional
+	AuthClientsUser TLSAuthClientsUser `json:"authClientsUser,omitempty"`
```
daanvinken: Cluster `TLSSpec` rejects `authClients: Disabled` with `authClientsUser: CN`/`URI`. `NodeTLSSpec` copies both fields and has no matching CEL.

A standalone ValkeyNode can still set that combo. Not sure what's best here.  We can add the same CEL here (and on the ValkeyNode CRD), or treat these as cluster-owned and resolved-only on the node so users cannot set them on `ValkeyNode` at all.

daanvinken (approved):
Shape looking good! Feel free to ping me for more in-depth review/testing when we resolve some comments 👍

api/v1alpha1/valkeycluster_types.go
```diff
@@ -454,6 +521,24 @@ type TLSSpec struct {
...
 	Certificates TLSCertificates `json:"certificates"`
+
+	// AuthClients controls whether clients must authenticate with a TLS
+	// certificate. `Required` enforces mTLS, `Optional` allows both authenticated
+	// and unauthenticated clients, and `Disabled` turns client certificate
+	// processing off entirely.
+	// Defaults to `Optional`.
+	// +kubebuilder:default=Optional
+	// +optional
+	AuthClients TLSAuthClients `json:"authClients,omitempty"`
```
bjosv: So we will rename this to `mode` and move it under a new `clientAuth` in a separate PR? (as in #360)

internal/controller/config.go
```diff
@@ -54,6 +54,14 @@ var versionGatedConfig = map[string]*semver.Version{
 	"tls-auto-reload-interval": semver.MustParse("9.1.0-rc1"),
 }
 
+// tlsAuthClientsUserMinVersion lists client certificate-to-user mappings that
+// need a newer Valkey release. Values absent from this map have no additional
+// version requirement.
+// TODO make this generic enough to fetch from enum values https://github.com/valkey-io/valkey-operator/issues/423
```
sandeepkunusoth, line 60: i can remove config warnings changes from this PR if its complicating. will need to do some work on fixing version gating for enum values before. included warning logic for now and also added todo comment for enum values fix.

api/v1alpha1/valkeynode_types.go
sandeepkunusoth: removed CEL on valkey node for now.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -440,6 +440,111 @@ type DiscoverySpec struct {
...
+	TLSAuthClientsRequired TLSAuthClients = "Required"
+	// TLSAuthClientsDisabled disables client certificate processing entirely.
+	// Maps to `tls-auth-clients no`.
+	TLSAuthClientsDisabled TLSAuthClients = "Disabled"
+)
+
+// tlsAuthClientsDirective maps the valkey CRD API spec field onto the values the
+// `tls-auth-clients` directive accepts.
+var tlsAuthClientsDirective = map[TLSAuthClients]string{
+	TLSAuthClientsRequired: "yes",
```
bjosv, line 463: We should probably not have valkey specific directives in the api since it becomes tightly connected. I'm not sure how we would handle version specific flags here in the future.

We can move `tlsAuthClientsDirective` and `tlsAuthClientsUserDirective` to `internal/controller/config.go` and keep the transformation internal, then we don't need to update the api if Valkey decides to rename fields (very unlikely but still).
I believe we use a similar split for PreferredEndpointType.

api/v1alpha1/valkeynode_types.go
```diff
@@ -172,6 +172,34 @@ type NodeTLSSpec struct {
 	// Certificates holds the certificate slots mounted into the node pod.
 	// +kubebuilder:validation:Required
 	Certificates NodeTLSCertificates `json:"certificates"`
+
+	// ClientAuth is the resolved client certificate authentication settings
+	// rendered into `tls-auth-clients` and `tls-auth-clients-user`. For
+	// cluster-owned nodes the ValkeyCluster controller copies
+	// spec.networking.tls.clientAuth here.
+	// +optional
+	ClientAuth *TLSClientAuthSpec `json:"clientAuth,omitempty"`
```
bjosv, line 181: We have validation rules on ValkeyNode for other elements, even when set by the cluster controller. Should we copy the [validation rules](https://github.com/valkey-io/valkey-operator/pull/242/changes#diff-d9af720dc936d3a24981d6705b64a1e7016e80089d2bb5cf57daf4d99c1a923eR511) for ClientAuth as well?

sandeepkunusoth: hi i didnt get this comment we already have validation present on same struct TLSClientAuthSpec type from api/v1alpha1/valkeycluster_types.go i am using same object in both api/v1alpha1/valkeycluster_types.go and api/v1alpha1/valkeynode_types.go

```
// TLSClientAuthSpec configures client certificate authentication for incoming
// TLS connections.
// +kubebuilder:validation:XValidation:rule="!(has(self.mode) && self.mode == 'Disabled' && has(self.certificateUser) && self.certificateUser != 'Disabled')",message="certificateUser has no effect when mode=Disabled: Valkey ignores client certificates in that mode"
type TLSClientAuthSpec struct {
	// Mode controls whether clients must authenticate with a TLS certificate.
	// `Required` enforces mTLS, `Optional` allows both authenticated and
	// unauthenticated clients, and `Disabled` turns client certificate processing
	// off entirely.
	// Defaults to `Optional`.
	// +kubebuilder:default=Optional
	// +optional
	Mode TLSAuthClients `json:"mode,omitempty"`

	// CertificateUser configures how Valkey maps an authenticated client
	// certificate to an ACL user. Set to `CN` to use the certificate's Common Name
	// (requires Valkey >= 9.0), or `URI` to use the first matching URI from the
	// certificate's Subject Alternative Name (SAN) (requires Valkey >= 9.1).
	// Defaults to `Disabled`, which leaves the directive unset.
	// +kubebuilder:default=Disabled
	// +optional
	CertificateUser TLSAuthClientsUser `json:"certificateUser,omitempty"`
}
```

api/v1alpha1/valkeycluster_types.go
sandeepkunusoth: moved tlsAuthClientsDirective and tlsAuthClientsUserDirective to internal/controller/config.go

api/v1alpha1/valkeynode_types.go
bjosv: oh, you're right. Not sure why I commented on that..

bjosv (approved):
LGTM!

PR #244, opened 2026-06-11
fix: recover cluster when majority of primaries are lost

### LINKED ISSUE #240: [bug] Operator cannot recover cluster when majority of primaries are lost
When a majority of shard primaries are down, the operator gets stuck in `Reconciling` / `SlotsUnassigned` indefinitely. It refuses to forget the dead primaries because live replicas still reference them.

**Example:**
A 3-shard cluster with 1 replica each. Two primaries die, their replicas survive but cannot promote (quorum requires 2 of 3 masters to vote, only 1 is alive). 

**Possible fix:**
1. Forget dead primaries when failover quorum is unreachable
2. Issue CLUSTER FAILOVER TAKEOVER on surviving replicas to promote them (preserves data)
3. Reattach new replacement pods as replicas of the promoted nodes

bjosv:
This PR closes #240 

### Summary

When a majority of shard primaries are down, the operator was stuck indefinitely because it refused to forget dead primaries that still had live replicas referencing them, and automatic failover could never succeed without quorum.

Add `promoteOrphanedReplicas` which issues `CLUSTER FAILOVER TAKEOVER` to the best replica (highest replication offset) of each dead primary when quorum is unreachable. TAKEOVER runs before FORGET so slots remain continuously owned. With persistence enabled, pods return with the same node ID and the operator skips TAKEOVER to let them rejoin naturally.

Also bypasses the `HasReplicaOf` guard in `forgetStaleNodes` when quorum is lost, allowing dead primaries to be cleaned up from the gossip table.

### Testing

Found with #203 and can be reproduced with:

`CHAOS_SHARDS=3 CHAOS_REPLICAS=1 CHAOS_TARGET_SHARDS=0,1 CHAOS_SCENARIOS=delete-primary-pod make test-chaos`

3 shards 1 replica each, kills 2 primaries every iteration for quorum loss.

### Checklist

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -988,7 +988,7 @@ func (r *ValkeyClusterReconciler) replicateToShardPrimary(ctx context.Context, c
 // the pod will return with the same node ID, so TAKEOVER is skipped.
 // Returns (result, true) if a promotion was made and reconcile should requeue.
 func (r *ValkeyClusterReconciler) promoteOrphanedReplicas(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, state *valkey.ClusterState) (ctrl.Result, bool) {
-	if state.HasFailoverQuorum() || cluster.Spec.Persistence != nil {
+	if cluster.Spec.Persistence != nil || state.HasFailoverQuorum() {
 		return ctrl.Result{}, false
 	}
 	log := logf.FromContext(ctx)
@@ -1054,7 +1054,7 @@ func (r *ValkeyClusterReconciler) forgetStaleNodes(ctx context.Context, cluster 
 				// voting in the auto-failover election, permanently
 				// blocking the replica's promotion.
 				if state.HasReplicaOf(failing.Id) {
-					if state.HasFailoverQuorum() || cluster.Spec.Persistence != nil {
+					if cluster.Spec.Persistence != nil || state.HasFailoverQuorum() {
 						log.V(1).Info("skipping forget; failover pending for node",
 							"address", failing.Address, "Id", failing.Id)
 						continue
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -82,27 +82,31 @@ type ValkeyClusterReconciler struct {
 // cluster one step closer to the desired state described by the ValkeyCluster
 // spec. The pipeline runs in the following order:
 //
-//  1. Ensure the headless Service exists (upsertService).
-//  2. Ensure the ConfigMap with valkey.conf and health-check scripts exists
+//   - Ensure the headless Service exists (upsertService).
+//   - Ensure PodDisruptionBudget exists (reconcilePodDisruptionBudget).
+//   - Ensure internal ACL users are configured (reconcileUsersAcl).
+//   - Ensure the ConfigMap with valkey.conf and health-check scripts exists
 //     (upsertConfigMap).
-//  3. Ensure one ValkeyNode per (shard, node) pair exists, creating missing
+//   - Ensure one ValkeyNode per (shard, node) pair exists, creating missing
 //     nodes and propagating spec changes one at a time in shard order with
 //     replicas updated before the primary (reconcileValkeyNodes).
-//  4. List all ValkeyNodes and build the Valkey cluster state by connecting to each
-//     node and scraping CLUSTER INFO / CLUSTER NODES.
-//  5. Forget stale nodes that no longer have a backing ValkeyNode.
-//  6. Phase 1 – MEET: batch-introduce all isolated pending nodes to the
-//     cluster via CLUSTER MEET. Requeue to let gossip propagate.
-//  7. Phase 2 – Assign slots: batch-assign hash-slot ranges to all
-//     primary-labeled pending nodes via CLUSTER ADDSLOTSRANGE.
-//  8. Phase 3 – Replicate: batch-attach all replica-labeled pending nodes
-//     to their matching primaries via CLUSTER REPLICATE.
-//  9. Scale-in: if the cluster has more shards than desired, drain slots
+//   - Build the Valkey cluster state by connecting to each node and scraping
+//     CLUSTER INFO / CLUSTER NODES.
+//   - Promote orphaned replicas via CLUSTER FAILOVER TAKEOVER when quorum
+//     is lost (promoteOrphanedReplicas).
+//   - Forget stale nodes that no longer have a backing ValkeyNode.
+//   - MEET: batch-introduce all isolated pending nodes to the cluster via
+//     CLUSTER MEET. Requeue to let gossip propagate.
+//   - Assign slots: batch-assign hash-slot ranges to all primary-labeled
+//     pending nodes via CLUSTER ADDSLOTSRANGE.
+//   - Replicate: batch-attach all replica-labeled pending nodes to their
+//     matching primaries via CLUSTER REPLICATE.
+//   - Scale-in: if the cluster has more shards than desired, drain slots
 //     from excess shards via CLUSTER MIGRATESLOTS and delete their
 //     ValkeyNodes once fully drained.
-//  10. Verify that the expected number of shards and replicas exist.
-//  11. Verify that all 16384 hash slots are assigned.
-//  12. If everything is healthy, mark the cluster Ready and requeue after 30s
+//   - Verify that the expected number of shards and replicas exist.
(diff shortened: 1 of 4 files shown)


internal/valkey/clusterstate.go
```diff
@@ -268,6 +268,74 @@ func (s *ClusterState) HasReplicaOf(nodeId string) bool {
...
+			// Take the max across nodes in case gossip hasn't fully propagated.
+			if size, err := strconv.Atoi(node.ClusterInfo["cluster_size"]); err == nil && size > clusterSize {
+				clusterSize = size
+			}
+		}
+	}
+	if clusterSize == 0 {
+		clusterSize = livePrimaries
+	}
+	return livePrimaries > clusterSize/2
```
jdheyburn: Could we use parenthesis to help with readability? Assuming it is `livePrimaries > (clusterSize / 2)`

internal/valkey/clusterstate.go
```diff
@@ -268,6 +268,74 @@ func (s *ClusterState) HasReplicaOf(nodeId string) bool {
...
+		return false
+	}
+	var livePrimaries, clusterSize int
+	for _, shard := range s.Shards {
+		if shard.GetPrimaryNode() != nil && len(shard.Slots) > 0 {
+			livePrimaries++
+		}
+		for _, node := range shard.Nodes {
+			// Take the max across nodes in case gossip hasn't fully propagated.
+			if size, err := strconv.Atoi(node.ClusterInfo["cluster_size"]); err == nil && size > clusterSize {
```
jdheyburn, line 287: Any reason we get this from INFO as opposed to `spec.shards`?

internal/valkey/clusterstate.go
```diff
@@ -268,6 +268,74 @@ func (s *ClusterState) HasReplicaOf(nodeId string) bool {
...
+			}
+			offset, _ := strconv.ParseInt(node.Info["master_repl_offset"], 10, 64)
+			if best == nil || offset > bestOffset {
+				best = node
+				bestOffset = offset
+			}
+		}
+	}
+	return best
+}
```
jdheyburn, line 337: I like this, and think we want to reuse it for proactive failover replica selection.

internal/controller/valkeycluster_controller.go
```diff
@@ -973,6 +983,56 @@ func (r *ValkeyClusterReconciler) replicateToShardPrimary(ctx context.Context, c
 	return nil
 }
 
+// promoteOrphanedReplicas issues CLUSTER FAILOVER TAKEOVER to replicas whose
+// primary is dead and failover quorum is unreachable. With persistence enabled
+// the pod will return with the same node ID, so TAKEOVER is skipped.
+// Returns (result, true) if a promotion was made and reconcile should requeue.
+func (r *ValkeyClusterReconciler) promoteOrphanedReplicas(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, state *valkey.ClusterState) (ctrl.Result, bool) {
+	if state.HasFailoverQuorum() || cluster.Spec.Persistence != nil {
```
jdheyburn: Why does persistence enabled make a difference here? If the primary is down, wouldn't we still need to force a FAILOVER regardless?

jdheyburn:
@greptile-apps

internal/valkey/clusterstate.go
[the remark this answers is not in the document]
bjosv: Aimed for minimal impact, but reasonable. Will fix

bjosv: Fixed, thanks

internal/controller/valkeycluster_controller.go
```diff
@@ -973,6 +983,56 @@ func (r *ValkeyClusterReconciler) replicateToShardPrimary(ctx context.Context, c
...
+		} else {
+			promoted++
+		}
+	}
+	if promoted > 0 {
+		r.Recorder.Eventf(cluster, nil, corev1.EventTypeNormal, "ReplicasTakenOver", "FailoverTakeover", "Promoted %d orphaned replica(s) via TAKEOVER", promoted)
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, true
+	}
+	return ctrl.Result{}, false
+}
```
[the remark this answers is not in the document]
bjosv: Done

internal/controller/valkeycluster_controller.go
bjosv: My thinking was that with persistence a crashed/killed pod returns with the same node ID and data, then the takeover is unnecessary.
The edge case of a stuck pod (unschedulable, PVC issues) probably need a timeout or a pod-status check of some sort? Maybe I should add a issue about that as a followup?

internal/valkey/clusterstate.go
bjosv: This is whats available in the ClusterState, and also, cluster_size reflects the cluster's own view of how many slot-owning primaries exist.
(Then I don't think spec.shards match reality during scaling?)

internal/valkey/clusterstate.go
bjosv: Yes, I was thinking about that too. Here in this scenario we might have a bigger risk that they differ, but I could be needed there as well.

internal/valkey/clusterstate.go
```diff
@@ -268,6 +268,74 @@ func (s *ClusterState) HasReplicaOf(nodeId string) bool {
...
+// that references the given primary ID, or nil if none found.
+func (s *ClusterState) BestReplicaOf(primaryId string) *NodeState {
+	var best *NodeState
+	var bestOffset int64
+	for _, shard := range s.Shards {
+		for _, node := range shard.Nodes {
+			if node.PrimaryIdFromSelf() != primaryId {
+				continue
+			}
+			offset, _ := strconv.ParseInt(node.Info["master_repl_offset"], 10, 64)
```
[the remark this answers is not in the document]
bjosv: @greptile-apps hum..  on a replica `master_repl_offset` and `slave_repl_offset` seems to be kept in sync, both advance when a command is applied?

Verify if I'm right by checking this in the [valkey source code](https://github.com/valkey-io/valkey), see:
1. Open src/networking.c, function commandProcessed() (~line 3835)
2. Line 3842 updates c->repl_data->reploff -- this is server.primary->repl_data->reploff, printed as slave_repl_offset at src/server.c:6557
3. Lines 3851-3854: immediately after, calls replicationFeedStreamFromPrimaryStream() with the same byte count
4. Follow that into src/replication.c:671 --> it calls feedReplicationBuffer()
5. feedReplicationBuffer() at line 476 does server.primary_repl_offset += copy -- printed as master_repl_offset at src/server.c:6643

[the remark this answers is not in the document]
bjosv: I now see that `server.primary->repl_data->reploff` is used internally when a replica needs its offset (i.e. same as `slave_repl_offset`).
So, it might be more proper to use that even if they are synced, I will change.

internal/controller/valkeycluster_controller.go
jdheyburn: Hmm yeah you're right. I think @melancholictheory mentioned this case previously. Persistence will always get the best data guarantees, when we don't persist we need the operator to fix things.

As a small optimisation, we can short-circuit the if condition by putting the cheaper call on the left side of the OR statement:

```go
if cluster.Spec.Persistence != nil || state.HasFailoverQuorum() {
```

Then `state.HasFailoverQuorum()` would not be called if `cluster.Spec.Persistence != nil`.

I saw this line being done in a couple of places.

internal/valkey/clusterstate.go
jdheyburn: I mentioned it in #249 - I think we will get this PR in first then rebase that one to cover this.

internal/valkey/clusterstate.go
jdheyburn: Got it - thanks!

internal/controller/valkeycluster_controller.go
bjosv: I switched order now.
I was wondering if tracing the reconcile loop would be of any benefit (using OpenTelemetry) to see whats taking time in the loop. I cant see any existing operator using it internally, so maybe people just use counters and pprof to find performance issues/improvements..

jdheyburn: Tracing would definitely be great to have, but I think having metrics first would be ideal. When we want to triage high reconciliation times, at that point having tracing would be a good feature to implement to investigate.

jdheyburn (approved):
LGTM! I've not tested it, but looks like it does all the right things :shipit:

PR #245, opened 2026-06-11
test: add e2e test for operator permissions

### LINKED ISSUE #144: Add CI check to verify that the operator has appropriate permissions
The below PR limited the ACL permissions the _operator user could perform:

- https://github.com/valkey-io/valkey-operator/pull/136

As functionality is added/removed from the operator, we should ensure that this list is kept up-to-date.

A CI check to verify the permissions list is kept updated would help with this.

tkarger:
This PR closes #144 

### Summary

The `_operator` user needs specific permissions to operate. The PR #136 tightened the permissions. This PR includes 2 tests which test the permissions of the `_operator` user to ensure that the permissions don't get broken by future changes.

- Add a test case for permissions which should be allowed.
- Add a test case for permissions which shouldn't be allowed
- Update the e2e test `validating internal secrets were created` to check the secret of the system-passwords got recreated as well. This prevents a race condition because the e2e test will retry until the secret got recreated
- Update the CONTRIBUTING documentation to run specific e2e tests

#### verifying allowed commands succeed for operator user
The test will check the log for the string NOPERM. No executed command should result a NOPERM response.

```go
cmd = exec.Command("kubectl", "run", "client",
					fmt.Sprintf("--image=%s", valkeyClientImage), "--restart=Never",
					"--", "sh", "-c",
					fmt.Sprintf(
						`valkey-cli -c -h "%s" --user _operator --pass "%s" <<EOF
PING
CLUSTER INFO
CLUSTER MYID
INFO
EOF`,
						clusterFqdn,
						operatorPassword,
					),
				)
```
```go
g.Expect(output).NotTo(ContainSubstring("NOPERM"))
```

#### verifying denied commands fail for operator user

This test checks that every command result in a NOPERM response.
```go
disallowedCommands := []string{
					"SET foo bar",
					"GET foo",
					"DEL foo",
					"KEYS *",
					"CONFIG GET *"
					"CONFIG SET hz 15",
					"ACL LIST",
				}

				for _, command := range disallowedCommands {
					By(fmt.Sprintf("verifying operator user cannot run: %s", command))

					_ = exec.Command("kubectl", "delete", "pod", "client",
						"--ignore-not-found=true", "--wait=true", "--timeout=30s").Run()

					cmd = exec.Command("kubectl", "run", "client",
						fmt.Sprintf("--image=%s", valkeyClientImage), "--restart=Never",
						"--", "sh", "-c",
						fmt.Sprintf(
							`valkey-cli -c -h "%s" --user _operator --pass "%s" %s`,
							clusterFqdn, operatorPassword, command,
						),
					)
					_, err = utils.Run(cmd)
					Expect(err).NotTo(HaveOccurred())

					cmd = exec.Command("kubectl", "wait", "pod/client",
						"--for=jsonpath={.status.phase}=Succeeded", "--timeout=30s")
					_, err = utils.Run(cmd)
					Expect(err).NotTo(HaveOccurred())

					cmd = exec.Command("kubectl", "logs", "client")
					output, err := utils.Run(cmd)
					Expect(err).NotTo(HaveOccurred())

					_ = exec.Command("kubectl", "delete", "pod", "client",
						"--ignore-not-found=true", "--wait=true", "--timeout=30s").Run()

					Expect(output).To(ContainSubstring("NOPERM"),
						"expected command %q to be denied but got: %s", command, output)
				}
```

Happy to discuss!

Cheers,
Tim

### Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -185,12 +185,30 @@ valkey-operator/
 # Run unit tests
 make test
 
-# Run E2E tests (requires cluster)
-make test-e2e
-
 # Run specific tests
 go test ./internal/controller/... -v
 ```
+
+### E2E
+End-to-end tests are executed against pull requests and are required to pass in order for the PR to be mergeable.
+
+To run end-to-end tests locally, you can execute the below command.
+```bash
+# Run E2E tests (requires cluster)
+make test-e2e
+```
+
+To run specific e2e tests locally use the `Focus` decorator of the GinkGo library (see [docs](https://onsi.github.io/ginkgo/#focused-specs)). <br/>
+Select a test case e.g.
+```go
+It("creates a cluster with custom users", func() {...}
+```
+Add the `Focus` decorator
+```go
+It("creates a cluster with custom users", Focus, func() {...}
+```
+Run `make test-e2e`. It'll just run the selected test case.
+
 
 ## Making Changes to the API
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -198,7 +198,7 @@ To run end-to-end tests locally, you can execute the below command.
 make test-e2e
 ```
 
-To run specific e2e tests locally use the `Focus` decorator of the GinkGo library (see [docs](https://onsi.github.io/ginkgo/#focused-specs)). <br/>
+To run specific e2e tests locally use the `Focus` decorator of the Ginkgo library (see [docs](https://onsi.github.io/ginkgo/#focused-specs)). <br/>
 Select a test case e.g.
 ```go
 It("creates a cluster with custom users", func() {...}
--- a/test/e2e/valkeycluster_test.go
+++ b/test/e2e/valkeycluster_test.go
@@ -486,10 +486,14 @@ spec:
 			}
 			Eventually(verifyReady).Should(Succeed())
 
-			By("validating internal secret was created")
+			By("validating internal secrets were created")
 			verifyInternalSecretExists := func(g Gomega) {
 				cmd := exec.Command("kubectl", "get", "secrets", "internal-"+withUserClusterName+"-acl")
 				_, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				cmd = exec.Command("kubectl", "get", "secrets", "internal-"+withUserClusterName+"-system-passwords")
+				_, err = utils.Run(cmd)
 				g.Expect(err).NotTo(HaveOccurred())
 			}
 			Eventually(verifyInternalSecretExists).Should(Succeed())
--- a/test/e2e/valkeycluster_test.go
+++ b/test/e2e/valkeycluster_test.go
@@ -20,6 +20,7 @@ limitations under the License.
 package e2e
 
 import (
+	"encoding/base64"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


tkarger:
The test `verifying denied commands fail for operator user` is going to fail because the command `CONFIG SET hz 15` is not resulting in a `NOPERM` response. Is this behavior intended? IMO if the `_operator` user isn't allowed to run `CONFIG GET *`, it shouldn't be able to set the config by running `CONFIG SET hz 15`.

- [x] Check permission is set correctly or remove the command `CONFIG SET hz 15`

tkarger:
> The test `verifying denied commands fail for operator user` is going to fail because the command `CONFIG SET hz 15` is not resulting in a `NOPERM` response. Is this behavior intended? IMO if the `_operator` user isn't allowed to run `CONFIG GET *`, it shouldn't be able to set the config by running `CONFIG SET hz 15`.
> 
> * [ ]  Check permission is set correctly or remove the command `CONFIG SET hz 15`

@sandeepkunusoth @jdheyburn  Hi, could you please have a look at this?

CONTRIBUTING.md
```diff
@@ -185,13 +185,19 @@ valkey-operator/
...
 ```
 
+### E2E
+```bash
+# Run E2E tests (requires cluster)
+make test-e2e
+```
+
+To run specific e2e tests use the `Focus` decorator of the GinkGo library (see [docs](https://onsi.github.io/ginkgo/#focused-specs)). <br/>
+E.g. `It("just me please", Focus, func() { ... })`.
```
jdheyburn: Would you mind adding a code block of how we would do this?

CONTRIBUTING.md
```diff
@@ -185,13 +185,19 @@ valkey-operator/
...
 make test
 
-# Run E2E tests (requires cluster)
-make test-e2e
-
 # Run specific tests
 go test ./internal/controller/... -v
 ```
 
+### E2E
```
jdheyburn, line 192: Would you mind adding an intro to this? Something like:

```
End-to-end tests are executed against pull requests and are required to pass in order for the PR to be mergeable.

To run end-to-end tests locally, you can execute the below command.
```

test/e2e/valkeycluster_test.go
```diff
@@ -535,6 +539,116 @@ spec:
...
+				operatorPassword, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+
+				disallowedCommands := []string{
+					"SET foo bar",
+					"GET foo",
+					"DEL foo",
+					"KEYS *",
+					"CONFIG GET *",
+					//"CONFIG SET hz 15",
```
jdheyburn: Do we need this commented line?

jdheyburn (commented):
Thank you for raising this! I had some minor comments.

I apologise for not being clearer when I raised the issue, but what I had in mind was a way to parse the Go code for all the Valkey commands being executed, and ensuring that the ACL allows them. If there's another way to do this then I'm also keen for that too.

The problem is if the operator executes a new command and a contributor forgets to add it to the ACL, then these e2e tests you've added will still pass and the binary will error. A CI check that validates the ACL exists would help to prevent this.

test/e2e/valkeycluster_test.go
```diff
@@ -535,6 +539,116 @@ spec:
...
+				clusterFqdn := fmt.Sprintf("valkey-%s.default.svc.cluster.local", withUserClusterName)
+
+				_ = exec.Command("kubectl", "delete", "pod", "client",
+					"--ignore-not-found=true", "--wait=true", "--timeout=30s").Run()
+
+				cmd := exec.Command(
+					"sh", "-c",
+					fmt.Sprintf(`kubectl get secret internal-%s-system-passwords \
+    -n default \
+    -o jsonpath='{.data._operator}' | base64 -d`, withUserClusterName),
```
bjosv: An option to avoid pipes and `base64 -d` (GNU version only) is to return without it, and do the decoding in go 

decoded, err := base64.StdEncoding.DecodeString(b64Password)
g.Expect(err).NotTo(HaveOccurred())
operatorPassword := string(decoded)

test/e2e/valkeycluster_test.go
```diff
@@ -535,6 +539,116 @@ spec:
...
+					By(fmt.Sprintf("verifying operator user cannot run: %s", command))
+
+					_ = exec.Command("kubectl", "delete", "pod", "client",
+						"--ignore-not-found=true", "--wait=true", "--timeout=30s").Run()
+
+					cmd = exec.Command("kubectl", "run", "client",
+						fmt.Sprintf("--image=%s", valkeyClientImage), "--restart=Never",
+						"--", "sh", "-c",
+						fmt.Sprintf(
+							`valkey-cli -c -h "%s" --user _operator --pass "%s" %s`,
```
bjosv: We could loop over all commands here instead using a single pod, then check that we have as many NOPERM as elements in disallowedCommands. It should be faster if we avoid starting and stopping a pod multiple times.

bjosv (commented):
Some optional comments

tkarger:
> Thank you for raising this! I had some minor comments.
> 
> I apologise for not being clearer when I raised the issue, but what I had in mind was a way to parse the Go code for all the Valkey commands being executed, and ensuring that the ACL allows them. If there's another way to do this then I'm also keen for that too.
> 
> The problem is if the operator executes a new command and a contributor forgets to add it to the ACL, then these e2e tests you've added will still pass and the binary will error. A CI check that validates the ACL exists would help to prevent this.

Thanks for your comment & sorry for the late answer!

Do you mean something like a AST scan to detect calls of the builder pattern? We could create a cmd list by the AST scan and run them with `ACL DRYRUN _operator cmd` if they get a OK response.

A mapping of commands would be required e.g. `CLUSTERINFO` to `CLUSTER INFO`.
Then run all commands with the _operator user by `ACL DRYRUN _operator cmd` to check if they actually succeed.

Is that closer to what you had in mind @jdheyburn ?
Thanks!

test/e2e/valkeycluster_test.go
tkarger: No, I wanted to raise this issue [here](https://github.com/valkey-io/valkey-operator/pull/245#issuecomment-4680718860). IMO this cmd shouldn't succeed. Should I create another issue for this? I'll remove the line

test/e2e/valkeycluster_test.go
tkarger: @bjosv Thanks, you are right. It takes quite a time for all the pods to start and stop. My idea was to get a specific command noperm error instead of the whole test case failing. When we agreed to a general approach, I'll apply your suggestions.

jdheyburn:
@tkarger Apologies for the delay getting to this. Yeah that is what I had in mind since that would capture future cases where we've added a new required ACL but not permitted it.

We don't exclusively use the builder pattern, we'll want to capture .Arbitrary calls too, like in below:

- https://github.com/search?q=repo%3Avalkey-io%2Fvalkey-operator%20.Arbitrary&type=code

We may want to test for grouped commands too like [+@connection](https://github.com/valkey-io/valkey-operator/blob/a7fa7bb8c7751bcac442c1b1540a284876ac3652/internal/controller/users.go#L52), though this could be overkill.

The mapping sounds good to me. Though hyphenated commands would need to be considered too: `ClusterSetConfigEpoch -> CLUSTER SET-CONFIG-EPOCH`.

Having said that, I think having the e2e tests here are good. So pending the checks are green we can merge this and raise another PR for the above. How does that sound to you?

tkarger:
@jdheyburn Sounds good. I'll raise another PR for the AST scan topic.

For this PR, I applied the suggestions from @bjosv to avoid pipes and removed the loop of pods starting for every command to increase the performance. It now counts the amount of raised `NOPERM`. Finally I squashed the commits.

@bjosv @jdheyburn Thanks for the feedback on this PR!

sandeepkunusoth:
@greptile-apps

test/e2e/valkeycluster_test.go
```diff
@@ -549,12 +540,118 @@ spec:
...
+				cmd = exec.Command("kubectl", "get", "secrets", "internal-"+withUserClusterName+"-system-passwords",
+					"-o", "jsonpath={.data._operator}",
+				)
+
+				b64Password, err := utils.Run(cmd)
+				g.Expect(err).NotTo(HaveOccurred())
+				decoded, err := base64.StdEncoding.DecodeString(b64Password)
+				g.Expect(err).NotTo(HaveOccurred())
+				operatorPassword := string(decoded)
+
```
bjosv, line 557: The added "denied commands" test pre-deletes the client pod in case the test fails in one iteration, maybe needed here too? I think we use that pattern in other tests too.

CONTRIBUTING.md
```diff
@@ -185,12 +185,30 @@ valkey-operator/
...
+
+To run end-to-end tests locally, you can execute the below command.
+```bash
 # Run E2E tests (requires cluster)
 make test-e2e
+```
 
-# Run specific tests
-go test ./internal/controller/... -v
+To run specific e2e tests locally use the `Focus` decorator of the GinkGo library (see [docs](https://onsi.github.io/ginkgo/#focused-specs)). <br/>
```
bjosv: nit: lowercase seems to be use in https://github.com/onsi/ginkgo
```suggestion
To run specific e2e tests locally use the `Focus` decorator of the Ginkgo library (see [docs](https://onsi.github.io/ginkgo/#focused-specs)). <br/>
```

test/e2e/valkeycluster_test.go
```diff
@@ -549,12 +540,118 @@ spec:
...
+				cmd = exec.Command("kubectl", "run", "client",
+					fmt.Sprintf("--image=%s", valkeyClientImage), "--restart=Never",
+					"--", "sh", "-c",
+					fmt.Sprintf(
+						`valkey-cli -c -h "%s" --user _operator --pass "%s" <<EOF
+PING
+CLUSTER INFO
+CLUSTER MYID
+INFO
+EOF`,
```
bjosv, line 575: Hum, we have more commands allowed in
https://github.com/valkey-io/valkey-operator/blob/a7fa7bb8c7751bcac442c1b1540a284876ac3652/internal/controller/users.go#L67
but some might be destructive for the test.. we should be able to test "ROLE" I guess.
Possibly we can execute them without any arguments, then we might get ERR and not NOPERM.

We should add a comment about why we don't have all commands here otherwise.

bjosv (approved):
Thanks!

jdheyburn (approved):
Thank you! Would you mind raising another issue for the additional work mentioned in this comment?

- https://github.com/valkey-io/valkey-operator/pull/245#issuecomment-5107972380

PR #249, opened 2026-06-12
feat: fail over to the highest-offset replica

melancholictheory:
### Summary

When the operator proactively fails a primary over before rolling it, it currently promotes the first synced replica in discovery order, which can be the one furthest behind.

A graceful `CLUSTER FAILOVER` holds writes on the primary until the target replica catches up, so promoting the most caught-up replica shortens that write pause and narrows the exposure if the primary dies mid-failover. This is improvement #1 from the discussion in #231.

- Proactive failover picks the synced replica with the highest replication offset instead of the first one in discovery order.

### Implementation

- `(*NodeState).ReplicationOffset()` reads `slave_repl_offset` from the node's `INFO replication`, which `getNodeState` already collects into `node.Info`, so there are no extra queries.
- `highestOffsetReplica` chooses the furthest-ahead replica. Replicas with no available offset sort last and ties keep discovery order, so the result is stable and never nil for a non-empty input.

### Limitations

- It does not yet assert that the chosen replica is on the latest ValkeyNode spec (improvement #2 in #231). The sequential roll already promotes replicas before the primary, so in practice the target is on the new spec, but an explicit check would turn that ordering invariant into a hard guarantee. Happy to follow up with it.

### Testing

- Unit tests for `highestOffsetReplica`: greatest offset wins, a missing offset sorts last, ties keep discovery order, single replica.

### Checklist

### DIFF
--- a/internal/controller/failover.go
+++ b/internal/controller/failover.go
@@ -60,21 +60,6 @@ func findFailoverShard(state *valkey.ClusterState, address string) (*valkey.Shar
 	return shard, replicas
 }
 
-// highestOffsetReplica returns the synced replica with the greatest replication
-// offset, the one most caught up with the primary. Replicas whose offset is
-// unavailable sort last and ties keep discovery order, so the result is stable
-// and never nil for a non-empty slice.
-func highestOffsetReplica(replicas []*valkey.NodeState) *valkey.NodeState {
-	best := replicas[0]
-	bestOffset := best.ReplicationOffset()
-	for _, replica := range replicas[1:] {
-		if offset := replica.ReplicationOffset(); offset > bestOffset {
-			best, bestOffset = replica, offset
-		}
-	}
-	return best
-}
-
 // proactiveFailover issues CLUSTER FAILOVER to the best synced replica in shard,
 // then polls until the replica reports role:master or the timeout is reached.
 // shard must be non-nil; replicas must be non-empty.
@@ -86,7 +71,7 @@ func proactiveFailover(ctx context.Context, recorder events.EventRecorder, clust
 	// graceful CLUSTER FAILOVER holds writes on the primary until the target
 	// replica catches up, so promoting the furthest-ahead one minimises that
 	// write pause and the exposure if the primary dies mid-failover.
-	target := highestOffsetReplica(replicas)
+	target := valkey.HighestOffsetReplica(replicas)
 	log.Info("initiating proactive failover", "shard", shard.Id, "target", target.Address)
 
 	// Emit FailoverInitiated before the command so observers see the event at
--- a/internal/controller/failover.go
+++ b/internal/controller/failover.go
@@ -60,6 +60,21 @@ func findFailoverShard(state *valkey.ClusterState, address string) (*valkey.Shar
 	return shard, replicas
 }
 
+// highestOffsetReplica returns the synced replica with the greatest replication
+// offset, the one most caught up with the primary. Replicas whose offset is
+// unavailable sort last and ties keep discovery order, so the result is stable
+// and never nil for a non-empty slice.
+func highestOffsetReplica(replicas []*valkey.NodeState) *valkey.NodeState {
+	best := replicas[0]
+	bestOffset := best.ReplicationOffset()
+	for _, replica := range replicas[1:] {
+		if offset := replica.ReplicationOffset(); offset > bestOffset {
+			best, bestOffset = replica, offset
+		}
+	}
+	return best
+}
+
 // proactiveFailover issues CLUSTER FAILOVER to the best synced replica in shard,
 // then polls until the replica reports role:master or the timeout is reached.
 // shard must be non-nil; replicas must be non-empty.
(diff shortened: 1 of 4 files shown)


internal/controller/failover.go
```diff
@@ -60,16 +60,33 @@ func findFailoverShard(state *valkey.ClusterState, address string) (*valkey.Shar
...
 
-	// Pick the first synced replica as the failover target. The ordering is
-	// determined by node discovery order — no priority scheme is applied yet.
-	target := replicas[0]
+	// Fail over to the most caught-up replica (highest replication offset). A
+	// graceful CLUSTER FAILOVER holds writes on the primary until the target
+	// replica catches up, so promoting the furthest-ahead one minimises that
+	// write pause and the exposure if the primary dies mid-failover.
+	target := highestOffsetReplica(replicas)
 	log.Info("initiating proactive failover", "shard", shard.Id, "target", target.Address)
```
jdheyburn: What happens when there is no available replica?

internal/controller/failover.go
```diff
@@ -60,16 +60,33 @@ func findFailoverShard(state *valkey.ClusterState, address string) (*valkey.Shar
...
+// and never nil for a non-empty slice.
+func highestOffsetReplica(replicas []*valkey.NodeState) *valkey.NodeState {
+	best := replicas[0]
+	bestOffset := best.ReplicationOffset()
+	for _, replica := range replicas[1:] {
+		if offset := replica.ReplicationOffset(); offset > bestOffset {
+			best, bestOffset = replica, offset
+		}
+	}
+	return best
```
jdheyburn: There is some similar code also in #244, could these align to prevent duplication?

jdheyburn: What happens when offset=0 (replica is connected but is in replicating state)? We wouldn't want to failover to that replica.

In my scripts for our helm deployment, I look at this criteria to determine if the replica is healthy:

1. `status=online`
2. `offset>0`
3. `lag<=1`

Would it be useful to have something similar here?

melancholictheory:
thanks for the review. going through the three.

on the no-available-replica case: this path doesn't reach highestOffsetReplica. findFailoverShard returns nil when there are no synced replicas, so proactiveFailover is only ever called with a non-empty slice, and a shard with no synced replica just rolls the primary without a proactive failover (same as today). i left the function with that precondition rather than a nil return, but happy to add an explicit guard plus comment if you'd rather it be defensive at the boundary.

on the overlap with #244: agree, we're both picking the highest-offset replica (promoteOrphanedReplicas does the same for the TAKEOVER path), so it makes sense to have one primitive. i'd move the selection next to ReplicationOffset in the valkey package so the proactive-failover and the recovery path share it. happy to either pull it out here and have #244 rebase onto it, or rebase onto #244 if that lands first, whichever is easier for you and @bjosv.

on offset=0: good catch, that's the real gap. a replica that's link-up but still at offset 0 hasn't taken the dataset yet, so promoting it would lose data, and the graceful failover would probably just time out waiting for it to catch up. mapping your three criteria onto what the operator already has:

- status=online is roughly the synced filter we apply today (master_link_status:up in GetSyncedReplicas).
- offset>0 is directly available, it's the same slave_repl_offset this PR already reads, so i can gate on it right away.
- lag<=1 lives on the primary's INFO (the slaveN ...,lag= line), or as a byte-delta from master_repl_offset - slave_repl_offset on the replica. doable, just a bit more plumbing.

since the graceful CLUSTER FAILOVER already holds writes until the target catches up, offset>0 is the one that actually prevents data loss (don't promote an empty replica); lag mostly affects how long that write pause is, which picking the highest offset already minimises. so my inclination is offset>0 as a hard filter in the shared primitive, with lag as a follow-up if you want the full health triplet.

one question on the hard filter: if no replica clears offset>0, would you rather skip the proactive failover and just roll (letting the cluster do an unplanned failover if it needs to), or keep it best-effort? i'd lean skip-and-roll.

happy to push the offset>0 gate and the shared primitive once we settle where it should live.

jdheyburn:
Hey @melancholictheory, #244 is merged now - would you be able to rebase and consolidate the duplicate offset logic, if it makes sense please?

melancholictheory:
rebased onto main and consolidated the offset logic. there's now a single HighestOffsetReplica in the valkey package, backed by NodeState.ReplicationOffset, and both BestReplicaOf and the proactive-failover path go through it, so the slave_repl_offset parsing lives in one spot. BestReplicaOf keeps its behaviour (its test still passes).

i left the offset>0 gate (your point about a replica still sitting at offset 0) out of this one to keep it to the rebase plus dedup. happy to add it as a follow-up, or here if you'd rather, once you've had a look. still leaning skip-and-roll when no replica clears offset>0, but your call.

bjosv:
I'm not sure that we can have a offset=0 when we have a master_link_status:up.
It seems that the link is set to `up` when the valkey `repl_state` goes to `REPL_STATE_CONNECTED`,
and this is when the replica has received a RDB file.

If there is not data in the file, then there is code to make sure the offset in increased:
https://github.com/valkey-io/valkey/blob/436dcaeee95a9f38e8326049a7175e5e2929c7f1/src/replication.c#L1620

then its also 1 from start
https://github.com/valkey-io/valkey/blob/436dcaeee95a9f38e8326049a7175e5e2929c7f1/src/server.c#L6567

bjosv (approved):

melancholictheory:
thanks, that checks out. replicaStartCommandStream sends a PING when the stream is empty specifically so the offset goes non-zero ("In cluster mode, this allows replicas to be reported as available in CLUSTER SLOTS/SHARDS right away"), and INFO seeds slave_repl_offset at 1, so a replica reporting master_link_status:up is already at offset >= 1.

so the synced filter (GetSyncedReplicas requires master_link_status:up) already rules out the offset=0 case on this path, and the offset>0 gate isn't needed here. happy to leave the PR as the rebase plus consolidation then, unless you'd still want an explicit guard for belt and braces.

jdheyburn:
Scheduling / placement epic (ties proactive replica selection into broader placement work): #265.

jdheyburn:
@greptile-apps

jdheyburn:
I think maybe it makes sense to skip roll if there isn't a healthy replica to failover to - that's what my Sentinel setup does today. But I think let's raise another issue for discussing that to ensure we don't block this PR.

jdheyburn (approved):
Thank you!

PR #252, opened 2026-06-15
fix: roll primaries shard-by-shard during rolling update without persistence

### LINKED ISSUE #250: [bug] All replicas roll across all shards before primaries — roll one shard end-to-end instead
### Problem

When a spec change triggers a rolling update on a cluster **without persistence (no AOF/RDB)**, nodes roll in this order (one at a time, sequentially):

```
shard-0-replica → shard-1-replica → shard-2-replica → shard-0-primary → shard-1-primary → shard-2-primary
```

The intended order is to complete each shard fully before moving on:

```
shard-0-replica → shard-0-primary → shard-1-replica → shard-1-primary → ...
```

**This bug only manifests when persistence is not configured.** With persistence enabled, pods take longer to restart (RDB/AOF load), giving the readiness gate time to block correctly. Without persistence, pods restart almost instantly and the gate fails to hold.

This was noted as a known limitation in the PR #116 commit message:
> "I think also the pods are being rolled too fast, so we might want to revisit the readiness checks, or introduce some additional cluster health checks when reconciling pods to ensure we're not adding fuel to the fire."

The current behaviour means shard-0's replica is rolled to the new spec, then the operator advances to shard-1's replica before ever touching shard-0's primary. By the time shard-0's primary is eventually rolled, shard-0 has been in a mixed-spec state for the entire duration of all three shards' replica rolls.

### Suspected root cause 1

It may  have something to do with the primary not having any healthy replicas to failover to, so the reconciliation progresses onto the next shard.

### Suspected root cause 2

The readiness gate in `reconcileValkeyNodes` (`internal/controller/valkeycluster_controller.go`) relies on `isWorkloadRolledOut` returning `false` long enough for the ValkeyNode controller to mark `Status.Ready=false` during pod churn. Without persistence, pods come up and pass readiness probes almost instantly (only waiting for the Valkey process to start, no data to load), so the gate clears before the shard has truly settled.

For `WorkloadTypeDeployment`, the gate condition in `internal/controller/valkeynode_controller.go` is:

```go
return dep.Status.UpdatedReplicas >= replicas && dep.Status.ReadyReplicas >= replicas, nil
```

This becomes true almost immediately with no-persistence workloads, causing `reconcileValkeyNodes` to advance to the next shard's replica before the current shard's primary has been rolled.

### Expected behaviour

The reconciler completes one shard end-to-end (roll replica → roll primary) before advancing to the next shard. This should hold regardless of whether persistence is configured.

### Acceptance criteria

- [ ] Reconciler iterates shards; within each shard, replicas roll before the primary
- [ ] The operator does not advance to the next shard until the current shard's primary has completed its roll
- [ ] Behaviour is consistent with and without persistence (AOF/RDB)
- [ ] E2E test: 3-shard cluster rolling update **without persistence** — verify shard-by-shard ordering

### References
- PR #116 (introduced sequential update logic; noted this limitation in commit message)

bjosv:
This PR closes #250 

### Summary

Without persistence, a rolled replica gets a new node/shard ID and appears as an isolated PendingNode, not in its original shard. When the operator reaches the primary, GetSyncedReplicas finds zero replicas in the shard and defers the roll. Previously, deferring advanced to the next shard's replicas, resulting in all replicas rolling before any primaries.

Now, when a primary roll is deferred, the operator issues CLUSTER MEET and REPLICATE for isolated nodes inline, then requeues. On the next reconcile the replica has rejoined the shard, failover proceeds, and the primary is rolled; completing the shard before advancing.

Introduces a nodeResult enum to replace multiple boolean return values from reconcileValkeyNode.

### Testing

This is complex to set up in a unit test (needs mocked Valkey connections).
Verified on kind.
(`CHAOS_SCENARIOS=rolling-update make test-chaos`)

### Checklist

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -511,14 +511,24 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNodes(ctx context.Context, clus
 		// actual primary (which may differ from node-index=0 after a failover)
 		// and place it last.
 		for _, nodeIndex := range replicaFirstNodeOrder(shardIndex, nodesPerShard, nodes, clusterState) {
-			requeue, nodeCreated, err := r.reconcileValkeyNode(ctx, cluster, shardIndex, nodeIndex, clusterState, configHash)
+			result, err := r.reconcileValkeyNode(ctx, cluster, shardIndex, nodeIndex, clusterState, configHash)
 			if err != nil {
 				return false, err
 			}
-			if requeue {
+			switch result {
+			case nodeDeferred:
+				// Roll deferred; MEET and REPLICATE isolated nodes so they
+				// rejoin the shard before next reconcile.
+				if _, meetErr := r.meetIsolatedNodes(ctx, cluster, clusterState); meetErr != nil {
+					log.Error(meetErr, "meetIsolatedNodes failed during deferred roll")
+				}
+				if _, replErr := r.replicatePendingReplicas(ctx, cluster, clusterState, nodes); replErr != nil {
+					log.Error(replErr, "replicatePendingReplicas failed during deferred roll")
+				}
 				return true, nil
-			}
-			if nodeCreated {
+			case nodeRequeued:
+				return true, nil
+			case nodeCreated:
 				totalCreated++
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -511,14 +511,24 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNodes(ctx context.Context, clus
...
+			case nodeDeferred:
+				// Roll deferred; MEET and REPLICATE isolated nodes so they
+				// rejoin the shard before next reconcile.
+				if _, meetErr := r.meetIsolatedNodes(ctx, cluster, clusterState); meetErr != nil {
+					log.Error(meetErr, "meetIsolatedNodes failed during deferred roll")
+				}
+				if _, replErr := r.replicatePendingReplicas(ctx, cluster, clusterState, nodes); replErr != nil {
+					log.Error(replErr, "replicatePendingReplicas failed during deferred roll")
+				}
 				return true, nil
```
jdheyburn, line 528: Am I right in saying this is the net new code that fixes the bug? The other lines are just adapting to the enum refactor?

bjosv: Yes, this and reconcileValkeyNode returning a 3rd variant. Instead of adding yet another bool I changed it to typed constants.
Optionally reconcileValkeyNode can handle the meet/replicate but then it needs `nodes`, while working on a single node.

jdheyburn (approved):
LGTM, thank you for the enum refactor too!

Issue #260, opened 2026-06-18
[enhancement] Validate or enforce terminationGracePeriodSeconds >= cluster-manual-failover-timeout


jdheyburn:
### Problem

The graceful `CLUSTER FAILOVER` handover (both proactive and `shutdown-on-sigterm`) requires the primary pod to stay alive long enough for the handover to complete before Kubernetes sends SIGKILL. If `terminationGracePeriodSeconds` is shorter than `cluster-manual-failover-timeout`, Kubernetes will forcibly kill the pod mid-handover, leaving the cluster without a healthy primary.

The default `terminationGracePeriodSeconds` in Kubernetes is 30s. The default `cluster-manual-failover-timeout` in Valkey is 5000ms (5s), but users may configure longer timeouts.

### Proposed fix

The operator should validate — either at admission via a CEL rule or during reconciliation — that `spec.template.terminationGracePeriodSeconds` ≥ the configured `cluster-manual-failover-timeout`. If they conflict, surface a `ConfigurationWarning` condition on the `ValkeyCluster` explaining the risk.

Since `shutdown-on-sigterm failover` is always enabled (see #248), the operator should automatically set `terminationGracePeriodSeconds` to `max(current, cluster-manual-failover-timeout + 10s)` if the user has not explicitly set it.

### Acceptance criteria

- [ ] Validation warns (condition + event) when `terminationGracePeriodSeconds` < `cluster-manual-failover-timeout`
- [ ] CEL admission rule or reconcile-time check
- [ ] Documented in the `ValkeyCluster` CRD field comments

### References

- Discussion #231 (terminationGracePeriodSeconds note)
- Companion: #248 (`shutdown-on-sigterm failover` support)

jdheyburn:
Also ensure that the docs for graceful shutdown are updated to reflect the new behaviour, is any has changed.

- https://github.com/valkey-io/valkey-operator/pull/268/changes/5838146569dd82b90082620615185f6e75ea2173

melancholictheory:
started looking at this and wanted to check the shape before building it, since it touches the CRD.

[...]

PR #268, opened 2026-06-19
feat: inject shutdown-on-sigterm failover by default

### LINKED ISSUE #248: [feat] Support shutdown-on-sigterm failover config
### Problem

The operator's proactive failover path covers **planned** pod rolls initiated by the operator. It does not cover **unplanned** termination — node drains, evictions, spot preemption, or `kubectl delete pod` — where the operator never has a chance to run the failover before the pod disappears.

Valkey 9.0 introduced the `shutdown-on-sigterm failover` config directive (valkey/valkey#1091). When set on a primary node, Valkey issues a graceful `CLUSTER FAILOVER` to a replica on SIGTERM before the process exits. This is the node-local safety net for unplanned termination.

Since the operator targets Valkey 9.0+, this directive should be injected by default and enforced the `terminationGracePeriodSeconds` constraint needed for it to work reliably (see companion issue).

### Proposed design

- Inject `shutdown-on-sigterm failover` into the managed Valkey config by default for all `ValkeyCluster` resources
- No double-failover risk: the directive only fires if the node is still primary at SIGTERM; it is a no-op on replicas

### Acceptance criteria

- [ ] `shutdown-on-sigterm failover` directive injected into `valkey.conf` by default
- [ ] Works alongside the operator's proactive failover without double-firing
- [ ] E2E test: drain the node running the primary; verify a replica is promoted before the pod exits

### References
- Discussion #231 (shutdown-on-sigterm section)
- Discussion comment referencing #120
- valkey/valkey#1091 (upstream)

melancholictheory:
Closes #248

### Summary

Inject `shutdown-on-sigterm failover` into the managed Valkey config by default for every ValkeyCluster. On SIGTERM (node drain, eviction, preemption, or `kubectl delete pod`), a primary fails its slots over to a replica during graceful shutdown, which covers the unplanned termination that the operator's own proactive failover never sees.

### Implementation

- Added the directive to `getBaseConfig`, alongside the other operator-managed defaults.
- It is a Valkey 9.0+ directive, which matches the operator's documented baseline, so it renders unconditionally with no version gate.
- No double-failover risk: it only fires if the node is still primary at SIGTERM, and it is a no-op on replicas.

### Acceptance criteria

- [x] `shutdown-on-sigterm failover` injected into valkey.conf by default
- [x] Works alongside the operator's proactive failover without double-firing (no-op on replicas)
- [ ] E2E test (drain the primary's node, verify a replica is promoted before the pod exits)

I left the E2E test out for now since it needs node-drain orchestration in the e2e harness. Happy to add it here or as a follow-up, whichever you prefer. The unit test asserts the directive is rendered.

The `terminationGracePeriodSeconds` constraint is tracked separately in #260; the docs note the relationship.

### Testing

- Unit test asserts `shutdown-on-sigterm failover` is in the rendered config.

### References

- Discussion #231 (shutdown-on-sigterm section)
- #120 (out-of-band termination discussion)
- valkey/valkey#1091

### Checklist

### DIFF
--- a/docs/valkeycluster.md
+++ b/docs/valkeycluster.md
@@ -120,6 +120,12 @@ The operator creates a `PodDisruptionBudget` with `maxUnavailable: 1` selecting 
 | `Managed` | Operator creates and owns the PDB |
 | `Disabled` | Operator deletes the PDB if it exists and does not recreate it |
 
+### Graceful shutdown
+
+On `SIGTERM` (a node drain, eviction, or preemption), a cluster primary fails its slots over to a replica before exiting, so descheduling a primary the operator did not initiate does not leave the shard without a writer. This is enabled by default through the `shutdown-on-sigterm failover` server config and requires Valkey 9.0+.
+
+The handoff runs inside the pod's termination grace period, so set `terminationGracePeriodSeconds` at least as high as `cluster-manual-failover-timeout`; otherwise `SIGKILL` can cut the failover short.
+
 ### Private image registries
 
 ```yaml
--- a/docs/valkeycluster.md
+++ b/docs/valkeycluster.md
@@ -124,7 +124,7 @@ The operator creates a `PodDisruptionBudget` with `maxUnavailable: 1` selecting 
 
 On `SIGTERM` (a node drain, eviction, or preemption), a cluster primary fails its slots over to a replica before exiting, so descheduling a primary the operator did not initiate does not leave the shard without a writer. This is enabled by default through the `shutdown-on-sigterm failover` server config and requires Valkey 9.0+.
 
-The handoff runs inside the pod's termination grace period, so set `terminationGracePeriodSeconds` at least as high as `cluster-manual-failover-timeout`; otherwise `SIGKILL` can cut the failover short.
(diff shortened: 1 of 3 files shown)


melancholictheory:
on the grace-period note: the numbers are the other way round. the default cluster-manual-failover-timeout is 5s, not 30s (config.c registers it with a 5000ms default), so with the Kubernetes default terminationGracePeriodSeconds of 30s there is roughly 25s of margin and the failover completes out of the box. updated the docs to spell out both defaults and to say you only need to bump the grace period if you raise the timeout.

on whether shutdown-on-sigterm should be overridable: fair distinction, it is more of an operational preference than the correctness settings it sits next to in getBaseConfig. the operator doesn't have an "overridable defaults" layer today, it is either enforced base config or user config. #248 framed it as injected by default, so i kept it enforced, but happy to move it to an overridable layer if you'd rather users could set nosave / default. @jdheyburn that one is a design call for you.

jdheyburn:
Let's keep it enforced for now, if there is a need to override the defaults then we'll consider that when we get there.

jdheyburn (approved):
LGTM, thank you for updating the docs too!

jdheyburn:
For the e2e test, I think a follow up issue is fine for now.

bjosv (approved):
Nice!

PR #271, opened 2026-06-23
feat: configurable terminationGracePeriodSeconds for graceful failover

### LINKED ISSUE #260: [enhancement] Validate or enforce terminationGracePeriodSeconds >= cluster-manual-failover-timeout

melancholictheory:
Closes #260

### Summary

Add `spec.terminationGracePeriodSeconds` to `ValkeyCluster` (threaded through to `ValkeyNode` and onto the pod) so the graceful `CLUSTER FAILOVER` triggered on SIGTERM has time to hand the shard off to a replica before SIGKILL.

### Design

Following the direction in #260:

- New `TerminationGracePeriodSeconds *int64` on both `ValkeyClusterSpec` (user-facing) and `ValkeyNodeSpec` (the cluster controller threads it through to the pod). Not grouped under a pod-template surface yet, per your note that we can redesign that while still in alpha.
- When unset, the operator derives a safe value: `max(30s, cluster-manual-failover-timeout / 1000 + 10s)`. With the defaults (5s timeout) that stays at the Kubernetes default of 30s. Raising the timeout pulls the grace period up with it.
- An explicit value is honoured as-is, so a user who wants a long grace period (for example to allow a final RDB snapshot) gets exactly what they asked for. If the value is below the recommended minimum, the operator emits a `GracePeriodTooShort` warning event on the `ValkeyCluster` instead of silently overriding it.

### On the "block the apply" idea

You floated blocking the manifest apply when the value is too short. i went with respect-the-value-and-warn instead, because the recommended minimum depends on `cluster-manual-failover-timeout`, which lives in `spec.config` as a string map entry. a CEL admission rule can't cleanly parse and compare that, so a hard block would need a validating webhook. happy to add that as a follow-up if you'd rather it be a hard stop; the warning event is the lighter "inform the user" mechanism for now.

### Acceptance criteria

- [x] Warns (event + log) when `terminationGracePeriodSeconds` is below the recommended minimum
- [~] Reconcile-time check (CEL block deferred, see above)
- [x] Documented in the CRD field comments and `docs/valkeycluster.md`

### Testing

- Unit tests for the timeout / recommended / effective grace-period helpers.

Docs note: i added a `Termination grace period` section, which sits next to the `Graceful shutdown` section from #268 once that merges.

### Checklist

### DIFF
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -147,6 +147,7 @@ type ValkeyClusterSpec struct {
 	// replica before SIGKILL. When unset, the operator derives a safe default
 	// from cluster-manual-failover-timeout. A value below that derived minimum
 	// is honoured but reported as a warning.
+	// +kubebuilder:validation:Minimum=1
 	// +optional
 	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
 
@@ -245,6 +246,10 @@ const (
 	ConditionDegraded      = "Degraded"
 	ConditionClusterFormed = "ClusterFormed"
 	ConditionSlotsAssigned = "SlotsAssigned"
+	// ConditionConfigurationWarning flags a spec value the operator accepted but
+	// considers risky, for example a terminationGracePeriodSeconds too short for
+	// graceful failover.
+	ConditionConfigurationWarning = "ConfigurationWarning"
 )
 
 const (
@@ -264,6 +269,8 @@ const (
 	ReasonTopologyComplete         = "TopologyComplete"
 	ReasonAllSlotsAssigned         = "AllSlotsAssigned"
 	ReasonSlotsUnassigned          = "SlotsUnassigned"
+	ReasonGracePeriodTooShort      = "GracePeriodTooShort"
+	ReasonConfigurationValid       = "ConfigurationValid"
 	ReasonPrimaryLost              = "PrimaryLost"
 	ReasonNoSlots                  = "NoSlotsAvailable"
 	ReasonRebalancingSlots         = "RebalancingSlots"
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -141,6 +141,15 @@ type ValkeyClusterSpec struct {
 	// +optional
 	Config map[string]string `json:"config,omitempty"`
 
+	// TerminationGracePeriodSeconds is the pod termination grace period for the
+	// Valkey nodes. It must give the graceful CLUSTER FAILOVER triggered on
+	// SIGTERM (shutdown-on-sigterm) enough time to hand the shard off to a
+	// replica before SIGKILL. When unset, the operator derives a safe default
+	// from cluster-manual-failover-timeout. A value below that derived minimum
+	// is honoured but reported as a warning.
+	// +optional
+	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
+
 	// TLS configuration for the cluster
 	// +optional
 	TLS *TLSConfig `json:"tls,omitempty"`
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -270,7 +270,6 @@ const (
 	ReasonAllSlotsAssigned         = "AllSlotsAssigned"
 	ReasonSlotsUnassigned          = "SlotsUnassigned"
 	ReasonGracePeriodTooShort      = "GracePeriodTooShort"
-	ReasonConfigurationValid       = "ConfigurationValid"
 	ReasonPrimaryLost              = "PrimaryLost"
 	ReasonNoSlots                  = "NoSlotsAvailable"
 	ReasonRebalancingSlots         = "RebalancingSlots"
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -153,6 +153,18 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// ConfigMap to avoid a race condition where the cache does not have the ConfigMap
 	configHash := serverConfigRollHash(cluster)
 
+	// Warn when an explicit terminationGracePeriodSeconds is too short for the
+	// graceful failover on SIGTERM to finish before SIGKILL. The value is still
+	// honoured; the operator does not silently override it.
+	if g := cluster.Spec.TerminationGracePeriodSeconds; g != nil {
+		if rec := recommendedGracePeriodSeconds(cluster); *g < rec {
+			log.Info("terminationGracePeriodSeconds is below the recommended minimum for graceful failover",
+				"requested", *g, "recommended", rec)
+			r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, "GracePeriodTooShort", "ReconcileValkeyCluster",
+				"spec.terminationGracePeriodSeconds (%ds) is below the recommended %ds for cluster-manual-failover-timeout; SIGKILL may interrupt the graceful failover on shutdown", *g, rec)
+		}
+	}
+
 	nodes := &valkeyiov1alpha1.ValkeyNodeList{}
 	if err := r.List(ctx, nodes, client.InNamespace(cluster.Namespace), client.MatchingLabels(map[string]string{LabelCluster: cluster.Name})); err != nil {
 		log.Error(err, "failed to list ValkeyNodes")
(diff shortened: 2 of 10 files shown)


internal/controller/valkeycluster_controller.go
```diff
@@ -675,28 +732,31 @@ func buildClusterValkeyNode(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex
...
+			Affinity:                      cluster.Spec.Affinity,
+			Tolerations:                   cluster.Spec.Tolerations,
+			TopologySpreadConstraints:     cluster.Spec.TopologySpreadConstraints,
+			Exporter:                      cluster.Spec.Exporter,
+			Containers:                    cluster.Spec.Containers,
+			ServerConfigMapName:           GetServerConfigMapName(cluster.Name),
+			UsersACLSecretName:            getInternalSecretName(cluster.Name),
+			TLS:                           cluster.Spec.TLS,
+			Config:                        cluster.Spec.Config,
+			TerminationGracePeriodSeconds: &gracePeriod,
```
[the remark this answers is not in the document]
jdheyburn: @greptile-apps Is there a best practice we could be following to avoid these full rolling restarts on operator upgrades?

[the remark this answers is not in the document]
melancholictheory: that targeted fix is what i pushed in c390120: only set the field when the resolved value deviates from the k8s default of 30s, so existing nodes that resolve to 30 keep nil and don't see a diff on upgrade.

on the general question: the nil-on-default trick works but it's per-field and easy to forget next time someone adds an optional field. the two structural options are a defaulting webhook (`+kubebuilder:default`, or a mutating webhook materialises the value at admission, so stored objects always carry it and there's no nil->value transition on later upgrades), or moving the node reconcile from `CreateOrUpdate` to server-side apply with field ownership, so the manager only patches what it owns and new fields don't churn existing resources. SSA is the more general fix but a bigger change than this PR. happy to open a follow-up issue to track it if you think it's worth doing.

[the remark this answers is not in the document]
jdheyburn: Let's raise a discussion on `CreateOrUpdate` vs `SSA`, so that we don't lose track of it.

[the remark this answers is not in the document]
melancholictheory: raised it here so we don't lose track: #272

internal/controller/valkeycluster_controller.go
```diff
@@ -153,6 +153,24 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// ConfigMap to avoid a race condition where the cache does not have the ConfigMap
 	configHash := serverConfigRollHash(cluster)
 
+	// Surface a ConfigurationWarning condition when an explicit
+	// terminationGracePeriodSeconds is too short for the graceful failover on
+	// SIGTERM to finish before SIGKILL. The value is honoured; the operator does
+	// not silently override it. The condition is idempotent, so the event fires
+	// only when the cluster first enters the warning state.
+	if g := cluster.Spec.TerminationGracePeriodSeconds; g != nil && *g < recommendedGracePeriodSeconds(cluster) {
+		rec := recommendedGracePeriodSeconds(cluster)
```
sandeepkunusoth: nit: can we call recommendedGracePeriodSeconds once here? we can create constant and reuse that? if recalculation is not required?

internal/controller/valkeycluster_controller.go
```diff
@@ -686,29 +749,38 @@ func buildClusterValkeyNode(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex
 	l[LabelShardIndex] = strconv.Itoa(shardIndex)
 	l[LabelNodeIndex] = strconv.Itoa(nodeIndex)
 
+	// Only set the field when it differs from the Kubernetes default, so existing
+	// clusters that resolve to the default keep a nil value and are not rolled on
+	// operator upgrade.
+	var gracePeriod *int64
+	if g := effectiveGracePeriodSeconds(cluster); g != defaultGracePeriodSeconds {
+		gracePeriod = &g
```
sandeepkunusoth, line 763: what happens when previous graceperiod is 60 sec and if user wants to change graceperiod to default value 30 sec will this check cause updates to ignore silently?

internal/controller/valkeycluster_controller.go
```diff
@@ -671,6 +689,51 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNode(ctx context.Context, clust
...
+// failoverTimeoutSeconds returns cluster-manual-failover-timeout in whole
+// seconds (rounded up), or the Valkey default when unset or unparseable.
+func failoverTimeoutSeconds(config map[string]string) int64 {
+	v, ok := config["cluster-manual-failover-timeout"]
+	if !ok {
+		return defaultFailoverTimeoutSeconds
+	}
+	ms, err := strconv.ParseInt(strings.TrimSpace(v), 10, 64)
+	if err != nil || ms <= 0 {
+		return defaultFailoverTimeoutSeconds
```
sandeepkunusoth, line 713: (nit) i think its better to add logs in error cases in 709 and 712 when it falls back to defaultFailoverTimeoutSeconds

api/v1alpha1/valkeycluster_types.go
```diff
@@ -249,6 +263,8 @@ const (
 	ReasonTopologyComplete         = "TopologyComplete"
 	ReasonAllSlotsAssigned         = "AllSlotsAssigned"
 	ReasonSlotsUnassigned          = "SlotsUnassigned"
+	ReasonGracePeriodTooShort      = "GracePeriodTooShort"
+	ReasonConfigurationValid       = "ConfigurationValid"
 	ReasonPrimaryLost              = "PrimaryLost"
```
sandeepkunusoth, line 273: ReasonConfigurationValid is not used anywhere

internal/controller/valkeycluster_controller.go
```diff
@@ -153,6 +153,24 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	// SIGTERM to finish before SIGKILL. The value is honoured; the operator does
+	// not silently override it. The condition is idempotent, so the event fires
+	// only when the cluster first enters the warning state.
+	if g := cluster.Spec.TerminationGracePeriodSeconds; g != nil && *g < recommendedGracePeriodSeconds(cluster) {
+		rec := recommendedGracePeriodSeconds(cluster)
+		msg := fmt.Sprintf("spec.terminationGracePeriodSeconds (%ds) is below the recommended %ds for cluster-manual-failover-timeout; SIGKILL may interrupt the graceful failover on shutdown", *g, rec)
+		if !meta.IsStatusConditionTrue(cluster.Status.Conditions, valkeyiov1alpha1.ConditionConfigurationWarning) {
+			log.Info("terminationGracePeriodSeconds is below the recommended minimum for graceful failover",
+				"requested", *g, "recommended", rec)
+			r.Recorder.Eventf(cluster, nil, corev1.EventTypeWarning, valkeyiov1alpha1.ReasonGracePeriodTooShort, "ReconcileValkeyCluster", "%s", msg)
```
sandeepkunusoth, line 167: (nit) can u update docs/status-conditions.md with this new status.

internal/controller/valkeycluster_controller.go
melancholictheory: good call, computed it once before the branch now. `4c16269`.

internal/controller/valkeycluster_controller.go
melancholictheory: it doesn't drop the update silently: going from `&60` to nil is still a spec change, so the pod does update to 30 (via the kubernetes default). but you're right that storing nil for an explicitly requested value is confusing, so i changed it to store explicit values verbatim, including 30. the field only stays nil for an *unset* value that resolves to the default, which is the case that keeps existing clusters from rolling on upgrade. `4c16269`.

internal/controller/valkeycluster_controller.go
melancholictheory: the missing-key path (709) is the common case, so a log there would fire on every reconcile for clusters that never set the timeout. a malformed value (712) also fails at valkey boot, and validating config values is tracked separately in #141, so i left the helper silent for now. happy to add a warning specifically for the unparseable case if you'd prefer it here.

api/v1alpha1/valkeycluster_types.go
melancholictheory: removed, thanks. `4c16269`.

internal/controller/valkeycluster_controller.go
melancholictheory: added a `ConfigurationWarning` section to `docs/status-conditions.md` in `4c16269`.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -236,6 +246,10 @@ const (
 	ConditionDegraded      = "Degraded"
 	ConditionClusterFormed = "ClusterFormed"
 	ConditionSlotsAssigned = "SlotsAssigned"
+	// ConditionConfigurationWarning flags a spec value the operator accepted but
+	// considers risky, for example a terminationGracePeriodSeconds too short for
+	// graceful failover.
+	ConditionConfigurationWarning = "ConfigurationWarning"
```
jdheyburn, line 252: This will be useful elsewhere, thanks!

jdheyburn (approved):
LGTM - thanks for raising! Will let @sandeepkunusoth have another look too.

api/v1alpha1/valkeycluster_types.go
melancholictheory: thanks for the review. agreed, the `ConfigurationWarning` condition and the recommended-minimum helper should carry over to any config-versus-resource guard down the line. happy to iterate if anything comes up downstream.

sandeepkunusoth (approved):

jdheyburn (approved):
Thank you!

PR #273, opened 2026-06-23
fix: filter excess shard pods in scheduling issue check

daanvinken:
## Summary

Follow-up to #258. The inner `handlePodSchedulingIssues` call in the `reconcileValkeyNodes` requeue branch (line 167) still lists all cluster pods, including those from excess shards pending removal. During a concurrent rolling update and scale-in, this can block progress.

## Implementation

Filter `findPodSchedulingIssue` to skip pods whose `shard-index` label is >= the desired shard count. Only pods belonging to desired shards are checked for scheduling issues. Excess shard pods are ignored since they will be cleaned up by `handleScaleIn`.

## Limitations

None. The filter is safe because excess shard pods are always cleaned up by `handleScaleIn` regardless of their scheduling state.

## Testing

Compiles clean. Existing test failures on main are pre-existing (unrelated `reconcileValkeyNode` signature change).

### DIFF
--- a/internal/controller/valkeycluster_controller.go
+++ b/internal/controller/valkeycluster_controller.go
@@ -417,7 +417,14 @@ func (r *ValkeyClusterReconciler) findPodSchedulingIssue(ctx context.Context, cl
 		return nil, fmt.Errorf("list Valkey pods: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


jdheyburn (approved):
Looks good, thank you!

PR #274, opened 2026-06-23
refactor: use deferred status patch in ValkeyNode controller

daanvinken:
## Summary

Replace scattered `Status().Patch` calls in the ValkeyNode controller with a single deferred patch at the end of Reconcile, following the common pattern in operators like CNPG and Strimzi. Idea was discussed at https://github.com/valkey-io/valkey-operator/pull/257#discussion_r3432773916.

## Implementation

- Snapshot node at the top of Reconcile with `client.MergeFrom(node.DeepCopy())`
- Local closures mutate `node.Status` in memory
- Single `defer` flushes the patch, with named return values to propagate errors
- Eliminates `setReadyCondition`, `setLiveConfigCondition`, `clearLiveConfigCondition`, `updateStatus`

## Testing

Unit tests pass. Added Reconcile-level tests for Ready=false when no pod exists and idempotency.

### DIFF
--- a/internal/controller/valkeynode_controller.go
+++ b/internal/controller/valkeynode_controller.go
@@ -21,7 +21,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"maps"
-	"reflect"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


daanvinken:
Closing this for now. After looking at the diff more carefully, I don't think the refactor justifies itself at the current codebase size. With only 4 status update paths, the old code with #257's setReadyCondition fix is already clear and each helper method is independently testable. 

The deferred pattern adds closures, named return values, and error merging in the defer, which makes the Reconcile function harder to follow without solving a real problem at this scale. The extra API calls (4 vs 1 per reconcile) are negligible.

If the controller grows significantly more status update paths in the future, this pattern would make more sense. For now the current code is fine.

PR #285, opened 2026-06-27
Init cluster metrics

SuperQ:
### Summary

Early in the cluster controller reconcile loop call `WithLabelValues` in order to make sure all metrics exist for a cluster. This espectially helps init the error counters at zero.

This is a followup to https://github.com/valkey-io/valkey-operator/pull/159.

All of the operator cluster metrics will now exist for each managed cluster.

### Checklist

### DIFF
--- a/internal/controller/failover.go
+++ b/internal/controller/failover.go
@@ -41,6 +41,20 @@ const (
 	eventActionProactiveFailover = "ProactiveFailover"
 )
 
+type FailoverType string
+
+const (
+	FailoverProactive FailoverType = "proactive"
+)
+
+var FailoverTypes = []FailoverType{
+	FailoverProactive,
+}
+
+func (ft FailoverType) String() string {
+	return string(ft)
+}
+
 // findFailoverShard returns the shard and its synced replicas if the node at
 // address is a primary that should be gracefully failed over before being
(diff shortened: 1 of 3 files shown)


SuperQ:
I would also recommend renaming the `type` label to `failover_type`. The Prometheus recommendation is to avoid this as it's a bit generic and can bloat TSDB inverted indexes. As well as conflict with external labels.

bjosv (approved):

bjosv:
> I would also recommend renaming the `type` label to `failover_type`. The Prometheus recommendation is to avoid this as it's a bit generic and can bloat TSDB inverted indexes. As well as conflict with external labels.

Changing it sounds good to me. What about the label "state" in clusterStateInfo, does it have the same problem?

SuperQ:
Good question, `state` is commonly used across a lot of metrics. Including Prometheus core stuff. But I haven't seen anyone use it as an external label.

But having it renamed to `cluster_state` does make sense for consistency and adds some extra semantic meaning. 

Do you want to do the renames in this PR or a separate one?

bjosv:
> Do you want to do the renames in this PR or a separate one?

We can do that in a separate since this was about metrics init, I'll merge this

PR #292, opened 2026-07-04
test: add default user password for e2e cluster test

### LINKED ISSUE #246: [test]: Valkey falls back to default user on auth failure, masking ACL permission test failures
## Background

While implementing #144, I noticed that Valkey silently falls back to the `default` user when authentication fails. This undermines any CI check we build for verifying `_operator` permissions, since a misconfigured credential would cause tests to pass under the wrong identity.

## Problem

When auth fails, commands are executed as `default` (which currently has `nopass` and full permissions). This can mask failures in several ways:

- A permission missing from `_operator`'s ACL appears to work because `default` has it
- A credential bug (wrong password, secret not mounted) goes undetected
- ACL regressions introduced in future PRs ship silently — no test fails
- It's impossible to assert that `_operator` is the sole identity performing operations, which matters for security audits

## Steps to reproduce

```bash
kubectl apply -f ./config/samples/v1alpha1_valkeycluster-with-user.yaml
```

```bash
kubectl exec valkey-cluster-sample-with-users-0-0-0 -- valkey-cli acl list
Defaulted container "server" out of: server, metrics-exporter
user _exporter on sanitize-payload #275d14db7ec2199b9193440729c9af7e5e7542900828004948a07478fece7f23 resetchannels -@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking
user _operator on sanitize-payload #9df726b6a40cdc02a379fde548759079c25c66500b57fd1cfd3df58bd982a10b resetchannels -@all +@connection +cluster|myid +cluster|myshardid +cluster|info +cluster|nodes +cluster|meet +cluster|addslotsrange +cluster|replicate +cluster|forget +cluster|failover +cluster|getslotmigrations +cluster|migrateslots +cluster|set-config-epoch +config|set +info
user alice on sanitize-payload #a71153805265764af6f55b4e0ce38858cde64e6e24b9a9b14e32262760572137 ~app:* ~cache:* %R~shared:* %R~config:* %W~logs:* %W~metrics:* resetchannels &notifications:* &events:* -@all +@read +@write +@connection -@admin -@dangerous +client|setname +debug
user bob on nopass sanitize-payload ~* &* +@all -@admin
user charlie off sanitize-payload %R~* resetchannels -@all +@read
user david on sanitize-payload #7447bd019c69af5975c54072b40f9c24d1105836cbd68408d6df7be76ac42ab1 #4b31cf3c1347d94fe80efb0c848579c5730d63efef2f5eaf32f78a7ca251833b resetchannels -@all +@admin
user default on nopass sanitize-payload ~* &* +@all
user edward on sanitize-payload resetchannels -@all +@admin
```

Notice `default` is `on nopass ~* &* +@all` — any failed auth silently runs as this user.

## Constraints

The `default` user can't simply be disabled today: the operator uses it to run commands like `ACL LIST` that `_operator` doesn't have permission for. Any fix needs to account for this.

## Proposed solutions

1. **Set a password on `default`** — prevents silent fallback, forces explicit auth failure
2. **Disable `default` entirely and grant `_operator` the missing permissions** (e.g. `ACL LIST`) — cleaner from a least-privilege standpoint but requires an ACL audit
3. **Assert authenticated identity before running permission checks** — call `ACL WHOAMI` immediately after connecting in the CI test and fail fast if the active user isn't `_operator`

tkarger:
This PR closes #246 

### Summary

Sets a password on the default Valkey user in the e2e test cluster to prevent silent authentication fallback. Previously, any failed auth attempt (wrong credential, missing secret mount, misconfigured ACL) would silently execute commands as default (which had nopass and full permissions), masking real failures.

The default user in the sample cluster manifest now has a password set. The e2e test fetches that password from the Kubernetes secret at runtime and passes it explicitly to valkey-cli via -a, ensuring commands are executed under the correct identity.

### Implementation

Two changes:

- config/samples/v1alpha1_valkeycluster-with-user.yaml: adds a defaultpw entry to the secret and a default user spec with `+@all ~* &*` permissions and a passwordSecret reference.
- test/e2e/valkeycluster_test.go: before running valkey-cli, fetches defaultpw from the secret via kubectl get secret, base64-decodes it, and passes it as the -a flag. This is the minimal change — no ACL WHOAMI assertion is added yet.

### Limitations

- Does not grant ACL LIST to _operator, so the operator still relies on default having sufficient permissions to run management commands.

### Testing

The existing e2e test verifyCreatedUsers now connects with the default user's password. A missing or incorrect credential will cause valkey-cli to fail rather than silently fall back, so the test now fails fast on auth regressions.

### Checklist

### DIFF
--- a/test/e2e/valkeycluster_test.go
+++ b/test/e2e/valkeycluster_test.go
@@ -503,12 +503,7 @@ spec:
 			verifyCreatedUsers := func(g Gomega) {
 				clusterFqdn := fmt.Sprintf("valkey-%s.default.svc.cluster.local", withUserClusterName)
 
-				cmd := exec.Command(
-					"sh", "-c",
-					`kubectl get secret valkey-cluster-sample-users \
-    -n default \
-    -o jsonpath='{.data.defaultpw}'`)
-
+				cmd := exec.Command("kubectl", "get", "secrets", "valkey-cluster-sample-users", "-o", "jsonpath={.data.defaultpw}")
 				b64Password, err := utils.Run(cmd)
 				g.Expect(err).NotTo(HaveOccurred())
 
--- a/test/e2e/valkeycluster_test.go
+++ b/test/e2e/valkeycluster_test.go
@@ -20,6 +20,7 @@ limitations under the License.
 package e2e
 
 import (
+	"encoding/base64"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


sandeepkunusoth (dismissed):

test/e2e/valkeycluster_test.go
```diff
@@ -502,10 +503,24 @@ spec:
 			verifyCreatedUsers := func(g Gomega) {
 				clusterFqdn := fmt.Sprintf("valkey-%s.default.svc.cluster.local", withUserClusterName)
 
-				cmd := exec.Command("kubectl", "run", "client",
+				cmd := exec.Command(
+					"sh", "-c",
```
bjosv: I think we can remove `sh -c`  and run "kubectl" like we do in other tests?

tkarger: Yes, I applied your suggestions

bjosv (approved):
Great, thanks!

Issue #261, opened 2026-06-18
[enhancement] Event-driven ValkeyNode reconcile for timely Status.Role after failover / cluster changes


jdheyburn:
### Problem

`ValkeyNode.Status.Role` is derived from `INFO replication` during reconcile (`getValkeyRole` in `valkeynode_controller.go`). Today it is only refreshed inside the `if podReady` branch, and the controller commonly requeues on a **~60s** cadence on the healthy path.

That means `Status.Role` can stay wrong or stale for a long window after a **role change while the pod remains Ready** — for example after an unplanned or planned failover where Valkey flips master/replica but Kubernetes readiness does not force an immediate reconcile. Users (and any future internal consumer of `Status.Role`) cannot rely on the field reflecting the current Valkey role promptly after topology changes.

Additionally, when a pod becomes **not Ready**, the last role may continue to be shown because it is not cleared; fixing misleading status during not-ready is in scope as part of making role status trustworthy, but the **primary** gap this issue targets is **timeliness** after cluster/failover events without waiting on periodic requeue.

### Current implementation (summary)

- `current.Status.Role = r.getValkeyRole(ctx, current)` only inside the `if podReady` branch (~line 437).
- `RequeueAfter: 60 * time.Second` on a common healthy path (~line 176).
- Role is not cross-checked against full cluster state.

### Proposed direction (implementer discretion)

Drive **event-driven** reconcile for affected `ValkeyNode` objects so role (and related status) is refreshed soon after relevant signals, for example:

- Watch `ValkeyCluster` status / generation / conditions and re-enqueue `ValkeyNode`s that belong to the cluster when something indicates failover, rolling update, or shard topology work.
- Consider Pod watch events where appropriate (restarts, IP changes) without relying solely on `RequeueAfter`.

The outcome should be that after failover-relevant cluster changes, `Status.Role` is updated **without** waiting for an arbitrary 60s tick when Kubernetes/controller-runtime already has a signal.

Reasonable related improvements that fit the same PR if small:

- Clear `Status.Role` when the pod is not Ready so users are not shown a stale master/slave label during an unhealthy window.
- Optionally attempt role refresh logic outside the ready-only branch where safe.

### Acceptance criteria

- [ ] After failover-relevant cluster signals, affected `ValkeyNode`s reconcile without depending solely on `RequeueAfter: 60s` for the next `Status.Role` refresh (document which signals in the PR).
[...]

PR #295, opened 2026-07-06
test(e2e): add shutdown-on-sigterm failover test

### LINKED ISSUE #270: E2E test for shutdown-on-sigterm failover (drain primary node, verify replica promotion)
Follow-up to #248 / #268. The `shutdown-on-sigterm failover` directive is injected by default in #268; this is the E2E coverage that was deferred from that PR.

### Goal

Verify that on graceful termination of a primary pod, Valkey hands the shard off to a replica before the pod exits, so the shard keeps a writer through the disruption.

### Test outline

- Deploy a ValkeyCluster with at least one replica per shard.
- Pick a shard, note its primary pod and the node it runs on.
- Trigger a graceful termination of that primary: a node drain, or `kubectl delete pod` with the default grace period (both send SIGTERM).
- Assert a replica in the shard is promoted to primary before the old pod's grace period ends (role:master via INFO replication, or the shard's PrimaryId moving in CLUSTER NODES).
- Assert the shard's slots stay served and no keys are lost across the handoff.

### Notes

- The handoff has to complete inside `terminationGracePeriodSeconds`, so the e2e setup should keep that comfortably above `cluster-manual-failover-timeout` (default 5s) so the test isn't racing SIGKILL. See the grace-period note in docs and the companion #260.
- A node drain exercises the real "descheduling the operator didn't initiate" case; a plain pod delete is a lighter proxy if draining a kind node is awkward in CI.

### Refs

- #248 (parent feature)
- #268 (implementation, e2e deferred here)
- #260 (terminationGracePeriodSeconds companion)

Sagar2366:
This PR closes #270

### Summary

Adds the E2E test deferred from #268: verify that when a primary pod is gracefully terminated, the `shutdown-on-sigterm failover` directive hands the shard off to a replica before the pod exits, so the shard keeps a writer through the disruption and no data is lost.

Test-only change; no operator behaviour is modified. The new spec runs under the `failover` Ginkgo label.

### Implementation

The test follows the outline in #270:

1. Creates a `ValkeyCluster` with `shards: 3, replicas: 1` and waits for `Ready`.
2. Identifies shard 0's primary and replica, and records the primary pod's UID so the StatefulSet-recreated pod (same name) can be distinguished from the old one.
3. Writes 50 keys across the keyspace, then deletes the primary pod with the default grace period (SIGTERM path — the lighter proxy for a drain mentioned in the issue notes).
4. Asserts the replica reports `role:master` within the 30s grace window (`Eventually` timeout = `terminationGracePeriodSeconds`), i.e. the handover beat SIGKILL.
5. Asserts the shard keeps accepting writes, the replaced pod comes back (new UID) and rejoins as `role:slave`, `cluster_state:ok`, and all 50 keys read back intact.

Two things reviewers may want to pay attention to, both learned from runs of this test against Kind:

- **Roles are read live from `INFO replication`, not from `ValkeyNode.status.role`.** Right after cluster formation the status can report two primaries for a shard (every node boots as a master before `CLUSTER REPLICATE`, and the status refresh lags) — this is the staleness described in #261, and the first draft of this test flaked on exactly that. The replica is also only accepted once `master_link_status:up`, so the failover is not attempted against a still-syncing replica.
- **`VALKEYCLI_AUTH` is unset before running `valkey-cli` inside the server container** (`execValkeyPodShell` helper). The operator injects that variable for the probe scripts, and `valkey-cli` auto-sends `AUTH` as the *default* user whenever it is set — which fails (`ERR AUTH ... without any password configured for the default user`) and pollutes command output. Commands run as the default nopass user, consistent with the rest of the e2e suite. The `_operator` user cannot be used instead because its ACL has no `SET`/`GET`.

### Limitations

- Uses `kubectl delete pod` (graceful, default grace period) rather than a node drain; the issue notes name this as the acceptable lighter proxy for CI. A drain-based variant can be layered on later.
- "Promoted before the grace period ends" is asserted by bounding the promotion check at 30s after the delete, matching the default `terminationGracePeriodSeconds`; the observed promotion latency in practice is ~4s.

### Testing

Run against a 3-node Kind cluster via:

```
KIND_CLUSTER=<cluster> go test -tags=e2e ./test/e2e/ -v -ginkgo.v -ginkgo.label-filter failover
```

Result: `1 Passed | 0 Failed` in 159s. Timeline from the passing run: SIGTERM at `22:21:06.4`, replica reported `role:master` by `22:21:10.1` (~4s, well inside the 30s grace period), replaced pod rejoined as replica ~4s later, `cluster_state:ok`, `readable=50` keys plus the write made during the disruption.

`go vet -tags=e2e` and `golangci-lint run --build-tags e2e` are clean for the new file.

### Checklist


test/e2e/failover_sigterm_test.go
```diff
@@ -0,0 +1,203 @@
...
+			"ok=0; for i in $(seq 1 %d); do "+
+				"r=$(valkey-cli -c set e2e:failover:$i v$i); "+
+				"[ \"$r\" = \"OK\" ] && ok=$((ok+1)); "+
+				"done; echo written=$ok", keyCount)
+		output, err := execValkeyPodShell(primaryPod, script)
+		Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to write keys: %s", output))
+		Expect(output).To(ContainSubstring(fmt.Sprintf("written=%d", keyCount)),
+			fmt.Sprintf("Not all keys were written: %s", output))
+
+		By("gracefully terminating the primary pod (SIGTERM with default grace period)")
```
sandeepkunusoth: i think there is some overlap between this test and https://github.com/sandeepkunusoth/valkey-k8s-operator/blob/5eb8a3380eb55a6151d6e9e7df7f838de654726e/test/e2e/valkeycluster_test.go#L951. can we merge both of them together?

jdheyburn (commented):
Thanks for raising! It looks like this PR added supported for default user password for e2e cluster tests. Are you able to rebase and use that instead of unsetting the VALKEYCLI_AUTH env var?

- https://github.com/valkey-io/valkey-operator/pull/292

melancholictheory:
nice, this closes the e2e i raised in #270. the structure is solid, and a couple of details are better than the outline i sketched: reading roles live from INFO replication with the `master_link_status:up` gate (sidesteps the status lag in #261), and keying replacement detection off the pod UID.

one thing worth thinking about, since i wrote the shutdown-on-sigterm change (#268): the current assertions would mostly pass with the feature turned off too, so the test doesn't fully isolate what it's meant to prove.

- "replica reports role:master within 30s": without the directive, valkey's default SIGTERM behaviour is a plain shutdown of the primary, and the cluster then promotes the replica through normal failure detection, which can also land inside 30s. so a promotion in that window doesn't by itself mean the graceful handoff drove it.
- "no keys lost": the 50 keys are written and replicated before the delete, so they already live on the replica. they'd survive an abrupt failover just as well, since nothing is in flight during the disruption.

what shutdown-on-sigterm actually buys is an orderly handoff: the primary fails over to a caught-up replica before it exits, so there's no window where the shard has no writer and no acknowledged write is dropped. to target that, i'd write continuously through the termination (a loop of SET with unique values while the pod is terminating) and then assert every acknowledged write is readable afterwards, plus that the write-error window is near zero. that's the behaviour the directive changes; the current version mostly checks that a failover eventually happened.

not blocking, it's a good baseline, just flagging that it would likely stay green even if #268 were reverted.

test/e2e/failover_sigterm_test.go
Sagar2366: @sandeepkunusoth can you get your PR merged if not done already? I can rebase my PR on your changes.

sandeepkunusoth: This is already in main from very long time https://github.com/valkey-io/valkey-operator/blob/a635afd9f7c2f960d7bcf619e911c24e43b36aef/test/e2e/valkeycluster_test.go#L961

sandeepkunusoth: hi i think u didn't get my previous comment instead of having duplicate test at both places i was suggesting if we we can move this to existing e2e test where we are doing failover already.   you may just need to verify key are persisted and some other changes.

test/e2e/valkeycluster_test.go
```diff
@@ -1508,3 +1591,207 @@ spec:
...
+		_, err := fmt.Sscanf(idx, "%d", &i)
+		Expect(err).NotTo(HaveOccurred())
+		if i > maxIdx {
+			maxIdx = i
+		}
+	}
+
+	script := fmt.Sprintf(
+		"for i in $(seq 0 %d); do v=$(valkey-cli -t 2 -c get e2e:cw:$i 2>/dev/null | tail -n 1); echo \"$i $v\"; done", maxIdx)
+	output, err := execValkeyPodShell(pod, script)
```
sandeepkunusoth: Right now every key starts a new valkey-cli process

Instead is there anyway we can open valey-cli once and do these commands through a single instance.

sandeepkunusoth (approved):

bjosv (approved):

PR #296, opened 2026-07-07
feat: announce cluster nodes by hostname for TLS-friendly discovery

### LINKED ISSUE #297: Umbrella: spec.networking for in-cluster endpoint & connectivity configuration
## Why 

TLS on `spec.tls` is effectively unusable for standard in-cluster clients today. After `CLUSTER SLOTS` the client re-dials each node by the announced pod IP; DNS-only server certs fail SNI verification, and `InsecureSkipVerify` defeats most of the point of TLS. The first field under this block (`preferredEndpointType`, [PR](https://github.com/valkey-io/valkey-operator/pull/296) ) unblocks that path. 

## Summary

Tracking issue for networking-related configuration on `ValkeyCluster` that belongs under a shared `spec.networking` block, distinct from the out-of-cluster access work in https://github.com/valkey-io/valkey-operator/issues/276

## Scope

**In scope**: anything that only matters to clients running inside the same Kubernetes cluster as the operator:
- Which endpoint form nodes announce to in-cluster peers/clients
- The k8s DNS domain used to build those endpoints
- Operator-managed NetworkPolicies for pod-to-pod and operator-to-pod traffic

**Out of scope** — external client access, per-shard external Services, human node names, Gateway API integration. All tracked under [#276](https://github.com/valkey-io/valkey-operator/issues/276) (`spec.externalAccess`).

Overlap with [#276](https://github.com/valkey-io/valkey-operator/issues/276) is limited to a single valkey config directive (`cluster-preferred-endpoint-type`); both eventually want `hostname`, so no runtime conflict, just a small precedence rule when whichever lands second wires the config write.

## Related work

[PR](https://github.com/valkey-io/valkey-operator/pull/296) — the first field, `spec.networking.preferredEndpointType: IP | Hostname`, lands in an open PR. Motivation: standard TLS clients (go-redis, Lettuce, redis-py) verify SNI against the connection host. After `CLUSTER SLOTS`, in-cluster clients re-dial each node by the announced value; with `--cluster-announce-ip $(POD_IP)` those re-dials fail SNI against DNS-only server certs. `Hostname` mode sets pod `subdomain` to the headless Service, announces `<pod>.<svc>.<ns>.svc.cluster.local`, and adds `cluster-preferred-endpoint-type: hostname` to `valkey.conf`.

## Candidate follow-ups

**Cluster DNS domain override.**
```yaml
spec:
  networking:
    clusterDomain: cluster.local   # default
```
Today the FQDN constructed in Hostname mode hardcodes `svc.cluster.local`. Clusters running with a non-default `--cluster-domain` (`mycompany.internal`, etc.) need this configurable. Naturally paired with `preferredEndpointType: Hostname` — arguably a prerequisite for that field to work on all clusters.

**Headless Service annotations / labels.**
```yaml
spec:
  networking:
    service:
      annotations: {}
      labels: {}
```

Filed to give `spec.networking` a home so the initial `preferredEndpointType` PR isn't a solo-field block. Individual items get their own PRs referencing this issue.

daanvinken:
## Summary

fixes https://github.com/valkey-io/valkey-operator/issues/297

Add `spec.networking.preferredEndpointType` (`IP` | `Hostname`, default `IP`) to `ValkeyCluster`, mirroring valkey's own `cluster-preferred-endpoint-type` directive, so operators of TLS clusters can have nodes advertise DNS-resolvable FQDNs instead of pod IPs.

## Motivation

Standard TLS clients (go-redis, Lettuce, redis-py, …) verify SNI against the connection host. A cluster-mode client seeds against the headless Service, receives `CLUSTER SLOTS`, and re-dials each node by the announced value. The operator has always used `--cluster-announce-ip $(POD_IP)`, which forces those re-dials to pod IPs. Server certificates usually only carry DNS SANs — pod IPs are ephemeral and not practical to list — so every re-dial fails SNI verification unless the client sets `InsecureSkipVerify`. The workaround defeats most of the point of TLS.

## Implementation

New `NetworkingSpec` on `ValkeyClusterSpec`, mirrored on `ValkeyNodeSpec` and populated by the cluster controller. When `PreferredEndpointType == Hostname`:

- `pod.spec.subdomain` is set to the cluster's headless Service name so kubelet publishes `<pod>.<svc>.<ns>.svc.cluster.local` DNS records.
- Server container's command becomes `--cluster-announce-hostname <FQDN>` using `$(POD_NAME)` from the downward API.
- `cluster-preferred-endpoint-type: hostname` is added to `valkey.conf` via `buildManagedConfig`.

A cert-manager Certificate that covers `*.<headless-svc>.<ns>.svc.cluster.local` then satisfies SNI on every dial.

## Limitations

- Switching an existing cluster from `IP` to `Hostname` rolls all pods (subdomain is part of the pod template hash). Called out in the docs.
- Assumes the cluster domain is `cluster.local`. Non-default cluster domains would need this constant lifted out.
- No API change required for downstream `ValkeyCluster` consumers on the `IP` path; the field is optional and defaults to `IP`.

## Testing

- `internal/controller/valkeynode_resources_test.go`: table-driven coverage of `valkeyAnnounceArgsAndEnv` for nil / IP / Hostname, plus `TestBuildValkeyNodePodTemplateSpec_HostnameSetsSubdomain` and a matching negative case asserting IP-mode leaves `Subdomain` empty (so operator upgrades don't roll existing clusters).
- `internal/controller/config_test.go`: `buildManagedConfig` emits `cluster-preferred-endpoint-type` only on `Hostname`.
- `internal/controller/valkeycluster_controller_test.go`: `buildClusterValkeyNode` propagates `Networking` from cluster to ValkeyNode.
- `make test` passes locally; `pre-commit run --all-files` passes.

## Checklist

- [x] API change: new optional `spec.networking.preferredEndpointType` with kubebuilder default
- [x] `make generate manifests` regenerated `zz_generated.deepcopy.go` and CRDs
- [x] Backwards compatible: existing clusters without `spec.networking` unchanged
- [x] Unit tests added for both modes
- [x] Sample under `config/samples/`
- [x] Docs updated in `docs/valkeycluster.md`
- [x] DCO sign-off

Generated with Claude Code Opus 4.8.


melancholictheory:
this is squarely the TLS-discovery gap, nice to see it addressed. the shape looks good: defaulting to IP so existing clusters don't roll, wiring `pod.spec.subdomain` only in Hostname mode, and keeping the announce flags per-node.

two things from having run hostname-announce with TLS on a parallel operator:

the FQDN hardcodes `svc.cluster.local` (`valkeyAnnounceArgsAndEnv`). #297 lists the DNS domain as in-scope later, but for this PR it's more than cosmetic: on a cluster with a non-default kubelet `--cluster-domain` (plenty of on-prem and managed clusters use one), the announced name doesn't resolve, and even if it did it wouldn't match the cert SAN, so SNI fails, which is the exact thing this PR is fixing, one level up. i'd either make the domain configurable now (the DNS-domain field from #297) or, if that's deferred, document `cluster.local` as a hard prerequisite so it fails loudly rather than silently.

the announced value is a per-pod FQDN, so the server cert has to carry each pod's name as a SAN. a single wildcard `*.<svc>.<ns>.svc.<domain>` covers them all (a wildcard matches one label, and the pod name is exactly one label under the subdomain), so it's worth showing in the sample cert / docs. otherwise users wire up a DNS-SAN cert, still hit SNI failures, and it's not obvious why.

minor: does the per-pod FQDN resolve under `workloadType: Deployment`? stable `<pod>.<svc>` DNS is really a StatefulSet property; if Hostname mode only makes sense with StatefulSets, might be worth guarding or noting.

daanvinken:
@melancholictheory incredibly useful comment, thanks.

**1. Hardcoded `cluster.local`.** 
You're right, breaks on non-default kubelet `--cluster-domain`. Two options:
I'd lean toward folding in `spec.networking.clusterDomain` it's small, and shipping Hostname mode without it means anyone on a custom cluster domain gets a broken announced FQDN.

**2. Wildcard SAN guidance.** 
Fair. The sample references a Secret but never says "issue the cert with `*.<svc>.<ns>.svc.<domain>` as SAN". I'll add that to both the sample comment and the docs section.

**3. Deployment workloadType.** 
True that. Pod names under `workloadType: Deployment` aren't stable across restarts, so the announced FQDN would change on every roll and cluster gossip would treat the node as new. I'll add a CEL validation rejecting `preferredEndpointType: Hostname` combined with `workloadType: Deployment`, plus a docs note.

melancholictheory:
all three sound right. keeping clusterDomain defaulted to `cluster.local` means it stays a no-op for everyone already on the default and only the custom-domain folks touch it. and the CEL reject for Hostname + Deployment is the right call, stable per-pod DNS just isn't a thing without a StatefulSet. thanks for turning these around so fast.

daanvinken:
Rebased onto current `main` (scheduling move, grace period, PDB struct, etc.) and added e2e coverage for the rediscovery path:

`test/e2e/valkeycluster_tls_hostname_test.go`
- multi-shard (`shards: 3`) TLS cluster with `preferredEndpointType: Hostname`
- cert with wildcard SAN for pod FQDNs
- asserts `pod.spec.subdomain` is the headless Service
- asserts `CLUSTER SLOTS` announces hostnames (not pod IPs)
- multi-key `valkey-cli -c` SETs over TLS via Service DNS after topology refresh

Still draft until e2e is green in CI if we want a final pass there.

daanvinken:
Tracking issue for the networking-tree API is #365 (under epic #318). This PR will need reshape onto `spec.networking.discovery.preferredEndpointType` + root `spec.networking.clusterDomain`, and no full NetworkingSpec on ValkeyNode (cluster owns the API; node gets announce primitives only).

daanvinken:
replaced by #365 / #378

Issue #263, opened 2026-06-18
[enhancement] Introduce runtime version detection and a framework for gating features by Valkey version


jdheyburn:
### Problem

The operator supports **Valkey 9.0+** as its baseline. Features that only require 9.0 do **not** need runtime version gating against 9.0 — that is already implied by supported images.

The gap is **features that require a newer minor** than the baseline (or future majors): the operator has no consistent way to detect the running Valkey version in a pod, and no established pattern for gating those features. That leads to:

1. **Unsafe or silent behaviour** when a user enables a spec field that the running binary does not support.
2. **Ad-hoc guards** as contributors add version-dependent behaviour.

Concrete example that needs gating today:

- **`db=X` ACL database permissions** — Valkey **9.1+** only (#262)

#### 1. Runtime version detection

At reconcile time, detect the Valkey version from each running pod using `INFO server` (field `redis_version` or `valkey_version`). Cache this per pod/node to avoid a command on every reconcile pass.

Surface the detected version on `ValkeyNode.Status` (e.g. `status.valkeyVersion: "9.1.0"`) so users and the cluster controller can both observe it.

#### 2. Version-gating helper

Introduce an internal helper in the operator codebase that contributors can call to check whether a running node meets a minimum version requirement:

```go
// MeetsMinVersion returns true if the node's detected Valkey version is >= the given minimum.
// Returns false (not an error) if the version is unknown.
func MeetsMinVersion(node *ValkeyNode, min semver.Version) bool
```

#### 3. Behavior when version is unknown or insufficient

The behavior when the detected version doesn't meet the minimum is a **design decision** left to the implementer. Options to consider:

- Log a warning and skip the feature silently
- Set a `ConfigurationWarning` condition on the `ValkeyCluster` / `ValkeyNode`
- Reject the config at admission (webhook)

We don't have a strong preference; a condition-based warning approach is consistent with how other validation warnings are surfaced in the operator.

### Acceptance criteria

- [ ] Version detection reads `INFO server` at reconcile time and populates `ValkeyNode.Status.ValkeyVersion`
- [ ] A `MeetsMinVersion` (or equivalent) helper is available for use in controllers
- [ ] At least one feature **above the 9.0 baseline** is wired through the gate (e.g. #262 `Databases` / `db=` ACLs when implemented, requiring 9.1+)
- [ ] Behavior when version is unknown is documented and consistent
- [ ] Unit tests for version parsing (semver, pre-release handling) and the gate logic

### References

[...]

Issue #262, opened 2026-06-18
[feat] CRD support for Valkey 9.1 database-level ACL permissions (db=)


jdheyburn:
### Problem

Valkey 9.1 introduced database-level ACL permissions, allowing a user's access to be scoped to specific logical database indices. The full set of directives is:

- `db=<id>[,<id>...]` — restrict the user to the listed database IDs
- `alldbs` — allow the user to access all databases (default for new users)
- `resetdbs` — clear all allowed databases and the `alldbs` flag; user cannot access any database until `db=` or `alldbs` is re-added

The current `UserAclSpec` in `valkeyacls_types.go` has no `Databases` field. Users who need `db=` scoping today must use the `permissions` (raw ACL) escape hatch, which gives up the structured API and validation.

### Proposed fix

Add a `DatabasesAclSpec` struct and a `Databases` field to `UserAclSpec`, following the same pattern as `KeysAclSpec` and `ChannelsAclSpec`:

```go
type DatabasesAclSpec struct {
    // Specific database IDs the user is allowed to access.
    // Maps to Valkey: db=0,1,2
    // +optional
    IDs []int `json:"ids,omitempty"`

    // Allow the user to access all databases. Default for new users.
    // Maps to Valkey: alldbs
    // +kubebuilder:default=false
    AllDbs bool `json:"allDbs,omitempty"`

    // Clear allowed databases and the alldbs flag.
    // Maps to Valkey: resetdbs
    // +kubebuilder:default=false
    ResetDbs bool `json:"resetDbs,omitempty"`
}
```

`buildUserAcl` in `internal/controller/users.go` must be updated to emit the `db=` clause (or `alldbs`/`resetdbs`) when `Databases` is set.

### Version requirement

`db=` ACL directives require **Valkey 9.1+**. The operator should guard this at reconcile time (ideally via a shared version-gating framework once that exists). If the detected version is < 9.1 and a user spec includes `Databases`, the operator should surface a condition warning rather than silently ignoring the field.

### Acceptance criteria

- [ ] `DatabasesAclSpec` added to `valkeyacls_types.go`; `Databases` field added to `UserAclSpec`
- [ ] `buildUserAcl` emits `db=<id>[,<id>...]` / `alldbs` / `resetdbs` correctly
- [ ] CRD validation: `IDs` and `AllDbs`/`ResetDbs` are mutually exclusive where appropriate
- [ ] Version gate: condition warning if `Databases` is set on a cluster running < 9.1
[...]

PR #307, opened 2026-07-13
(feat) version gating initial support 

### LINKED ISSUE #287: (feat) Support tls-auto-reload-interval (Valkey 9.1+)
## Problem

Valkey 9.1 introduced the `tls-auto-reload-interval` configuration directive, which allows the server to periodically reload TLS certificates, keys, and CA files from disk. This enables automatic certificate rotation (for example, when using cert-manager or Kubernetes projected volumes) without requiring a Valkey restart.

The operator currently does not expose this configuration, so users must restart pods after every TLS certificate rotation.

## Proposed Change

Add an optional field to the CRD:

```yaml
spec:
  tls:
    autoReloadInterval: <int32>
```

### Behavior

* Supported only on **Valkey 9.1+**.
* The operator should set config:

  ```
  tls-auto-reload-interval <value>
  ```

  only when:

  * `spec.tls.autoReloadInterval` is set, and
  * the detected Valkey version is **>= 9.1**.
* For versions below 9.1, the directive should not be rendered or throw warning if its set according to version.

## Acceptance Criteria

* [ ] Add `spec.tls.autoReloadInterval` to `ValkeyCluster` and `ValkeyNode`.
* [ ] Render `tls-auto-reload-interval` only when the version gate passes (Valkey >= 9.1).
* [ ] Skip rendering the directive on older versions without affecting reconciliation.
* [ ] Add unit tests covering both supported and unsupported versions.
* [ ] Update documentation with the version requirement.

Contribution:
- [X] I’m willing to implement this feature.

sandeepkunusoth:
This PR closes https://github.com/valkey-io/valkey-operator/issues/287 and initial support for https://github.com/valkey-io/valkey-operator/issues/263

### Summary

This PR adds initial support for Valkey version detection and uses it to gate Valkey 9.1-specific behavior in the operator.

### Changes
- Detects the effective Valkey image version from spec.image and stores it in status. Floating tags (latest, digest-pinned) fallback to "unknown" and it will drop 9.1+ directives.
- When user set directive is dropped (version too old or undeterminable), the operators updates ConfigurationWarning condition with reason UnSupportedDirective and emits warning event.

### Implementation
- Added version parsing utilities using github.com/Masterminds/semver/v3.
- versionGatedConfig map (directive -> min supported version), gatedUserkeystoSupress filters gated user directives out of rendered valkey.conf

### Limitations

- Version detection relies on parsing the image tag. If the version cannot be determined (for example, latest), version-gated features are skipped.
- This PR only provides the initial framework for version-gated configuration.
- Only tls-auto-reload-interval is currently listed in versionGatedConfig. Future any directives can extend the map.

### Testing

- Added unit tests for version parsing and version-gated configuration rendering.
- Added e2e tests to verify gated directives are dropped when its not supported by detected version.

### Checklist

- [X] Documentation files are updated. Behavior when version is unknown is documented and consistent.
- [X] At least one feature above the 9.0 baseline is wired through the gate validated using 9.1.0 feature tls-auto-reload-interval.


jdheyburn:
Today, the config gets dropped if the user specified it but not against a supporting version. Could we have some observability to feedback when this happens? #271  introduced a `ConfigurationWarning` - perhaps we can reuse that.

Is it necessary to set the default value in the config when it is omitted? We don't do this for any other configs.

Is a new Status field ValkeyVersion necessary? It's not being consumed by anything in this PR - or was it added for another reason?

melancholictheory:
this is the version-gating shape from #287, reading the tag and bailing to unversioned for latest/digest, good. answering your three:

observability / reusing `ConfigurationWarning`: yes, that's the right home, with one distinction. fire it only when the user explicitly set a gated directive in `spec.config` and it got dropped because the image is too old or undeterminable, that's the surprise worth surfacing. don't fire it for the operator's own auto-default being skipped, the user never asked for that. a reason like `ConfigVersionGated` with a message naming the directive, the required version, and the detected one, set idempotently and removed once it's back in range, matches how `GracePeriodTooShort` behaves in #271.

the auto-default: your consistency instinct is fair, and it honestly cuts against what i argued on #287. the deciding question is whether the operator wants to be opinionated about turning cert auto-reload on for TLS clusters (keep the default, but make it observable and documented as operator-set), or stay a thin config passthrough like it is for every other directive (drop the 86400, only gate user-set values). i'd lean passthrough now: least surprising, removes the "why is there a config i didn't set" question, and a user who wants auto-reload sets it explicitly with the gate protecting them on <9.1. the cost is TLS users don't get it for free, which is a fine follow-up if there's demand.

`status.valkeyVersion`: as written it's populated from the tag and the gate also reads the tag, so the field is a parallel copy nothing consumes, which is what you're seeing. where it earns its keep is as the runtime-detected version from `INFO server`, which is what rescues a `latest` or floating tag that actually resolved to 9.1, the tag can't tell you that. so either wire it now as the authoritative source with the tag as fallback (gate prefers status, falls back to tag), or defer the field to the PR that reads INFO and consumes it. shipping it tag-derived and unread is the one option i'd avoid, it's API surface with no consumer.

internal/controller/valkeycluster_controller.go
```diff
@@ -1233,7 +1243,6 @@ func (r *ValkeyClusterReconciler) updateStatus(ctx context.Context, cluster *val
 		current.Status.ReadyShards = r.countReadyShards(state, cluster)
 		current.Status.Shards = int32(len(state.Shards))
 	}
-
```
bjosv, line 1236: nit: this is a unrelated change for this PR

docs/valkeycluster.md
```diff
@@ -139,6 +139,16 @@ When omitted, the operator picks a safe value: the larger of the Kubernetes defa
...
 
+### Version-gated config
+
+Some user-set `spec.config` directives are only valid on newer Valkey releases. When the operator detects a directive that the selected image does not support, it drops that directive from the rendered `valkey.conf` and sets a `ConfigurationWarning` condition with reason `UnsupportedConfigDirective`.
+
+The warning message names the directive, the minimum supported Valkey version, and the detected version. The operator also emits a Kubernetes `Warning` event on the transition into this state. If you later switch to a supporting image, the condition clears on the next reconcile.
+
+for example:
+
+- `tls-auto-reload-interval` requires Valkey `9.1.0` or newer
```
bjosv:
```suggestion
For example, `tls-auto-reload-interval` requires Valkey `9.1.0` or newer.
```

internal/controller/config.go
```diff
@@ -143,6 +163,47 @@ func liveConfigToApply(config map[string]string) map[string]string {
...
+	}
+
+	for key, minVersion := range versionGatedConfig {
+		if _, userSet := cluster.Spec.Config[key]; !userSet {
+			continue
+		}
+		if valkey.MeetsMinVersion(image, minVersion) {
+			continue
+		}
+		return fmt.Sprintf("spec.config.%s requires Valkey %s+, detected %s from spec.image %q", key, minVersion, version, image), true
```
bjosv: The warning will only show a single unsupported config key, even when there are multiple unsupported.
Should we iterate all and append to a string before returning it?

internal/controller/utils.go
```diff
@@ -128,6 +128,13 @@ func annotations(cluster *valkeyv1.ValkeyCluster) map[string]string {
 	return maps.Clone(cluster.Annotations)
 }
 
+func effectiveImage(image string) string {
+	if image == "" {
+		return DefaultImage
```
bjosv, line 133: It looks like we can update the DefaultImage in https://github.com/valkey-io/valkey-operator/pull/326#discussion_r3643939865, maybe you have changed the PR?

internal/controller/config.go
sandeepkunusoth: good catch.

internal/controller/utils.go
sandeepkunusoth: i think we can update defaultImage to 9.1.0 in other PR. it doesnt effect gating logic in this PR.

docs/valkeycluster.md
```diff
@@ -276,6 +286,13 @@ tls:
 | `tls.crt` | Server certificate (or chain) |
 | `tls.key` | Private key for the certificate |
 
+Set `tls-auto-reload-interval` in `spec.config` to have automatic reload of certificates (for example certificates auto-renewed from cert-manager) without a restart. it requires Valkey `9.1.0` or newer; on older images the directive is ignored and a `ConfigurationWarning` condition is emitted.
```
bjosv, line 289: nit:
```suggestion
Set `tls-auto-reload-interval` in `spec.config` to have automatic reload of certificates (for example certificates auto-renewed from cert-manager) without a restart. It requires Valkey `9.1.0` or newer; on older images the directive is ignored and a `ConfigurationWarning` condition is emitted.
```

internal/controller/status_test.go
```diff
@@ -57,3 +59,67 @@ func TestSetCondition(t *testing.T) {
...
+	cond = meta.FindStatusCondition(cluster.Status.Conditions, valkeyiov1alpha1.ConditionConfigurationWarning)
+	g.Expect(cond).NotTo(BeNil())
+	g.Expect(cond.Reason).To(Equal(valkeyiov1alpha1.ReasonUnsupportedConfigDirective))
+	g.Expect(cond.Message).To(Equal("another directive warning; directive warning"))
+	g.Expect(cond.Status).To(Equal(metav1.ConditionTrue))
+
+	// Reapplying the exact same warning set should keep the condition stable.
+	before := meta.FindStatusCondition(cluster.Status.Conditions, valkeyiov1alpha1.ConditionConfigurationWarning)
+	r.applyConfigurationWarnings(ctx, cluster, []configWarning{
+		{reason: valkeyiov1alpha1.ReasonGracePeriodTooShort, message: "grace warning"},
```
bjosv, line 114: Hum, this is not the same as the config on line 101 and 102? But it works anyway? The message should be different

internal/controller/utils.go
bjosv: Great!

test/e2e/valkeycluster_tls_test.go
```diff
@@ -268,7 +295,124 @@ spec:
...
+				g.Expect(err).NotTo(HaveOccurred())
+
+				for _, c := range cr.Status.Conditions {
+					g.Expect(c.Type).NotTo(Equal(valkeyiov1alpha1.ConditionConfigurationWarning),
+						"the image is still pre-9.1, but a directive the user never set must not be reported as unsupported")
+				}
+			}).Should(Succeed())
+		})
+
+		It("drops the gated directive from the rendered config", func() {
```
bjosv, line 410: I think we should move this part to before `It("stops warning once the gated directive is removed from spec.config"`, i.e. check that the config is removed by the feature, not by removing the config.

melancholictheory:
flagging the `ConfigurationWarning` collision greptile caught (controller.go:180), since it's the condition i added in #271 and the reuse here trips over how it's keyed.

`meta.SetStatusCondition` keys by condition type, so the version-gate path setting `ConfigurationWarning`/`UnsupportedConfigDirective` overwrites the `GracePeriodTooShort` one set just above when a cluster hits both. it's worse on the clear side: my #271 code clears with `removeConditionIfReason(ConfigurationWarning, GracePeriodTooShort)`, which no-ops once the reason is `UnsupportedConfigDirective`, and the version-gate clear has the mirror problem. so the two concerns silently stomp each other's warning and each other's clear.

a single condition type with a single reason field can't hold two independent warnings, which is the real lesson of reusing it. two ways out:

- roll-up: compute `ConfigurationWarning` once, from all the configuration concerns together (grace period + version gating + whatever's next), set it True with a combined message and remove it when none fire. keeps the "one config-health signal" intent, at the cost of a reason that represents "one or more" (generic reason, detail in the message).
- separate types: `GracePeriodWarning` and `UnsupportedConfigDirective` as distinct conditions, each set and cleared independently. more idiomatic, at the cost of the extra type.

i'd lean roll-up to honor the reuse, but either fixes it. the thing to avoid is two independent `setCondition(ConfigurationWarning, ...)` writers keyed by type.

separately, greptile's config.go:177 one is the point from my earlier comment: `versionGateConfigWarnings` iterates every gated directive without checking `spec.Config`, so the default 9.0.0 image warns `UnsupportedConfigDirective` even when the user set nothing. it should only warn for directives the user actually put in `spec.config` and that got dropped, otherwise every pre-9.1 cluster carries a warning it can't act on.

internal/controller/config.go
```diff
@@ -44,8 +46,18 @@ const (
...
+
+	tlsAutoReloadIntervalKey = "tls-auto-reload-interval"
 )
 
+var minTLSAutoReloadVersion = semver.MustParse("9.1.0")
+
+// versionGatedConfig maps user-facing config directives to the minimum Valkey
+// version that understands them.
+var versionGatedConfig = map[string]*semver.Version{
+	tlsAutoReloadIntervalKey: minTLSAutoReloadVersion,
```
jdheyburn, line 58: nit: Would it make sense to remove the variable assign? I'm thinking what this looks like when we add more gated-configs in the future.

```suggestion
	tlsAutoReloadIntervalKey: semver.MustParse("9.1.0"),
```

internal/valkey/version.go
```diff
@@ -0,0 +1,79 @@
...
+	tag := image[idx+1:]
+	if tag == "" || strings.Contains(tag, "/") {
+		return "", false
+	}
+	return tag, true
+}
+
+// VersionStringFromImage returns the parsed version string or an empty string
+// when the version cannot be determined.
+func VersionStringFromImage(image string) string {
```
jdheyburn, line 62: Is this needed? It's only called from a test.

internal/controller/status.go
```diff
@@ -38,3 +44,53 @@ func removeConditionIfReason(conditions *[]metav1.Condition, condType, reason st
 		meta.RemoveStatusCondition(conditions, condType)
 	}
 }
+
+// configWarning is a entry that represents a configuration warning with a reason and message.
```
jdheyburn, line 48: nit: `a entry` -> `an entry`

But maybe this rewords it better:

```
// configWarning represents a configuration warning with a reason and message.
```

jdheyburn (commented):
There is a potential race condition that goes like this:

1. User adds a version gated config, and bumps the image to a version that supports that config
2. ConfigMap is generated with the version gated config, this propagates to all pods
3. Operator starts to roll ValkeyNodes
4. Another ValkeyNode that has not yet had its spec updated, is restarted for whatever reason - its Spec.Image is still the old one
5. Restarted ValkeyNode is crash-loop-back-off because it cannot read the version gated config

Therefore I think we should include docs to say:
- Update image in one change, wait for roll to complete
- Add the version gated config in another change

Outside of that, I'm thinking how we can extend this functionality beyond configurations, such the operator gating features, etc., based on what version is being run.

internal/controller/status_test.go
sandeepkunusoth: in this case, applyConfigurationWarnings is just passthrough its taking warning string as input. its not depending on actual warning emitted in other function.

jdheyburn:
Are we blocking the unsupported config anywhere? I saw that we're updating the ConfigMap then bubbling that warning to the user. If the container were to restart for whatever reason like a node drain, it would try to read the unsupported config and crash. Would it be better to block that config from being written and bubble the warning there?

Outside of that, it will be useful to document how a developer can implement version gating into the [development guide](https://github.com/valkey-io/valkey-operator/blob/main/docs/developer-guide.md).

melancholictheory:
Blocking the write instead of warning after it does not close that window, because both paths read the same signal. `gatedUserKeysToSuppress` and `versionGateConfigWarnings` both gate on `cluster.Spec.Image`, so whichever way the decision goes, it was made against the version the user asked for and not the version of the pod that reads the file. The node that crashes in your scenario is one whose pod never saw that image.

The ConfigMap is a single object shared by every node in the cluster (`valkey-<cluster>`, handed to each ValkeyNode as `ServerConfigMapName`), so the only value that is safe to gate on is the lowest version that can read it: the minimum across the desired image and everything still running the old one.

Getting that number is the part that needs a decision, because no field holds it today. `buildClusterValkeyNode` copies `cluster.Spec.Image` onto every ValkeyNode on each reconcile, and the StatefulSet template update is deferred until `Spec.WorkloadRevision` matches. So mid-roll every node's spec already reads 9.1 while its pod is still on 9.0. Neither `cluster.Spec.Image` nor `node.Spec.Image` is the running version, and `ValkeyNodeStatus` carries no image or version at all.

If the gate reads an observed floor instead, the sequence converges without asking anything of the user. Mid-roll the floor is still the old version, the directive stays out of the ConfigMap, and a node drained at the wrong moment restarts on a config it can parse. Once every pod is on the new image the floor rises, the directive lands on the next reconcile, and the resulting ConfigMap change rolls nodes that can all parse it. One spec change from the user, no ordering to document, which would retire the two-step note you suggested in your earlier review.

That also gives the `ValkeyVersion` status field the job you were asking about back in July: it has a clear purpose if it holds the observed floor rather than the version parsed from the desired image. `unknown` for floating and digest-pinned tags then reads as "no floor can be established", which is the same conservative skip already implemented.

melancholictheory:
Two notes while this is in review.

The new e2e block collides with #377. It applies

```yaml
networking:
  tls:
    certificate:
      secretName: ...
```

and #377 renames that to `certificates.server` in the same file, with `certificates` marked required. The awkward part is that the two changes sit in different regions of `valkeycluster_tls_test.go`, so git will most likely merge them without reporting a conflict and leave the new block sitting on the removed field path. `certificate` then prunes as unknown, `certificates` is absent, and the apply is rejected, so it shows up as a CI failure after the merge rather than at it. Worth a rebase onto whichever lands first.

On CodeRabbit's `config.go` note about naming the image source: that is already handled on the current head. `versionGateConfigWarnings` sets `imageSource` to "default image" when `cluster.Spec.Image` is empty and builds the message from it, so the warning does not claim the version came from `spec.image`. The part still open is the test case it suggests, an empty `spec.image` with an explicit `tls-auto-reload-interval`, asserting the warning names the default image.

jdheyburn:
#377 is merged in now, so you should be good to merge and resolve the conflict

internal/valkey/version.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, false
+	}
+
+	return semver.New(parsed.Major(), parsed.Minor(), parsed.Patch(), "", ""), true
+}
+
+// imageTag extracts the tag from a container image reference.
+func imageTag(image string) (string, bool) {
+	if image == "" || strings.Contains(image, "@") {
```
bjosv: The following image name is a valid one, K8s validates the hash and it also contains a version tag. We would be able to extract 9.1.1 from this.

`  image: "valkey/valkey:9.1.1@sha256:70739f85ad2ee01a726a965584a0f94895f01b0c60b3cc8b0aeef11eaa6888cf"`

internal/controller/config.go
```diff
@@ -44,8 +46,16 @@ const (
...
 	// Average-ish length of Valkey parameter + value
 	averageParameterLength = 20
+
+	tlsAutoReloadIntervalKey = "tls-auto-reload-interval"
 )
 
+// versionGatedConfig maps user-facing config directives to the minimum Valkey
+// version that understands them.
+var versionGatedConfig = map[string]*semver.Version{
+	tlsAutoReloadIntervalKey: semver.MustParse("9.1.0"),
```
bjosv:
```suggestion
	"tls-auto-reload-interval": semver.MustParse("9.1.0"),
```

and remove `tlsAutoReloadIntervalKey` since its not used then

docs/valkeycluster.md
```diff
@@ -143,6 +143,18 @@ When omitted, the operator picks a safe value: the larger of the Kubernetes defa
 
 An explicit value is honoured as-is, even if it is below the recommended minimum. In that case the operator sets a `ConfigurationWarning` condition (reason `GracePeriodTooShort`) on the `ValkeyCluster` and emits an event when the cluster first enters that state, rather than silently overriding the value. The value must be a positive integer; the CRD rejects zero or negative values.
 
+### Version-gated config
```
bjosv: Shouldn't we have this info under the Config section? At least the instructions how to lift the valkey version vs. version specific configs is important.

Alt. a short text under Config with a link to this section would also do.

bjosv (commented):
Looks good as initial support.

Its good that you documented the issue regarding pod restarts during a config roll. Maybe we should do the move from having a common ConfigMap to having one for each ValkeyNode (but keep common scripts). Something to investigate in another issue/pr.

jdheyburn:
> Maybe we should do the move from having a common ConfigMap to having one for each ValkeyNode (but keep common scripts). Something to investigate in another issue/pr.

@bjosv I think this has come up enough to warrant an issue for us to track it.

melancholictheory:
Raised as #404. It records the mechanism from the review here, why per-node rendering removes the gate rather than making it more accurate, and the object-count cost.

One thing worth weighing before this PR lands: if the fix here goes the observed-minimum-version route, that machinery is thrown away once configs are per node. Deciding the direction first would save building it twice.

jdheyburn:
We still require version gating as there are other issues that would benefit from it like #262. So this PR would still be beneficial, though I think that the inclusion of the tls- config has conflated the scope.

melancholictheory:
Fair, I overstated that. Version detection outlives per-node ConfigMaps, and #262 is a good example: a CRD field gated on 9.1 never touches `valkey.conf`, so #404 does nothing for it.

Worth noting where #262 does land though. `db=` rules are rendered into the aclfile, and that is also one object per cluster, the internal ACL Secret mounted by every node. So it inherits the same problem through a different door: the operator writes the file from the desired version while some nodes still run the old binary. Per-node ConfigMaps would not help there either. What both cases want is the same answer to which version is actually running.

The failure modes differ, which is the useful part. An unrecognised `valkey.conf` directive stops the process, so the node crash-loops. An unrecognised aclfile rule fails the `ACL LOAD`, which once #400 lands surfaces as `Degraded` with reason `ACLApplyFailed` rather than passing unnoticed.

On scope, the line you are pointing at looks clean to me. Detecting the version and reporting a `ConfigurationWarning` is the reusable half and has no dependency on any of this. `tls-auto-reload-interval` is the single entry in `versionGatedConfig` whose safety actually depends on how the shared ConfigMap question is settled. Splitting there lands the part everything else needs without waiting.

melancholictheory (commented):
Ran the failure this gate prevents rather than taking it on trust. Setting `tls-auto-reload-interval: "3600"` on a 9.0.0 image gives:

```
*** FATAL CONFIG FILE ERROR (Version 9.0.0) ***
Reading the configuration file, at line 2
>>> 'tls-auto-reload-interval 3600'
Bad directive or wrong number of arguments
```

One node down, cluster to `Reconciling`, and the roll stops there rather than carrying the mistake to the rest. Removing the directive again recovered the cluster on its own in about 105 seconds. So the problem is real and the remediation completes without help, which is worth knowing since it means the gate is preventing an outage rather than papering over one.

Two things on scope and one detail.

The gate covers user directives only. `gatedUserKeysToSuppress` filters `spec.config`, while `getBaseConfig` is untouched and renders unconditionally. That holds together today because the baseline is 9.0 and everything in the base map is 9.0-safe, but #263 frames the gap as features needing a newer minor than the baseline, and the first operator-owned 9.1 directive would go straight past this. Worth saying whether that is deliberate for the initial framework.

Release candidates are flattened to GA. `imageVersionRE` goes to the trouble of matching `-rcN`, and then `semver.New(parsed.Major(), parsed.Minor(), parsed.Patch(), "", "")` discards it, so `9.1.0-rc1` gates exactly like `9.1.0`. Keeping the prerelease would make an RC compare below GA, dropping `-rcN` from the regex would make RC tags read as undeterminable, and either is defensible. Silently treating them as GA is the one option that is not stated anywhere.

The doc line telling people to apply the image change and let it roll before adding a gated directive is the honest mitigation for the shared ConfigMap window, and worth keeping prominent.

internal/valkey/version.go
```diff
@@ -0,0 +1,84 @@
...
+
+// imageTag extracts the tag from a container image reference.
+func imageTag(image string) (string, bool) {
+	if image == "" {
+		return "", false
+	}
+
+	if i := strings.IndexByte(image, '@'); i >= 0 {
+		image = image[:i]
+	}
```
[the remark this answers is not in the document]
sandeepkunusoth: following image is valid as per this comment https://github.com/valkey-io/valkey-operator/pull/307#discussion_r3842043090 review

sandeepkunusoth:
> Ran the failure this gate prevents rather than taking it on trust. Setting `tls-auto-reload-interval: "3600"` on a 9.0.0 image gives:
> 
> ```
> *** FATAL CONFIG FILE ERROR (Version 9.0.0) ***
> Reading the configuration file, at line 2
> >>> 'tls-auto-reload-interval 3600'
> Bad directive or wrong number of arguments
> ```
> 
> One node down, cluster to `Reconciling`, and the roll stops there rather than carrying the mistake to the rest. Removing the directive again recovered the cluster on its own in about 105 seconds. So the problem is real and the remediation completes without help, which is worth knowing since it means the gate is preventing an outage rather than papering over one.

On Valkey 9.0.0, `tls-auto-reload-interval` in `valkey.conf` is a hard parse failure (`FATAL CONFIG FILE ERROR`), which causes the server to crash-loop and the cluster to remain in `Reconciling`. 
With version gating, the operator drops unsupported user-set directives from the rendered `valkey.conf`, sets `ConfigurationWarning / UnsupportedConfigDirective`, and the cluster still reaches `Ready`. The warning does not block `Ready`.

The E2E test now verifies this behavior on a pre-9.1 image: `status.state == Ready`, `spec.config.tls-auto-reload-interval` remains set, the warning is present.


> Two things on scope and one detail.
> 
> The gate covers user directives only. `gatedUserKeysToSuppress` filters `spec.config`, while `getBaseConfig` is untouched and renders unconditionally. That holds together today because the baseline is 9.0 and everything in the base map is 9.0-safe, but #263 frames the gap as features needing a newer minor than the baseline, and the first operator-owned 9.1 directive would go straight past this. Worth saying whether that is deliberate for the initial framework.


Agreed, this is deliberate for the initial framework. `gatedUserKeysToSuppress` filters only `spec.config`, while `getBaseConfig` remains unchanged and renders unconditionally. The current base configuration is 9.0-compatible, so this is safe for now. If we introduce operator-owned directives that require 9.1+, we should extend the gating framework to cover those as well.


> Release candidates are flattened to GA. `imageVersionRE` goes to the trouble of matching `-rcN`, and then `semver.New(parsed.Major(), parsed.Minor(), parsed.Patch(), "", "")` discards it, so `9.1.0-rc1` gates exactly like `9.1.0`. Keeping the prerelease would make an RC compare below GA, dropping `-rcN` from the regex would make RC tags read as undeterminable, and either is defensible. Silently treating them as GA is the one option that is not stated anywhere.

Agreed. I updated `VersionFromImage` to return the parsed semver directly (`return parsed`) instead of reconstructing it with `semver.New(parsed.Major(), parsed.Minor(), parsed.Patch(), "", "")`, so the prerelease information is preserved. For `tls-auto-reload-interval`, the minimum supported version is now `9.1.0-rc1`, and semver comparison correctly treats prereleases as lower than the corresponding GA release. Unit tests have been updated accordingly.


> The doc line telling people to apply the image change and let it roll before adding a gated directive is the honest mitigation for the shared ConfigMap window, and worth keeping prominent.

updated docs.

internal/controller/config.go
```diff
@@ -46,6 +48,12 @@ const (
 	averageParameterLength = 20
 )
 
+// versionGatedConfig maps user-facing config directives to the minimum Valkey
+// version that understands them.
+var versionGatedConfig = map[string]*semver.Version{
+	"tls-auto-reload-interval": semver.MustParse("9.1.0-rc1"),
```
bjosv, line 54: How do we know which version introduced a specific config, is there source similar to commands?
E.g. `valkey repo/src/commands/x.json`. These only uses full version like `"since": "9.2.0"`

bjosv (dismissed):
LGTM.

Another issue might be if valkey adds additional enum-options to enum configs in newer version.
It looks like a older valkey version that don't know about that option will crash with `*** FATAL CONFIG FILE ERROR ***` We probably need to improve user docs closer to GA.

melancholictheory (approved):
Re-reviewed at `117c54c`. All three points from the previous pass are addressed.

The scope note is in the docs, and it says the right thing: gating covers `spec.config` only, and a future operator-owned directive needing a newer minor has to be gated where it is added.

`VersionFromImage` returns the parsed version now, so the prerelease survives, and the minimum moved to `9.1.0-rc1` with a test for it.

I checked that the minimum is the correct one rather than just a mechanical change. `tls-auto-reload-interval` is present in the `9.1.0-rc1` tag, and in `rc2` and GA, so rc1 is the earliest release that understands the directive and the gate admits exactly the right set.

Build and `make test` are green on this head.

internal/controller/config.go
sandeepkunusoth: I checked the Valkey source, Valkey has a machine-readable `since` field for commands, but there does not appear to be an equivalent source for individual `valkey.conf` directives.


For configuration directives, however, they are registered directly in src/config.c. For example, tls-auto-reload-interval is defined as:
```
createIntConfig("tls-auto-reload-interval", NULL, MODIFIABLE_CONFIG,
                0, INT_MAX,
                server.tls_ctx_config.auto_reload_interval,
                0, INTEGER_CONFIG, NULL, applyTlsCfg),
```

need to revisit on this if there is better way. opened 1 issue.

sandeepkunusoth:
> LGTM.
> 
> Another issue might be if valkey adds additional enum-options to enum configs in newer version. It looks like a older valkey version that don't know about that option will crash with `*** FATAL CONFIG FILE ERROR ***` We probably need to improve user docs closer to GA.

I opened follow-up issues for these cases:

#422 — Prevent Valkey FATAL CONFIG FILE ERROR when spec.config exceeds image version
#423 — Guard unknown enum config values on older Valkey images

sandeepkunusoth:
@bjosv can u please review again

internal/controller/config.go
bjosv: I now heard that there is a new command in Valkey 9.2 `CONFIG INFO`, but it wont help (no since info)

bjosv (approved):
Thanks!

internal/controller/config_test.go
```diff
@@ -114,6 +115,162 @@ var _ = Describe("Live config", Label("liveconfig"), func() {
...
+	It("versionGateConfigWarnings says so when the detected version cannot be determined", func() {
+		cluster := newTLSCluster("valkey/valkey:latest", map[string]string{
+			"tls-auto-reload-interval": "3600",
+		})
+		warnings := versionGateConfigWarnings(cluster)
+		Expect(warnings).To(HaveLen(1))
+		Expect(warnings[0].message).To(ContainSubstring("no version could be detected from spec.image \"valkey/valkey:latest\""))
+	})
+
+	It("versionGateConfigWarnings names the default image when spec.image is empty", func() {
```
bjosv, line 229: This test fails when changing the default image in #326, should we remove it in #326?
Maybe you have some other ideas?

sandeepkunusoth: yes this test needs to be removed for newer versions of default image or we can update it to mock versionConfig

```

	It("versionGateConfigWarnings names the default image when spec.image is empty", func() {
		versionGatedConfig["mock-higher-version-config-key"] = semver.MustParse("99.0.0")
		DeferCleanup(func() {
			delete(versionGatedConfig, "mock-higher-version-config-key")
		})

		cluster := newTLSCluster("", map[string]string{
			"mock-higher-version-config": "mock-value",
		})
		warnings := versionGateConfigWarnings(cluster)
		Expect(warnings).To(HaveLen(1))
		Expect(warnings[0].message).To(ContainSubstring("default image"))
		Expect(warnings[0].message).To(ContainSubstring(DefaultImage))
		Expect(warnings[0].message).NotTo(ContainSubstring("spec.image"))
	})
```

PR #317, opened 2026-07-15
fix: stop updating workloads on every reconcile by owning API-server defaults

### LINKED ISSUE #315: ValkeyNode reconcile updates StatefulSets on every pass (fighting API-server defaults) — audit-log noise at ~6 events per reconcile
## Symptom

On a 3-shard / 6-node `ValkeyCluster` (operator v0.3.0, GKE), every reconcile pass emits a StatefulSet **update** for all six nodes — surfaced in GKE as a continuous stream of `io.k8s.apps.v1.statefulsets.update` audit events (~1-2k/hour), with `metadata.generation` stuck at 1 (no real spec change ever) while `resourceVersion` churns.

Pods are unaffected (rollouts key off generation), but the cost is audit-log volume/billing, wasted API-server writes, and real events getting buried.

## Root cause

`ensureStatefulSet` (internal/controller/valkeynode_controller.go) uses `controllerutil.CreateOrUpdate` — which does skip no-op updates — but its mutate function **wholesale-replaces the live spec with a freshly built one**:

```go
result, err := controllerutil.CreateOrUpdate(ctx, r.Client, sts, func() error {
    sts.Labels = desired.Labels
    sts.Spec = desired.Spec            // ← zeroes every API-server-defaulted field
    ...
})
```

`buildValkeyNodeStatefulSet` never sets `podManagementPolicy`, `revisionHistoryLimit`, `updateStrategy`, template `imagePullPolicy`, `dnsPolicy`, `schedulerName`, probe defaults, etc. The live object has those **defaulted by the API server** (`OrderedReady`, `10`, `IfNotPresent`, `ClusterFirst`, `default-scheduler`, …). So every mutate produces object ≠ live → CreateOrUpdate issues an Update → the apiserver re-applies defaults → next reconcile diffs again. The controller is perpetually fighting the defaulter; desired == actual yet it never goes quiet.

`ensureDeployment` and (from reading) the Service upsert share the pattern.

## Suggested fixes (either resolves it)

1. **Server-Side Apply** for the ensure* functions (`client.Apply` patch with field ownership): the apiserver computes the diff against owned fields only; a no-op apply does not bump resourceVersion. Cleanest long-term.
2. **Create-vs-update split in the mutate fn**: on `sts.CreationTimestamp.IsZero()` set the full spec; on update copy only operator-owned mutable fields (replicas / template / updateStrategy) — though template replacement still fights template-level defaults, so (1) is the robust option.

## Repro

Any ValkeyCluster; watch `kubectl get sts <node> -w -o custom-columns=RV:.metadata.resourceVersion,GEN:.metadata.generation` — RV bumps on each reconcile pass while GEN stays 1; or observe `statefulsets.update` audit events.

Happy to send a PR for option (1) if that direction is acceptable to maintainers.

matka12:
Fixes #315.

## Problem

`ensureStatefulSet` / `ensureDeployment` assign the whole desired spec inside their `CreateOrUpdate` mutate functions, while the builders leave every API-server-defaulted field unset. Each reconcile clobbers the stored defaults back to zero values, `CreateOrUpdate` sees a diff, and the operator issues an Update — every pass, forever, with `metadata.generation` stuck. Observed impact: ~1–2k `statefulsets.update` Admin Activity audit events/hour on a 3-shard/6-node cluster; on GKE these land in the non-excludable `_Required` bucket (details in #315).

This implements the **narrow fix** discussed in the issue thread (own the defaults in the builders), as suggested there — small and backportable; a Server-Side Apply migration can follow separately.

## Changes

**`buildValkeyNodePodTemplateSpec` / `buildContainersDef`** — set the pod-template fields the API server would default, so the built template already equals the stored one:
- pod: `restartPolicy`, `dnsPolicy`, `schedulerName`, `terminationGracePeriodSeconds` (30 when unset), empty `securityContext` when none is configured, volume `defaultMode`s
- containers (applied after `mergePatchContainers`, so user container patches are normalized the same way the API server would): `imagePullPolicy` (tag-aware, mirroring API-server logic), `terminationMessagePath`/`Policy`, port `protocol`, env `fieldRef.apiVersion`

**`buildValkeyNodeStatefulSet`** — `podManagementPolicy`, `updateStrategy` (RollingUpdate, partition 0), `revisionHistoryLimit`, `persistentVolumeClaimRetentionPolicy` (Retain/Retain; note: requires the `StatefulSetAutoDeletePVC` gate, on by default since 1.27, GA 1.32 — on older clusters with the gate off the API server drops the field and this one field would resume churning).

**`buildValkeyNodeDeployment`** — `revisionHistoryLimit`, `progressDeadlineSeconds`.

**`upsertService`** — the rebuilt `ports` slice now sets `protocol` and `targetPort`, which made the headless Service churn the same way.

The PDB reconcile already mutates field-wise with values that round-trip cleanly, so it needed no change.

## Tests

- New envtest regression test: reconcile a ValkeyNode, capture the StatefulSet `resourceVersion`, reconcile 3 more times, assert it is unchanged. Fails on `main`, passes here (envtest runs a real API server, so real defaulting is exercised).
- `TestBuildValkeyNodePodTemplateSpec_PodSecurityContext_NilIsNoop` updated: omitting the field now yields the empty `securityContext` the API server defaults to — semantically identical (no security settings applied), and required for the no-op property.
- `make test` green (75/75 specs + unit tests, envtest k8s 1.35).

## Caveat

This is intentionally the treadmill-y narrow fix: a future Kubernetes version defaulting a new field would reintroduce churn for that field until added here. The durable fix is SSA with generated ApplyConfigurations, per the discussion in #315.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


melancholictheory:
this is the narrow fix from the thread done thoroughly, nice. owning the defaults across the pod template, containers, volumes, and the service ports (not just the STS) is the right scope, and the envtest regression is the part that actually keeps it fixed, since the failure mode here is silent drift, a hand-maintained mirror is only safe if something fails loudly when the api server adds or changes a default.

two things to close it out:

the PDB reconcile isn't in the diff. after #308 it builds a PodDisruptionBudget spec, and if it assigns the whole spec the same way the workloads did, it'll churn on anything the api server defaults there (`unhealthyPodEvictionPolicy` is the one that bites on newer clusters). worth confirming it's clean or folding it in, since the goal is consistency across all owned resources.

the `imagePullPolicy` mirror is the easiest one to get subtly wrong. the api server's rule is Always for a `latest` tag or no tag, IfNotPresent otherwise (including digests). if `defaultImagePullPolicy` diverges on any of those, that image churns silently and the envtest won't catch it unless the case is in the table. worth a fixture per branch (plain tag, latest, no tag, digest).

structurally this is the treadmill i mentioned on #272: every default is now a hand-mirror, so a new defaulted field in a future k8s reintroduces churn. fine as the fix for today, and the envtest guard is what makes it maintainable, but it's still the argument for SSA owning a field set eventually rather than chasing defaults by hand.

matka12:
Thanks for the careful read — both points addressed in 19822d6.

**imagePullPolicy**: you were right that this was the easy one to get subtly wrong — my first cut checked the digest before the tag, which diverges from the API server on `repo:latest@sha256:...` (the server resolves the effective tag first, so an explicit `latest` tag defaults to `Always` even alongside a digest; only a digest-*only* reference gets `IfNotPresent`). Rewrote `defaultImagePullPolicy` to resolve the tag the way `parsers.ParseImageName` does and added a table test pinning every branch: plain tag, `latest`, no tag, digest-only, tag+digest, latest+digest, and registry-port variants (`reg:5000/img` with/without tag/digest).

**PDB**: it's clean — `reconcilePodDisruptionBudget` assigns `maxUnavailable`/`minAvailable`/`selector` individually and never replaces the whole spec, so anything the API server defaults on the stored object (`unhealthyPodEvictionPolicy` included) survives the mutate untouched. Rather than leave that as an assertion in a PR comment, I pinned it: an envtest regression asserts the PDB `resourceVersion` is stable across repeated reconciles, and I added the same guard for the headless Service upsert. So all three owned resource kinds (STS/Deployment via the existing test, Service, PDB) now fail loudly if a future defaulting change reintroduces churn.

Agreed on the treadmill — this is the backportable fix for today, and the no-churn regressions are what make the hand-mirror maintainable until an SSA + ApplyConfigurations migration owns the field set properly. Happy to take a run at that as a follow-up PR if maintainers are open to it.

melancholictheory:
that digest-vs-tag ordering is exactly the trap, and your reading matches the upstream defaulting: it resolves the name first and only checks the tag, so `repo:latest@sha256:...` keeps the explicit `latest` tag and defaults Always, a digest-only ref leaves the tag empty and falls to IfNotPresent, and a bare `repo` gets the tag filled in as latest and goes back to Always. the table pinning all six is the right way to hold it.

the resourceVersion-stability regressions are a better guard than what i suggested. asserting "reconcile twice, resourceVersion doesn't move" catches the whole class rather than one field, so a future defaulted field fails the test instead of silently churning. worth keeping that shape if the SSA migration lands too, it's the same invariant.

on the follow-up: jdheyburn already said on #315 he wants a consistent approach across all resources, so there's appetite. the one thing that decides whether it delivers is the sparse-apply part from #272: a typed `client.Apply` still sends zero-valued fields as intent, so without generated ApplyConfigurations (or an unstructured patch of owned fields) you get the same overwrite with a different verb and the churn comes back. if you take it, that's the piece to land first.

matka12:
Agreed on all counts. I'll keep the resourceVersion-stability regressions as the invariant regardless of mechanism — they assert the outcome (steady-state reconciles don't write), not the implementation, so they carry over to SSA unchanged.

For the follow-up I'll take it in that order: generated ApplyConfigurations for the owned resources first (the sparse-apply piece — without it a typed `client.Apply` sends zero-valued fields as intent and just reproduces the overwrite under a different verb), then swap the reconcilers to SSA with a fixed field owner, with these same regressions as the acceptance gate. Will reference #315/#272 and this thread when I open it.

jdheyburn:
@matka12 I am fine with this being an interim fix that we can get out for 0.4 in a few days. If you are wanting to contribute to migrating everything to SSA that would be appreciated - though it may undo some of the work you've done here. We can aim for 0.5 for that in a couple of weeks.

You may come across the same issue with ValkeyNode reconciliation too. Are you able to run a test through there too?

- https://github.com/valkey-io/valkey-operator/blob/1f741666d8327498db5e09caf90ab72cfbe564e9/internal/controller/valkeycluster_controller.go#L648

matka12:
0.4 as the interim works great for us — we're carrying this patch in production and would love to drop it for the release tag. On the SSA migration for 0.5: I'd like to take it if time permits — I'll open a tracking issue once this lands either way, so the 0.5 plan isn't blocked on my availability. Fully expected that SSA supersedes parts of this diff — the piece that carries over is the resourceVersion-stability regressions, which assert the outcome (steady-state reconciles don't write) rather than the mechanism.

On the ValkeyNode reconciliation path (`reconcileValkeyNode` → the `CreateOrUpdate` you linked): it's covered by this PR's envtest — `"does not update ValkeyNodes when spec is unchanged"` in `valkeycluster_controller_test.go` drives that exact code path twice against a real envtest API server (with CRD schema defaulting active) and asserts every ValkeyNode's resourceVersion is unchanged on the second pass. Just re-ran the suite on this branch: green.

Why that path was never in the churn loop despite the same `node.Spec = desired.Spec` shape: `buildClusterValkeyNode` copies its fields from the *already-stored* ValkeyCluster — which the API server defaulted at admission — and the two CRDs share their schema defaults (`workloadType: StatefulSet` on both; nested defaults like `persistence.volumeReclaimPolicy` ride along the same way). So the copy always writes exactly what the node's own defaulting would produce. The STS/Deployment case broke precisely because there the builder's source was *not* post-defaulting (it left native-typed fields unset for the server to fill). Matches what we observed in prod: constant `statefulsets.update` audit events, zero `valkeynodes.update`.

Happy to make any final adjustments to get this into 0.4.

jdheyburn (approved):

PR #316, opened 2026-07-15
fix!: Rename go module to github.com/valkey-io/valkey-operator

deepakpunjabi:
This PR closes https://github.com/valkey-io/valkey-operator/issues/312.

### Summary

Renamed the Go module to github.com/valkey-io/valkey-operator.

### Features / Behaviour Changes

- The module can now be resolved with `go get github.com/valkey-io/valkey-operator`, so it can be imported as a dependency by downstream projects.
- Breaking change for importers: the import path changes from `valkey.io/valkey-operator` to `github.com/valkey-io/valkey-operator`. Consumers must update their imports (and any replace directives). No runtime behaviour changes.
- The previous path valkey.io/valkey-operator never resolved via go get — valkey.io serves no go-get meta tag (returns 404), so real-world breakage is expected to be minimal.

### Implementation

- go.mod: updated the module path via go mod edit -module, followed by go mod tidy.
- PROJECT: updated the repo and api.path fields.
- Rewrote the import path in all 34 affected .go files.
- Intentionally left unchanged (these are independent of the Go module path and must not be renamed):
  - CRD API group remains valkey.io (e.g. `valkeycluster.valkey.io`).
  - Container image path ghcr.io/valkey-io/valkey-operator (OCI registry, not a Go path).
  - Docs / README / issue-template URLs — already point at https://github.com/valkey-io/valkey-operator/....
 
### Limitations

None. This is a mechanical rename with no functional impact.

### Testing

- `go mod tidy` - passes.
- `go build cmd/main.go` - passes.
- `go vet ./...` - passes.
- `go test ./...` - passes.
- `gofmt -l` on changed files — no diffs.
- `pre-commit` - passes.
```
$ pre-commit run --all-files
generate code and manifests..............................................Passed
run go mod tidy..........................................................Passed
run golangci-lint........................................................Passed
Check for merge conflicts................................................Passed
```

### Checklist

Before submitting the PR make sure the following are checked:

- [x] This Pull Request is related to one issue.
- [x] Commit message explains what changed and why
- [x] Tests are added or updated.
- [x] Documentation files are updated.
- [x] I have run pre-commit locally (`pre-commit run --all-files` or hooks on commit)

PR #319, opened 2026-07-16
feat: apply ACL changes live without rolling pods

### LINKED ISSUE #212: [feat] Apply ACL changes live
Today, ACL changes require a pod roll in order for them to be applied on the node.

Similar to #210, any ACL changes that are made should be applied live on the node. We can reuse a similar framework to what is being doing in the PR #209.

melancholictheory:
Closes #212

### Summary

ACL changes only took effect when a pod restarted and reloaded the aclfile. The ValkeyNode controller now applies them in place with `ACL LOAD`, so a password rotation or a permission change no longer needs a roll.

- ACL changes reach the running server without a pod roll, eventually consistent with the mounted Secret.
- New `ACLApplied` condition on `ValkeyNode` reporting whether the server's ACL matches the aclfile Secret.
- No API changes.

### Implementation

Per the direction on #212: `ACL LOAD` of the mounted file at the end of the node reconcile, since that file is already the full desired state and is exactly what a restart would load.

Two things shaped the rest:

`ACL SAVE` is out. The aclfile is mounted read-only from its Secret, so it isn't available, and it isn't needed either: the Secret the cluster controller writes is what a restarting pod loads. So the Secret stays the source of truth for restart, and `ACL LOAD` is purely the live path.

The mounted copy lags. kubelet refreshes it lazily, so an `ACL LOAD` issued right after the Secret changes can read the previous contents and silently do nothing. The operator can't read the pod's file without an exec, so the running server is the source of truth for verification instead: the controller compares the password hashes from `ACL GETUSER` against the aclfile Secret, and only issues a reload when they differ, then re-checks. That makes a premature reload self-correcting, it leaves the node `ACLApplied=False` with reason `PendingPropagation` rather than claiming applied, and the requeue retries once the volume catches up. A steady-state reconcile issues no writes at all, only the comparison.

`ACLApplied` deliberately does not gate the cluster's rolling update the way `LiveConfigApplied` does: a node whose ACL is still propagating keeps serving with its previous credentials, and a roll would load the new file regardless.

### Rotating without an auth gap

Documented in `docs/valkeycluster.md`: a Valkey user can hold several passwords at once, and `passwordSecret.keys` already accepts multiple keys, so adding the new password alongside the old, waiting for `ACLApplied=True`, moving clients over, then dropping the old key keeps every credential valid throughout the window. Replacing the key outright locks out any client still holding the old password until it catches up.

### Testing

- Unit tests for the aclfile parsing (multi-password users, `nopass`, system users, non-user lines) and the in-sync comparison.
- envtest coverage for `applyLiveACL`: no secret, missing secret, already in sync (asserts no `ACL LOAD` is issued), reload that converges, reload against a stale volume (asserts it reports not-synced rather than applied), and a failing reload.
- `make test` and `make lint` pass; no generated drift.

### Checklist

### DIFF
--- a/api/v1alpha1/valkeynode_types.go
+++ b/api/v1alpha1/valkeynode_types.go
@@ -190,6 +190,11 @@ const (
 	// of Spec.Config has been successfully applied via CONFIG SET. The cluster
 	// controller blocks one-at-a-time progress until this condition is True.
 	ValkeyNodeConditionLiveConfigApplied = "LiveConfigApplied"
+	// ValkeyNodeConditionACLApplied indicates that the ACL the cluster controller
+	// wrote to the mounted Secret is live on the server. It is applied via ACL
+	// LOAD without a pod roll, so ACL never enters Spec.WorkloadRevision. True
+	// means the desired user set and their password hashes are live.
+	ValkeyNodeConditionACLApplied = "ACLApplied"
 	// ValkeyNodeConditionWorkloadRollPending indicates a rolling pod-template update
 	// is intentionally deferred: the desired template differs from live, and
 	// Spec.WorkloadRevision has not yet authorized that template. Status True means
--- a/internal/controller/valkeynode_acl.go
+++ b/internal/controller/valkeynode_acl.go
@@ -52,14 +52,14 @@ func desiredUserPasswordHashes(aclFile string) map[string][]string {
 				hashes = append(hashes, h)
 			}
 		}
-		out[fields[1]] = normaliseHashes(hashes)
+		out[fields[1]] = normalizeHashes(hashes)
 	}
 	return out
 }
 
-// normaliseHashes sorts and deduplicates password hashes so both sides of the
+// normalizeHashes sorts and deduplicates password hashes so both sides of the
 // comparison are in the same shape.
-func normaliseHashes(hashes []string) []string {
+func normalizeHashes(hashes []string) []string {
 	slices.Sort(hashes)
 	return slices.Compact(hashes)
 }
@@ -69,7 +69,7 @@ func normaliseHashes(hashes []string) []string {
 // hashes.
 //
 // Permissions are deliberately not compared. ACL GETUSER returns Valkey's
-// normalised rendering of the rules, while the operator only holds the aclfile
+// normalized rendering of the rules, while the operator only holds the aclfile
 // text, so comparing the two would mean reimplementing Valkey's own ACL parser
 // and keeping it in step with the server. Correctness of the apply does not
 // depend on this check either way: the reload is unconditional, so permission
@@ -116,7 +116,7 @@ func aclObservablyInSync(ctx context.Context, c valkeyConfigClient, desired map[
 //
 // The reload is therefore unconditional. The operator cannot read the pod's
 // copy of the file, and it cannot compare the whole ACL against the server
-// either (ACL GETUSER renders Valkey's normalised form, the operator holds
+// either (ACL GETUSER renders Valkey's normalized form, the operator holds
 // aclfile text), so there is no reliable way to know the file is current. A
 // repeated LOAD is idempotent and cheap, and it is what makes every kind of
 // change converge, including permission edits and removed users, once the
--- a/internal/controller/valkeynode_acl.go
+++ b/internal/controller/valkeynode_acl.go
@@ -1,0 +1,157 @@
+/*
+Copyright 2025 Valkey Contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	valkeyiov1alpha1 "github.com/valkey-io/valkey-operator/api/v1alpha1"
+)
+
+// desiredUserPasswordHashes parses the aclfile the ValkeyCluster controller
+// builds into each user's SHA-256 password hashes, keyed by username. The file
+// is a sequence of "user <name> on #<hash> ... " lines (see buildUserAcl), so
+// the hashes are the '#'-prefixed tokens.
+//
+// Hashes are sorted and deduplicated. Order is not guaranteed by either side,
+// and Valkey keeps a user's passwords as a set: pointing two Secret keys at the
+// same password yields a repeated hash in the file but a single one back from
+// ACL GETUSER, which would otherwise never compare equal.
+func desiredUserPasswordHashes(aclFile string) map[string][]string {
+	out := map[string][]string{}
+	for line := range strings.SplitSeq(aclFile, "\n") {
+		fields := strings.Fields(line)
+		if len(fields) < 2 || fields[0] != "user" {
+			continue
+		}
+		hashes := []string{}
+		for _, f := range fields[2:] {
+			if h, ok := strings.CutPrefix(f, "#"); ok {
+				hashes = append(hashes, h)
+			}
+		}
+		out[fields[1]] = normaliseHashes(hashes)
+	}
+	return out
+}
+
+// normaliseHashes sorts and deduplicates password hashes so both sides of the
+// comparison are in the same shape.
+func normaliseHashes(hashes []string) []string {
+	slices.Sort(hashes)
+	return slices.Compact(hashes)
+}
+
+// aclObservablyInSync reports whether the parts of the ACL that can be compared
+// exactly are live on the server: the set of users, and each user's password
+// hashes.
+//
+// Permissions are deliberately not compared. ACL GETUSER returns Valkey's
+// normalised rendering of the rules, while the operator only holds the aclfile
+// text, so comparing the two would mean reimplementing Valkey's own ACL parser
+// and keeping it in step with the server. Correctness of the apply does not
+// depend on this check either way: the reload is unconditional, so permission
+// edits converge regardless. This only scopes what ACLApplied can honestly
+// claim.
+func aclObservablyInSync(ctx context.Context, c valkeyConfigClient, desired map[string][]string) (bool, error) {
+	actualUsers, err := c.UserNames(ctx)
+	if err != nil {
+		return false, err
+	}
+	serverUsers := slices.Sorted(slices.Values(actualUsers))
+	// Valkey always keeps a `default` user of its own, and ACL LOAD cannot
+	// remove it. When the aclfile does not manage `default` (it is absent from
+	// spec.users), ignore the server's copy; otherwise the sets never match and
+	// the node loops forever reporting the ACL as not yet live.
+	if _, managed := desired["default"]; !managed {
+		serverUsers = slices.DeleteFunc(serverUsers, func(u string) bool { return u == "default" })
+	}
+	if !slices.Equal(serverUsers, slices.Sorted(maps.Keys(desired))) {
+		// A user was added or removed and the server has not picked it up yet.
+		return false, nil
+	}
+	for _, user := range slices.Sorted(maps.Keys(desired)) {
+		actual, err := c.UserPasswordHashes(ctx, user)
+		if err != nil {
+			return false, err
+		}
+		if !slices.Equal(actual, desired[user]) {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+// applyLiveACL reloads the mounted aclfile into the running server so ACL
+// changes take effect without a pod roll, and reports whether the desired
+// passwords are live yet.
+//
+// The aclfile is mounted read-only from the Secret, so ACL SAVE is neither
+// available nor needed: the Secret the cluster controller writes is what a
+// restarting pod loads. That leaves ACL LOAD as the live path, with one catch.
+// The projected volume is refreshed lazily by kubelet, so a LOAD fired right
+// after the Secret changes can read the pre-update file and silently no-op.
+//
+// The reload is therefore unconditional. The operator cannot read the pod's
+// copy of the file, and it cannot compare the whole ACL against the server
+// either (ACL GETUSER renders Valkey's normalised form, the operator holds
+// aclfile text), so there is no reliable way to know the file is current. A
+// repeated LOAD is idempotent and cheap, and it is what makes every kind of
+// change converge, including permission edits and removed users, once the
+// volume catches up.
+//
+// The returned bool reports the parts that can be compared exactly, the user
+// set and their password hashes, which is what a rotation needs to wait on.
+func (r *ValkeyNodeReconciler) applyLiveACL(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) (bool, error) {
+	if node.Spec.UsersACLSecretName == "" {
+		return true, nil
+	}
+
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{Name: node.Spec.UsersACLSecretName, Namespace: node.Namespace}
+	if err := r.APIReader.Get(ctx, key, secret); err != nil {
+		if apierrors.IsNotFound(err) {
+			// The cluster controller owns this Secret and has not created it
+			// yet. There is nothing to apply until it exists.
+			return true, nil
+		}
+		return false, fmt.Errorf("get ACL secret %s: %w", key.Name, err)
+	}
+	desired := desiredUserPasswordHashes(string(secret.Data[aclFilename]))
+	if len(desired) == 0 {
+		return true, nil
+	}
+
+	c, err := r.newConfigClient(ctx, r, node)
+	if err != nil {
+		return false, err
+	}
+	defer c.Close()
+
+	if err := c.LoadACL(ctx); err != nil {
+		return false, err
+	}
+	return aclObservablyInSync(ctx, c, desired)
+}
--- a/internal/controller/valkeynode_controller.go
+++ b/internal/controller/valkeynode_controller.go
@@ -115,9 +115,9 @@ func (rc *realValkeyConfigClient) UserPasswordHashes(ctx context.Context, userna
 	if err != nil {
 		return nil, fmt.Errorf("ACL GETUSER %s passwords: %w", username, err)
 	}
-	// Valkey keeps passwords as a set, but normalise anyway so both sides of
+	// Valkey keeps passwords as a set, but normalize anyway so both sides of
 	// the comparison are in the same shape.
-	return normaliseHashes(hashes), nil
+	return normalizeHashes(hashes), nil
 }
 
 func (rc *realValkeyConfigClient) Close() { rc.client.Close() }
@@ -377,9 +377,6 @@ func (r *ValkeyNodeReconciler) ensureWorkload(ctx context.Context, node *valkeyi
 	}
 }
 
-// buildPodTemplateAnnotations assembles the annotations that must be present on
-// the pod template spec to trigger rolling updates when the ACL secret or the
-// server config changes.
 // buildPodTemplateAnnotations assembles the annotations stamped on the pod
 // template, and therefore the ones that feed the WorkloadRevision roll hash.
 // Only the server-config hash lives here: a config change that is not
--- a/internal/controller/valkeynode_controller.go
+++ b/internal/controller/valkeynode_controller.go
@@ -38,7 +38,9 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	valkeyiov1alpha1 "github.com/valkey-io/valkey-operator/api/v1alpha1"
 )
@@ -53,6 +55,15 @@ const (
 // fake (envtest has no running Valkey server).
 type valkeyConfigClient interface {
 	SetConfig(ctx context.Context, params map[string]string) error
+	// LoadACL reloads the mounted aclfile into the running server.
+	LoadACL(ctx context.Context) error
+	// UserNames returns the names of the users the server currently has.
+	UserNames(ctx context.Context) ([]string, error)
+	// UserPasswordHashes returns the user's currently configured password
+	// hashes, sorted and deduplicated. An unknown user yields an empty slice
+	// rather than an error, so a user that has not been loaded yet reads as out
+	// of sync.
+	UserPasswordHashes(ctx context.Context, username string) ([]string, error)
 	Close()
 }
 
@@ -70,6 +81,43 @@ func (rc *realValkeyConfigClient) SetConfig(ctx context.Context, params map[stri
 		return fmt.Errorf("CONFIG SET: %w", err)
 	}
 	return nil
+}
+
+func (rc *realValkeyConfigClient) LoadACL(ctx context.Context) error {
+	if err := rc.client.Do(ctx, rc.client.B().AclLoad().Build()).Error(); err != nil {
+		return fmt.Errorf("ACL LOAD: %w", err)
+	}
+	return nil
+}
+
+func (rc *realValkeyConfigClient) UserNames(ctx context.Context) ([]string, error) {
+	users, err := rc.client.Do(ctx, rc.client.B().AclUsers().Build()).AsStrSlice()
+	if err != nil {
+		return nil, fmt.Errorf("ACL USERS: %w", err)
+	}
+	return users, nil
+}
+
+func (rc *realValkeyConfigClient) UserPasswordHashes(ctx context.Context, username string) ([]string, error) {
+	m, err := rc.client.Do(ctx, rc.client.B().AclGetuser().Username(username).Build()).AsMap()
+	if err != nil {
+		if vclient.IsValkeyNil(err) {
+			// User is not defined on the server yet.
+			return []string{}, nil
+		}
+		return nil, fmt.Errorf("ACL GETUSER %s: %w", username, err)
+	}
+	passwords, ok := m["passwords"]
+	if !ok {
+		return []string{}, nil
+	}
+	hashes, err := passwords.AsStrSlice()
+	if err != nil {
+		return nil, fmt.Errorf("ACL GETUSER %s passwords: %w", username, err)
+	}
+	// Valkey keeps passwords as a set, but normalise anyway so both sides of
+	// the comparison are in the same shape.
+	return normaliseHashes(hashes), nil
 }
 
 func (rc *realValkeyConfigClient) Close() { rc.client.Close() }
@@ -99,6 +147,7 @@ type ValkeyNodeReconciler struct {
 // +kubebuilder:rbac:groups=valkey.io,resources=valkeynodes/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=valkey.io,resources=valkeynodes/finalizers,verbs=update
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
 // +kubebuilder:rbac:groups="apps",resources=statefulsets,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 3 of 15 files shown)


melancholictheory:
good catch, this was a real bug and both repros are right. fixed in 522ec5e.

skipping the reload when the password hashes matched meant exactly what you describe: a permissions-only edit, or a user removed from the aclfile, kept every desired hash intact, so the node skipped `ACL LOAD` and sat on the old ACL while `ACLApplied` claimed True. the PR advertised live permission changes and didn't deliver them.

the comparison can't just be widened, which is what led me down this path in the first place. `ACL GETUSER` returns Valkey's normalised rendering of the rules, while the operator only has the aclfile text, so comparing the two properly would mean reimplementing Valkey's ACL parser and keeping it in step. password hashes are the one part that compares exactly.

so the reload is unconditional now. it's idempotent and cheap, and it converges every kind of change once the mounted volume catches up, which is also what was originally proposed on #212. the hash comparison stays, but only to drive `ACLApplied`, and the condition is now scoped to what it can actually verify: whether the desired passwords are live on that node. that's the signal the rotation procedure waits on, and the docs no longer imply it covers permission edits.

added coverage for the case you found: passwords already matching must still issue the reload.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,131 @@
...
+		if len(fields) < 2 || fields[0] != "user" {
+			continue
+		}
+		hashes := []string{}
+		for _, f := range fields[2:] {
+			if h, ok := strings.CutPrefix(f, "#"); ok {
+				hashes = append(hashes, h)
+			}
+		}
+		slices.Sort(hashes)
```
jdheyburn: Shall we dedupe hashes too? In case the user duplicated passwords? I think Valkey stores hashed passwords as a set on the backend, so we should do the same otherwise we'd be stuck in reconciliation forever.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,131 @@
...
+	c, err := r.newConfigClient(ctx, r, node)
+	if err != nil {
+		return false, err
+	}
+	defer c.Close()
+
+	if err := c.LoadACL(ctx); err != nil {
+		return false, err
+	}
+	return passwordsInSync(ctx, c, desired)
```
jdheyburn: Why do we only check if passwords are in sync, and not other dimensions like users added/removed, permissions updated, etc.?

internal/controller/valkeynode_controller.go
```diff
@@ -71,6 +77,34 @@ func (rc *realValkeyConfigClient) SetConfig(ctx context.Context, params map[stri
...
+	}
+	passwords, ok := m["passwords"]
+	if !ok {
+		return []string{}, nil
+	}
+	hashes, err := passwords.AsStrSlice()
+	if err != nil {
+		return nil, fmt.Errorf("ACL GETUSER %s passwords: %w", username, err)
+	}
+	slices.Sort(hashes)
```
jdheyburn: Same here re deduplicating.

jdheyburn (commented):
Thanks for taking this on, just a couple of comments.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// aclObservablyInSync reports whether the parts of the ACL that can be compared
+// exactly are live on the server: the set of users, and each user's password
+// hashes.
+//
+// Permissions are deliberately not compared. ACL GETUSER returns Valkey's
+// normalised rendering of the rules, while the operator only holds the aclfile
+// text, so comparing the two would mean reimplementing Valkey's own ACL parser
+// and keeping it in step with the server. Correctness of the apply does not
```
jdheyburn, line 74: Are we able to reuse some of the below to make comparisons? 

- https://github.com/valkey-io/valkey-operator/blob/3a92e949218f64ee3a56808b2dc5abc0ceba37b1/internal/controller/users.go#L218-L262

internal/controller/valkeynode_acl.go
melancholictheory: you're right, and it's worse than cosmetic: two Secret keys pointing at the same password put a repeated hash in the aclfile, Valkey reports it once, and the comparison could never converge, so the node would sit at `ACLApplied=False` and requeue every 10s forever. sorted and deduped on both sides now, with a test for the repeated-hash case. 1844eab.

internal/controller/valkeynode_acl.go
melancholictheory: good push, and you're right i drew the line too conservatively. usernames compare exactly, so the user set is part of the check now (via `ACL USERS`): a user added to or removed from the aclfile that the server hasn't picked up reads as out of sync instead of being invisible to the condition.

permission rules are the part i still left out, and it isn't laziness: `ACL GETUSER` returns Valkey's normalised rendering of the rules, while the operator only holds the aclfile text, so comparing them properly means reimplementing Valkey's ACL parser and keeping it in step with the server.

worth stressing none of this gates the apply: the reload is unconditional (Greptile's catch earlier), so permission edits converge regardless. the check only scopes what `ACLApplied` can honestly claim, and the docs now say exactly that. 1844eab.

internal/controller/valkeynode_controller.go
melancholictheory: done, `UserPasswordHashes` normalises through the same helper (sort + compact), so both sides are in the same shape. 1844eab.

internal/controller/valkeynode_acl.go
melancholictheory: i looked at reusing `buildUserAcl` for this, and it's the same wall the passwords-only check ran into. `buildUserAcl` renders the operator's form of the ACL, while `ACL GETUSER` returns Valkey's normalised form (rule order, folded categories, implied defaults like `sanitize-payload`, `resetchannels` handling). they don't compare byte-for-byte, so lining them up would still mean running our string through Valkey's own normaliser, which is the parser reimplementation i was trying to avoid.

where the reuse does help is the exact parts: the user set and the password hashes. those i already take from the aclfile the cluster controller built with `buildUserAcl`, and check against `ACL USERS` / `ACL GETUSER`, so the desired side is the same source of truth, just compared on the dimensions that survive normalisation. and since the reload is unconditional, the rest converges without needing a full comparison at all.

melancholictheory:
gentle nudge on this one, it's been ready since the last review round. CI is green, and the points from that pass are addressed: password hashes are deduped on both sides, and the sync check now covers user membership (add/remove), not just passwords. happy to rebase past the module rename in #316 whenever it's useful, or i'll do it if it goes dirty. no rush, just flagging it's waiting on another look.

jdheyburn:
Sorry for the delay. I just tried to test it in a kind cluster locally, and I got these errors in the logs:

```
2026-07-28T18:40:42Z    DEBUG   events  Failed to apply live ACL: ACL LOAD: NOPERM User _operator has no permissions to run the 'acl|load' command      {"type": "Warning", "object": "nil", "action": "ApplyLiveACL", "reason": "LiveACLApplyFailed"}
```

It doesn't look like you've updated users.go to support the feature. Are you able to add that and perform local testing to see if it works as expected? Thank you!

melancholictheory:
good catch, and thanks for running it on a real cluster. that's exactly the gap: `_operator` had `+config|set` for the live-config path but none of the acl subcommands, so `ACL LOAD` hit NOPERM. the envtest suite uses a fake config client that doesn't enforce ACL permissions, so the miss stayed green in CI.

fixed by granting the three subcommands the feature actually needs on `_operator` in users.go:

```
+acl|load      // reload the aclfile
+acl|getuser   // read back a user's hashes to verify the reload landed
+acl|users     // read the user set to verify membership
```

`getuser`/`users` are in there because the reload is verified server-side (re-read the users and their hashes after `ACL LOAD`), not fired blind.

verified locally against a real valkey 9.0, not envtest:

- a user with only those three grants runs `ACL LOAD` / `ACL GETUSER` / `ACL USERS` with no NOPERM
- the same user is still denied `ACL SETUSER`, so it stays load-and-read only, no user management crept in
- dropping the grants reproduces the exact `NOPERM ... 'acl|load'` you saw

pushed to the branch. mind giving it another spin when you get a moment?

internal/controller/valkeynode_controller_test.go
```diff
@@ -1217,15 +1219,139 @@ type fakeConfigClient struct {
...
+		Expect(err).NotTo(HaveOccurred())
+		Expect(synced).To(BeTrue())
+		Expect(fake.aclLoads).To(BeZero())
+	})
+
+	It("still reloads when the passwords already match, so non-password edits converge", func() {
+		// A permissions-only change, or a removed user, leaves every desired
+		// password hash untouched. The reload must not be skipped on that basis
+		// or those edits would never reach the server.
+		fake := &fakeConfigClient{aclHashes: desiredHashes}
```
sandeepkunusoth, line 1395: can u update e2e test to check if acl test valkeycluster_test.go is actually applied live without rollout restart.

melancholictheory:
@sandeepkunusoth good call on the e2e test, it earned its keep. writing it surfaced a gap the unit tests couldn't see.

the live reload works, but adding a user to a running cluster still rolled every pod. the reason is pre-existing: `buildPodTemplateAnnotations` stamps the internal ACL secret's hash onto the pod template (from #164, back when rolling the pods was how an ACL change got applied). #319 added the live `ACL LOAD` but left that annotation in place, so both fired: the ACL applied live AND the pods rolled. the "without a pod roll" part was never true.

fixed in this PR:

- drop the ACL hash from the pod template, so an ACL edit no longer rolls the pods. the server-config hash stays, since a non-live config change still needs a restart. the ACL secret read in `ensureStatefulSet`/`ensureDeployment` only existed to feed that hash, so it goes too.
- to keep the live path prompt without the roll as its trigger, the node controller now watches the internal ACL secret and reconciles the nodes that mount it, which runs `applyLiveACL`.

the e2e test is the one you asked for: it adds a user to a running cluster and asserts the new user is usable (present in `ACL LIST`, authenticates against the server) with `ACLApplied=True` on every node, while the server pods keep their UIDs. verified locally on kind: without the fix it fails with all six pods recreated ~35s after the ACL change, and with the fix it passes.

melancholictheory:
following up on my last comment: my first cut dropped the ACL-hash roll entirely, and that was too aggressive. running the full e2e suite, not just the new spec, caught it.

the roll was also the only recovery path for a change that locks the operator out. delete the system-passwords secret and the operator regenerates it with a fresh password; the running server still has the old one, so the operator can neither `CONFIG SET` nor `ACL LOAD`, and with no roll the pod never reloads the new aclfile from disk. the cluster just sat in `Reconciling`, looping on WRONGPASS.

so the fix is narrower than "drop the roll": apply ACL changes live, and roll a node only when the live apply fails with WRONGPASS. on that signal the operator stamps the current ACL hash onto the pod template, the pod restarts and re-reads the aclfile, and the node recovers. the hash is preserved across reconciles so a recovered node doesn't roll again, and it's never stamped on the happy path, so an ordinary user add or permission change still applies live without touching the pods.

e2e covers both paths now: adding a user to a running cluster is live with no roll (the pods keep their UIDs), and deleting the password secret recovers through a roll (the cluster returns to Ready and the locked-out pods are recreated). both green locally on kind.

jdheyburn:
Thanks, I ran another test locally with the latest commit and deployed a cluster with below:

```
```

I came across a bug where there the unmanaged `default` ACL created by the server provides a mismatch in comparison of what new users need to be applied - so there is an infinite loop in trying to applying the ACL.

```
2026-07-30T09:19:34Z    DEBUG   desired ACL passwords not live yet, waiting for the aclfile volume to propagate {"controller": "valkeynode", "controllerGroup": "valkey.io", "controllerKind": "ValkeyNode", "ValkeyNode": {"name":"cluster-sample-2-0","namespace":"valkey-operator-system"}, "namespace": "valkey-operator-system", "name": "cluster-sample-2-0", "reconcileID": "4e401416-8812-4496-8c5b-1328f2a63848"}
2026-07-30T09:19:34Z    DEBUG   reconciling ValkeyNode  {"controller": "valkeynode", "controllerGroup": "valkey.io", "controllerKind": "ValkeyNode", "ValkeyNode": {"name":"cluster-sample-2-1","namespace":"valkey-operator-system"}, "namespace": "valkey-operator-system", "name": "cluster-sample-2-1", "reconcileID": "a08bf3e1-a7d1-4c38-8c13-7e6137af05b6"}
2026-07-30T09:19:34Z    DEBUG   reconciled StatefulSet  {"controller": "valkeynode", "controllerGroup": "valkey.io", "controllerKind": "ValkeyNode", "ValkeyNode": {"name":"cluster-sample-2-1","namespace":"valkey-operator-system"}, "namespace": "valkey-operator-system", "name": "cluster-sample-2-1", "reconcileID": "a08bf3e1-a7d1-4c38-8c13-7e6137af05b6", "result": "updated", "name": "valkey-cluster-sample-2-1"}
2026-07-30T09:19:34Z    DEBUG   desired ACL passwords not live yet, waiting for the aclfile volume to propagate {"controller": "valkeynode", "controllerGroup": "valkey.io", "controllerKind": "ValkeyNode", "ValkeyNode": {"name":"cluster-sample-2-1","namespace":"valkey-operator-system"}, "namespace": "valkey-operator-system", "name": "cluster-sample-2-1", "reconcileID": "a08bf3e1-a7d1-4c38-8c13-7e6137af05b6"}
```

Are you able to add a test for this case, before adding the fix?

Not for this PR, but I wonder if its worth making the `default` ACL always managed (that is, it exists in `spec.users` as enabled).

melancholictheory:
good find, and thanks for the test-first nudge. reproduced it, added the test, then the fix.

the mismatch is the unmanaged `default`. a cluster with no custom users has an aclfile of system users only, but the server always has its own `default`, and `ACL LOAD` can't remove it. `aclObservablyInSync` compared the two user sets for exact equality, so `default` sat on the server side forever, the node never read as synced, and it looped on "not live yet". the sync check now ignores `default` when the aclfile doesn't manage it (unit cases for both: unmanaged default ignored, managed default still compared), plus an e2e assertion that a plain cluster reaches `ACLApplied=True`.

while i was in there i took greptile's two summary points, since they land on the same recovery path:

- the WRONGPASS recovery now also fires on NOPERM and NOAUTH. that covers the upgrade case it flagged: existing pods run the old `_operator` ACL, so `ACL LOAD` returns NOPERM (or NOAUTH if `_operator` predates this), and the roll is what reloads the aclfile with the new grants. the hash stamp is idempotent, so a node that still can't be reached rolls once instead of looping.
- `currentACLHash` reads through the APIReader now, like `applyLiveACL`, so the recovery roll can't miss on a stale cached hash.

on always-managing `default`: agree it's worth doing, and it would drop the special-case here, but it's a spec and behaviour change, so happy to keep it out of this PR and pick it up as the follow-up you mentioned.

jdheyburn:
@melancholictheory We're going to prioritise the below PR as its a bug fix, I believe it blocks your PR since it includes logic on rolling pods if ACL has been updated. Your changes would undo that, so I think we will want to wait for this to be merged, then your change rebased against it. How does that sound? Appreciate the time you've put into this already.

- https://github.com/valkey-io/valkey-operator/pull/338

melancholictheory:
sounds right, and thanks for flagging it early. #338 is the better base: it makes the pod-template apply path staged and cluster-permitted, and #319 rewrites that same path (the ACL hash on the template, the recovery roll), so landing them the other way around would just have #319 stomping the permit gating. happy to wait and rebase on top.

one thing i'll want to reconcile on the rebase, flagging it now so it isn't a surprise: #319's recovery roll fires when the operator is locked out of a node (WRONGPASS/NOPERM/NOAUTH after a credential change) and stamps the ACL hash to force a restart, because a restart is the only way the node reloads the aclfile and comes back. under #338 that stamp goes through the cluster permit like any other template change. that's fine for the common case, but a locked-out node is already broken, so i'll want to check the permit doesn't hold the recovery behind a health gate the broken node can't satisfy. i'll work that out against #338's actual mechanism rather than guess at it now.

no rush on my end. ping me when it lands and i'll rebase.

jdheyburn:
@melancholictheory #338 is now merged in, if you would like to rebase and remove the ACL component from the WorkloadRevision hash. 

My understanding is that ACLs should be taken out of the WorkloadRevision, because they are applied live via ACL LOAD - correct me if I am wrong.

melancholictheory:
Thanks, rebased onto #338. You've got it right: ACL is out of the WorkloadRevision now. It's applied live with ACL LOAD instead, so an ACL edit no longer rolls the pods.

What's in this revision:
- `buildPodTemplateAnnotations` no longer stamps the ACL hash, so ACL changes don't enter the workload revision.
- The node controller reloads the mounted aclfile on reconcile and watches the internal ACL Secret, and reports an `ACLApplied` condition once the desired users and their password hashes are observably live.
- The operator user gains `+acl|load` / `+acl|getuser` / `+acl|users`.
- Dropped the ACL-secret threading that used to feed the revision through the cluster controller and the failover preflight.

Verified on a 3-shard cluster: adding a user shows up in `ACL LIST` on every node and authenticates, with no pod restarts, and `ACLApplied` goes True across the nodes.

One thing I pulled out into a follow-up rather than fold in here: recovering an operator that has locked itself out. A deleted or rotated password Secret leaves it on WRONGPASS, so it can't run ACL LOAD to fix itself. Previously the ACL-hash roll happened to recover that; without the roll it needs its own path. Two ways I can see:

A. The locked-out node records the lockout and the cluster controller factors it into WorkloadRevision, so recovery goes through the same staged, one-at-a-time, failover-aware roll as everything else.

B. A targeted delete of just the locked-out pod: node-local, without touching WorkloadRevision.

I lean towards A, since it keeps recovery inside the staging model from #338 rather than adding an out-of-band pod delete, but B is simpler if you'd rather keep the lockout path self-contained. Which would you prefer? I also removed the old secret-deletion e2e assertion (it asserted a roll); it comes back with whichever recovery path we land on.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,157 @@
...
+	c, err := r.newConfigClient(ctx, r, node)
+	if err != nil {
+		return false, err
+	}
+	defer c.Close()
+
+	if err := c.LoadACL(ctx); err != nil {
+		return false, err
+	}
+	return aclObservablyInSync(ctx, c, desired)
```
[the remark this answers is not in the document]
melancholictheory: Good catch, and the repro is right: for a permissions-only edit (no user or password change), `ACLApplied` can read True during the window before the mounted aclfile refreshes, because `aclObservablyInSync` only compares the user set and password hashes.

That scoping is deliberate, and the doc comment on the function says so: permissions aren't compared because `ACL GETUSER` returns Valkey's normalized form while the operator only holds the aclfile text, so a real comparison means reimplementing Valkey's ACL parser. The reload itself is unconditional, so a permissions edit does converge on a later reconcile once the volume catches up. What's imprecise is the condition during that window, not the eventual state.

There is a way to close the window without parsing: have the cluster controller stamp a revision into the aclfile as a sentinel user, `user _aclrev_<content-hash> off nopass`, and have the node confirm that user is present (via the `ACL USERS` call it already makes) before reporting applied. Then `ACLApplied` means "the current aclfile revision is loaded" rather than "users and passwords match", which is honest for permissions too. The cost is a visible bookkeeping user in `ACL LIST`.

@jdheyburn do you want that in this PR, or as a follow-up? I'm also fine leaving `ACLApplied` scoped to users and passwords with the comment as-is if you'd rather not add a sentinel user, since permissions converge either way. Happy to do whichever you prefer.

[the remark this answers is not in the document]
melancholictheory: Quick update: I prototyped the sentinel to see how it lands. It's contained, the cluster controller appends `user _operator_acl_sentinel off #<hash-of-managed-acl>`, and `aclObservablyInSync` picks it up through the user and hash comparison it already runs, so the node side needs no new logic.

Verified on a live cluster: a permission-only edit (same users and passwords) now holds `ACLApplied=False` while the mounted aclfile is stale and flips to True only once the server loads the new revision. Before the sentinel that edit read as applied right away.

Happy to fold it into this PR, @jdheyburn, or keep it as a follow-up, your call.

internal/controller/valkeynode_controller_test.go
melancholictheory: Done, this is covered by the "live ACL propagation" spec. It brings a cluster up, records each server pod's UID, adds a user through the spec, then checks the new user shows up in a live `ACL LIST` and can authenticate while every pod UID stays the same afterwards (with a `Consistently` window, so a rollout restart would fail it). Every ValkeyNode also has to report `ACLApplied=True`.

api/v1alpha1/valkeynode_types.go
```diff
@@ -190,6 +190,11 @@ const (
 	// of Spec.Config has been successfully applied via CONFIG SET. The cluster
 	// controller blocks one-at-a-time progress until this condition is True.
 	ValkeyNodeConditionLiveConfigApplied = "LiveConfigApplied"
+	// ValkeyNodeConditionACLApplied indicates that the ACL the cluster controller
+	// wrote to the mounted Secret is live on the server. It is applied via ACL
+	// LOAD without a pod roll, so ACL never enters Spec.WorkloadRevision. True
+	// means the desired user set and their password hashes are live.
+	ValkeyNodeConditionACLApplied = "ACLApplied"
```
bjosv, line 197: We can add this new condition to `docs/status-conditions.md` under `### ValkeyNode conditions`

bjosv (commented):
In `docs/valkeycluster.md` we mention that `configurations can be applied live without rolling pods`. We could add info about this new handling as well? (under `### Users`?)

api/v1alpha1/valkeynode_types.go
melancholictheory: Both added in bebad5e: an `ACLApplied` section alongside the other ValkeyNode conditions in `docs/status-conditions.md`, and a note under `Users` in `docs/valkeycluster.md` that ACL changes apply live via `ACL LOAD` without a pod roll, matching the live-config wording you pointed at.

internal/controller/valkeynode_controller.go
```diff
@@ -278,10 +380,15 @@ func (r *ValkeyNodeReconciler) ensureWorkload(ctx context.Context, node *valkeyi
 // buildPodTemplateAnnotations assembles the annotations that must be present on
 // the pod template spec to trigger rolling updates when the ACL secret or the
```
bjosv: The old comment should be removed I guess.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+		}
+		out[fields[1]] = normaliseHashes(hashes)
+	}
+	return out
+}
+
+// normaliseHashes sorts and deduplicates password hashes so both sides of the
+// comparison are in the same shape.
+func normaliseHashes(hashes []string) []string {
```
bjosv: hyper-nit: with this we get a mix of normalize and normalise (like the "americanized" normalizePodTemplate)

internal/controller/valkeynode_acl.go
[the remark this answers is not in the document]
bjosv: We could suggest that Valkey provide a flag to ACL LOAD, or add a new ACL DIGEST command, or a similar mechanism in the future.
But for now, the idea of keeping the hash via the extra user is interesting, we already create a bunch of users.
+1 for adding the user, but others would have to weigh in before we change anything.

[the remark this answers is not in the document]
melancholictheory: Opened valkey-io/valkey#4355 to track the native mechanism (an `ACL DIGEST`, or a digest returned by `ACL LOAD`); if that lands it replaces the sentinel entirely.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,157 @@
...
+	c, err := r.newConfigClient(ctx, r, node)
+	if err != nil {
+		return false, err
+	}
+	defer c.Close()
+
+	if err := c.LoadACL(ctx); err != nil {
+		return false, err
+	}
+	return aclObservablyInSync(ctx, c, desired)
```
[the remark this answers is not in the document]
melancholictheory: Thanks for the repro. The NOPERM is real, but on upgrade it's transient rather than a permanent deadlock, because the reconcile loop rolls the pod once for an unrelated reason and that roll bootstraps the new operator permissions.

This PR removes the ACL hash from the pod-template annotations. A cluster created by the current operator has that annotation on its live pods, so on upgrade the desired template (without it) differs from the live one, which is a WorkloadRevision change and a one-time staged roll. The pod restarts, loads the new aclfile from the mounted Secret at startup, and comes back with an `_operator` that has `+acl|load` / `+acl|getuser` / `+acl|users`. Live ACL works from there.

I ran it end to end on a cluster: created it on the operator built from current main (`_operator` without the ACL grants, acl-hash annotation present), then upgraded to this PR. Result:

- the pods rolled once and the acl-hash annotation was dropped,
- `ACL GETUSER _operator` afterward showed `+acl|load +acl|getuser +acl|users`,
- `ACLApplied` reached True on every node (a few transient NOPERM log lines during the roll window, then clean),
- adding a user afterward applied live in about 30s with no further roll.

The T-Rex harness exercises `applyLiveACL` in isolation, so it doesn't see that incidental roll from the full reconcile. The one case that wouldn't self-bootstrap is a cluster with no acl-hash annotation at all, which predates that annotation (before live ACL existed).

The other half of the repro, `synced=true` while old permissions are still live, is the permission-window point from the sentinel thread above rather than the bootstrap path, and that is what the sentinel closes.

Happy to add an upgrade e2e that pins this (old-style ACL in, upgrade, assert the one-time roll bootstraps and live apply works afterwards) if that's useful.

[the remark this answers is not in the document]
melancholictheory: Both added in the latest push.

- Upgrade note in `docs/valkeycluster.md` under Users: the one-time roll from dropping the annotation is what grants `_operator` the ACL commands, and the pre-annotation cluster is called out as needing a manual restart.
- An e2e (`acl-hash annotation migration`) that stamps the legacy `internal-acl-hash` on a running cluster's StatefulSets and asserts the operator strips it while the cluster stays Ready with `ACLApplied=True`.

The e2e simulates the pre-upgrade state by stamping the annotation rather than running two operator builds, since the harness deploys a single image. It exercises the same invariant: an old-style annotation on the template is removed by the operator, which is the migration roll.

jdheyburn:
@melancholictheory I did some testing on this to verify it. I noticed with changing some fields that aren't "observed" that we lose their visibility on the ACLApplied condition. For example, I set up this command to watch for the condition:

```
kubectl get valkeynodes \
  -o custom-columns='NAME:.metadata.name,ACL:.status.conditions[?(@.type=="ACLApplied")].status,REASON:.status.conditions[?(@.type=="ACLApplied")].reason' -w

NAME                 ACL    REASON
cluster-sample-0-0   True   Applied
cluster-sample-0-1   True   Applied
cluster-sample-1-0   True   Applied
cluster-sample-1-1   True   Applied
cluster-sample-2-0   True   Applied
cluster-sample-2-1   True   Applied
```

Then I changed fields under `spec.users`. The below all changed to `PendingPropagation` and one-by-one the ValkeyNodes went back to `Applied` when it was synced.

- Adding/removing a user
- Adding/removing a user password

For these actions, they did not change to `PendingPropagation`, but they did eventually get applied

- Changing `enabled`
- Changing permissions

I believe you called out this limitation earlier in the thread. Perhaps we can document the live ACL apply limitation for now and raise issues to fix these in the future - how does that sound? I am aware a lot of good work has gone into this PR and don't want to hold it out.

melancholictheory:
Sounds good, and thanks for testing it. That matches the design: `ACLApplied` compares the user set and password hashes exactly, so adding or removing a user or a password shows `PendingPropagation` then `Applied`. `enabled` and permission edits still apply through the same unconditional `ACL LOAD`, but the condition doesn't reflect them in transit, since comparing rules would mean reparsing Valkey's normalized ACL form.

I'll document that in this PR: what the condition observes exactly, and that `enabled`/permission changes converge without a `PendingPropagation` step.

For the follow-up, the revision sentinel I prototyped above closes exactly this gap (a hash of the whole managed ACL, so `enabled` and permission edits move the condition too), and I verified it live. Happy to raise an issue for it, or fold it in later, whichever you prefer. Documenting the current behaviour here and not holding the PR sounds right to me.

jdheyburn:
@melancholictheory Thanks, I agree with this:

1. Documenting the limitations in this PR
2. Follow up PR to improve on the limitations

test/e2e/valkeycluster_test.go
```diff
@@ -2050,6 +2015,324 @@ spec:
...
+				}
+			}, 5*time.Minute, 5*time.Second).Should(Succeed())
+
+			By("the cluster returns to Ready and ACL stays live after the migration")
+			Eventually(func(g Gomega) {
+				cr, err := utils.GetValkeyClusterStatus(clusterName)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+			}, 10*time.Minute, 5*time.Second).Should(Succeed())
+			Eventually(expectACLAppliedTrue, 5*time.Minute, 5*time.Second).Should(Succeed())
```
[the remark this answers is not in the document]
melancholictheory: Fair point, and you're right about what it covers. The spec runs against the current operator, whose `_operator` already holds `acl|load`/`acl|getuser`/`acl|users`, so it exercises the reconcile trigger (the operator strips the legacy annotation and ACL stays live across the roll) rather than the permission bootstrap itself.

I tried building the missing-permission state in-test by downgrading `_operator` with `ACL SETUSER _operator -acl|load ...`, but it isn't faithful: the operator's unconditional `ACL LOAD` reloads the aclfile (which still grants those commands) and heals the downgrade within a reconcile, without any roll. A real upgrade has no such window, the server boots from the old aclfile without the commands, so every `ACL LOAD` is `NOPERM` and the roll is the only thing that grants them. Reproducing that faithfully needs a second, pre-live-ACL operator build deployed first, which the single-image e2e harness can't do.

I did validate the full bootstrap end to end outside the suite: created a cluster on the operator built from current main (`_operator` without the grants, annotation present), upgraded to this build, and confirmed the pods rolled once, `ACL GETUSER _operator` then showed all three commands, and a later ACL edit applied live. That's the run I described above.

I've reworded the spec's title and comment so it claims only the trigger and the live-ACL survival, not the grant. If you'd rather I drop the spec, happy to do that too.

internal/controller/valkeynode_acl.go
```diff
@@ -0,0 +1,157 @@
...
+	for _, user := range slices.Sorted(maps.Keys(desired)) {
+		actual, err := c.UserPasswordHashes(ctx, user)
+		if err != nil {
+			return false, err
+		}
+		if !slices.Equal(actual, desired[user]) {
+			return false, nil
+		}
+	}
+	return true, nil
```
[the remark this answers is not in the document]
melancholictheory: This is the permissions-only window we've been working through with @jdheyburn above. He tested it and confirmed exactly this: user and password changes move `ACLApplied` through `PendingPropagation`, while `enabled` and permission edits apply but don't drive the condition. The call for this PR is to document the limitation and fix it in a follow-up, which is where it stands: the `ACLApplied` docs now spell out that only the user set and password hashes drive the condition, and the revision sentinel I prototyped earlier in the thread closes the gap (a hash of the whole managed ACL, so permission edits move the condition too).

jdheyburn (approved):
Thanks once again for taking this on!

daanvinken:
For the upgrade 0.4 -> 0.5 I noticed clusters that already had `_operator `from before this PR log:
```
ACL LOAD: NOPERM User _operator has no permissions to run the 'acl|load' command
```

The Secret already includes `+acl|load`, but the running process still has the old ACL. Until each pod restarts once and reloads `users.acl`, live load cannot grant itself that permission.

Clears after the staged roll (or a one-off pod restart). Not blocking for us; but we may want to post that on the release notes. The errors stood out during upgrade.

melancholictheory:
That's the expected upgrade path, thanks for flagging it. This version drops the ACL-hash annotation from the pod template, so every pre-existing cluster rolls once, and that roll is what restarts each pod onto the new `users.acl` and grants `_operator` the `acl|load` command. The NOPERM lines are the window before a pod has rolled; they clear as the staged roll moves through, or with a manual restart as you saw.

It's written up under Users in `docs/valkeycluster.md` (the upgrade note), and a release-notes line calling out the transient NOPERM on 0.4 to 0.5 is a good idea.

jdheyburn:
I've updated the release notes with the same: https://github.com/valkey-io/valkey-operator/releases/tag/v0.5.0

melancholictheory:
One small thing on the release-note wording: "manual restart required" reads a bit stronger than what actually happens. On upgrade the operator drops the ACL-hash annotation from the pod template, which triggers the one-time staged roll on its own, and that roll reloads `users.acl` and clears the NOPERM. So for a normal cluster it self-resolves as the roll moves through, without a manual step.

A manual restart only helps if you want it cleared immediately, or for a cluster old enough to predate that annotation (no annotation to drop means no automatic roll). Might be worth softening the note to something like: "the one-time staged roll on upgrade clears this on its own; a manual restart is only needed to clear it immediately, or for clusters that predate the ACL-hash annotation."

Issue #303, opened 2026-07-12
PodDisruptionBudget should be scoped per shard instead of per cluster


erkules:
# PodDisruptionBudget should be scoped per shard instead of per cluster

## Problem

The operator currently creates a single `PodDisruptionBudget` for the entire `ValkeyCluster` with:

* `maxUnavailable: 1`
* a selector matching only the cluster label (`valkey.io/cluster`)

As a result, voluntary disruptions are limited across the entire cluster instead of being evaluated independently for each shard.

For example, with:

```yaml
spec:
  shards: 3
  replicas: 2
```

the cluster consists of three independent shards:

```
Shard 0: Primary + Replica + Replica
Shard 1: Primary + Replica + Replica
Shard 2: Primary + Replica + Replica
```

A disruption affecting a pod in Shard 0 unnecessarily blocks a simultaneous disruption in Shard 2, even though the shards are independent failure domains.

## Proposed solution

Instead of creating a single cluster-wide `PodDisruptionBudget`, create one PDB per shard.

Each PDB should select pods using both:

* `valkey.io/cluster`
* `valkey.io/shard-index`

For example:

```yaml
selector:
  matchLabels:
    valkey.io/cluster: my-cluster
    valkey.io/shard-index: "0"
```

This results in one PDB for each shard.

## PDB policy

I suggest using:

```yaml
minAvailable: 1
```

for each shard.

The purpose of the PDB is to ensure that at least one node of each shard remains available during voluntary disruptions while allowing maintenance to proceed independently across different shards.

## Benefits

* PDBs align with the shard-based architecture of Valkey Cluster.
* Independent shards no longer block each other's maintenance.
* Node drains and rolling updates can progress concurrently on different shards.
* Availability guarantees become scoped to the actual failure domain (the shard) instead of the entire cluster.

I'd be happy to contribute a PR implementing this change if the maintainers agree with the approach.

jdheyburn:
Continuing the conversation from:
- https://github.com/valkey-io/valkey-operator/pull/304#issuecomment-4955563297

The PDB provided was always meant to be a strict, minimal PDB so that we had something while we iterated on the design.

> Availability guarantees become scoped to the actual failure domain (the shard) instead of the entire cluster.

[...]

Issue #266, opened 2026-06-18
[chore] Introduce client registry and provider to centralise Valkey client instantiation


jdheyburn:
### Problem

Valkey client creation is scattered across multiple controllers and utility files. Each caller constructs its own connection with its own TLS config, credential lookup, and error handling. This makes it hard to add cross-cutting concerns (retry logic, connection pooling, observability) consistently, and leads to subtle divergences between how different parts of the code connect to pods.

This was noted as a future improvement in PR #228:
> "I would like to propose a client registry that handles this in the future."

### Proposed solution

Introduce an internal `ClientProvider` interface and a `ClientRegistry` implementation:

```go
type ClientProvider interface {
    // ForNode returns a connected client for the given ValkeyNode.
    ForNode(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) (Client, error)
    // ForPod returns a connected client for a pod IP and port.
    ForPod(ctx context.Context, podIP string, port int, opts ...ClientOption) (Client, error)
}
```

The registry handles TLS config resolution, credential lookup, and connection reuse. All controllers go through this interface instead of instantiating clients directly. Tests inject a fake via the interface.

### Acceptance criteria

- [ ] `ClientProvider` interface defined in `internal/client/`
- [ ] All direct `vclient.NewClient` calls in controllers replaced with `ClientProvider.ForNode` or `ForPod`
- [ ] Retry/backoff policy centralised in the registry
- [ ] `FakeClientProvider` available for unit tests
- [ ] No behaviour change in existing functionality

### References

- PR #228 (note from jdheyburn)

jdheyburn:
I've already been working on a draft of this on a branch, so I'll like to continue working on it.

PR #320, opened 2026-07-16
enhancement: reliable, event-driven ValkeyNode.Status.Role

### LINKED ISSUE #261: [enhancement] Event-driven ValkeyNode reconcile for timely Status.Role after failover / cluster changes

jdheyburn:
This PR closes #261

### Summary

The existing Status.Role being reported on ValkeyNode was quite naive in that it would only be updated after a reconciliation cycle, which was every 60s. This can mean it lags behind when failovers happen for whatever reason.

While nothing depends on Status.Role for now (it is meant purely for the end user), behind this PR will be one to patch a label (smth like `valkey.io/current-role`) on Pods of what their current role is (to be used in PDBs per-shard as part of #303). That implementation will sit in the same logic as this PR is adding.

- Status.Role now reflects role changes (failover, restart) within a couple of seconds rather than waiting up to 60s.
- When the pod is not ready the role is reported as an empty string, since we don't want to serve stale data.
- Restarted replicas no longer transiently report primary, the role is read from cluster slot ownership rather than INFO replication (which briefly reports master before a restarted replica re-syncs).
- A pod keeps its role while its StatefulSet/Deployment is mid-roll. Only `status.ready` reports the rollout — a serving pod has a real role and blanking it was misleading.
- A new internal RolePoller samples live topology every 5s and triggers a re-resolve when it sees drift. It never writes the role itself.

### Implementation

Role resolution stays with the ValkeyNode controller (it remains the only writer of Status.Role), but it now becomes event-driven instead of relying on the 60s cadence:

- How the role is determined
  - In cluster mode the role is read from slot ownership in the node's own CLUSTER NODES (owns a slot → primary, else replica), not from INFO replication
  - A just-restarted replica reports role:master for a few seconds before it re-establishes replication, so INFO's role (and the master/slave flag) is transiently wrong; slot ownership is correct throughout
  - Standalone (non-cluster) nodes have no slots and fall back to INFO replication.
- When it re-resolves
  - A filtered Pod watch triggers a reconcile on readiness / IP / phase changes. That covers restarts.
  - It does not cover failover. When a replica is promoted between two healthy pods, *nothing in Kubernetes moves* — no restart, no readiness flip, no IP change — so no watch can fire.
    - `internal/controller/rolepoller.go` adds a `RolePoller` manager Runnable for exactly this case. Every 5s it scrapes live topology, compares each node's live role against its `status.role`, and pushes a `GenericEvent` down an in-process channel the ValkeyNode controller watches (`source.Channel`).
    - It triggers, it never writes. The ValkeyNode controller re-resolves from its own connection as usual, so the single-writer invariant holds and a false positive costs one extra reconcile and nothing else.
    - A tick where every role matches performs zero API operations. It runs on the elected leader only, and a node that fails to answer is backed off exponentially to a 1 minute ceiling rather than dialled every tick.
    - The interval is a constant, not a flag — polling cost is ours to reason about, not an operator-facing knob.
  - A 30s backstop requeue (10s if a ready node's role couldn't be resolved) remains the self-heal net. It is what makes a wedged poller a latency regression rather than an availability bug.
- Not-ready / transient handling
  - Role is cleared to "" when the pod is not ready
  - If the pod is ready but the live read fails transiently, the last-known role is kept (not blanked) and we requeue sooner.
- The ValkeyCluster controller now takes exactly one cluster-state snapshot per reconcile, hoisted into `Reconcile` and passed down, rather than scraping separately inside `reconcileValkeyNodes`. It also scrapes whenever nodes are running rather than only during rolls.

Reviewer attention:
- The two-controller boundary: the poller only ever enqueues; `Status.Role` is written only by the ValkeyNode controller.
- The choice to read role from slot ownership rather than INFO replication (see the restart-transient reasoning above).
- The cost of polling: each tick opens one connection per node (~1.2 connects/sec for a 6-node cluster at 5s). `scrapeFunc` which will be refactored once pooled client registry from #266 is implemented, after which the interval can go lower because there will always be a connection pool maintained to the node.
- Anything consuming role at this cadence needs to handle mid-failover transients, a shard with no slot-owning primary is now observable. For the #303 PDB that must be **fail-closed** (block the eviction), otherwise we trade a stale-role window for a permissive one.

### Limitations

- Role is best-effort with a bounded lag. A failover is detected within one poll interval (5s) plus a reconcile; anything the poller misses entirely is bounded by the 30s backstop.
- A cluster primary that owns no slots yet (during initial formation or resharding) reads as replica until it owns slots.
- Connections are unpooled, one per node per tick - which matches the current pattern we have in the status quo and will be pooled in #266.

### Testing

Unit + envtest cover the role parsers, the set/clear/keep-last-known logic, the role-survives-a-roll behaviour, the pod predicate and map function, and the poller's drift detection, per-node backoff and leader-election gating.

An e2e test issues `CLUSTER FAILOVER` on a replica and asserts the roles swap on the CRs within 20s — under the 30s backstop, so only the poller can close that gap, with neither pod replaced.

Measured on a Kind cluster:

| scenario | before | after |
|---|---|---|
| delete a replica pod | flashed primary for ~7s (INFO reports role:master until it re-syncs) | rejoins as replica, no flash |
| delete a primary pod | ~15s | ~2s |
| `CLUSTER FAILOVER`, no pod restart | ~13s (measured: failover at 07:59:41, Valkey swapped at 07:59:42, CRs caught up at 07:59:54) | within one poll interval |

### Checklist


jdheyburn:
@greptile-apps Review

melancholictheory:
this is the right foundation to build the label on, and keying the hint on the annotation value rather than `Status.Role` is a neat way to keep it churn-free. the bounded gap you documented is the honest tradeoff.

one thing worth pinning down now, since this PR fixes the meaning of "role" that the PDB in #303 will consume: valkey only counts masters that serve at least one slot as voters in a failover election (a master with `numslots == 0` does not vote). so the set that matters for the quorum math we landed on, `maxUnavailable = floor((shards-1)/2)` over current primaries, is "masters with slots", not "masters".

that distinction shows up during a reshard: a primary whose last slots have just migrated away is still flagged master until it demotes, but it has already stopped voting. if `liveRoleForAddress` resolves through the master flag rather than slot ownership, that node reads as a primary and gets counted in the quorum while contributing nothing to it, which makes the PDB more permissive than the math intends. does `GetPrimaryNode` key on slots here?

for the follow-up label: worth thinking about the lifecycle when a pod is recreated. the label lives on the pod, not the pod template, so a fresh pod starts without it and a primaries PDB selecting on it won't match until the next reconcile patches it back. brief and self-healing, but it's a window where the node is unprotected right after a roll, so patching it as soon as the pod reports Ready is worth the ordering care.

and to state the invariant from #310 explicitly, since both live in the same area now: this label is the right input for the PDB because eviction re-evaluates it continuously, and the wrong input for topology spread, which is evaluated once at placement and wants the stable node-index=0 identity. same label, opposite conclusions.

jdheyburn:
@greptile-apps @coderabbitai

jdheyburn:
Update: this is a lot more responsive with these changes, but there is still a bit of a delay in getting the role during failovers when compared to rolling pods. This is because this PR focuses on subscribing to Pod events as the mechanism for checking the role label.

When failovers happen outside of the operators control* there is no event to subscribe to, and Cluster does not have a pub/sub channel for exposing failover events unlike Sentinel. So the design should err towards a dedicated thread that can poll one or more of `INFO`, `ROLE`, and `CLUSTER NODES` on a faster cadence (2s?) to determine the role. Although the way the operator invokes clients to nodes today is inefficient.

\* `shutdown-on-sigterm failover` invokes a CLUSTER FAILOVER when the pod is terminated

jdheyburn:
@greptile-apps @coderabbitai

jdheyburn:
I've implemented the RollPoller thread as I mentioned on the previous comment, this will help plug the lag that's caused by any failovers that happen outside of the operator.

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+func (p *RolePoller) pollCluster(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster, now time.Time, seen map[types.NamespacedName]struct{}) {
+	log := logf.FromContext(ctx).WithValues("cluster", cluster.Name, "namespace", cluster.Namespace)
+
+	nodes := &valkeyiov1alpha1.ValkeyNodeList{}
+	if err := p.Client.List(ctx, nodes,
+		client.InNamespace(cluster.Namespace),
+		client.MatchingLabels(map[string]string{LabelCluster: cluster.Name}),
+	); err != nil {
+		log.Error(err, "failed to list ValkeyNodes")
+		return
```
deepakpunjabi, line 188: Should there be a condition in valkeycluster status to indicate that pollCluster is failing for this valkeycluster?

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+}
+
+// tick samples every cluster once and emits an event per drifted node. A tick
+// where every role matches performs no API writes at all.
+//
+// now is passed in rather than read from the clock so tests can drive the
+// per-node backoff deterministically.
+func (p *RolePoller) tick(ctx context.Context, now time.Time) {
```
deepakpunjabi: nit: tick seems a bit generic as a name. Would it make sense to make it more verbose? (eg. checkRoleDrift or something else)

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info("stopping role poller")
+			return nil
+		case <-ticker.C:
+			p.tick(logf.IntoContext(ctx, log), time.Now())
```
deepakpunjabi: Is passing logger in arguments a well known pattern? Would it make sense to put logger into RolePoller struct itself?

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+	// Candidates are nodes with a pod IP that are not currently backed off. A
+	// node still waiting for an IP has nothing to dial, and a node in backoff is
+	// left out of the address list entirely so no connection is attempted.
+	var addresses []string
+	candidates := make([]*valkeyiov1alpha1.ValkeyNode, 0, len(nodes.Items))
+	for i := range nodes.Items {
+		node := &nodes.Items[i]
+		seen[client.ObjectKeyFromObject(node)] = struct{}{}
+		if node.Status.PodIP == "" {
+			continue
```
deepakpunjabi, line 203: Should we log this?

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		if state, ok := p.backoff[client.ObjectKeyFromObject(node)]; ok && now.Before(state.nextAttempt) {
+			continue
```
deepakpunjabi, line 207: I guess we should log all this at least as part of debug logger.

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+		candidates = append(candidates, node)
+		addresses = append(addresses, node.Status.PodIP)
+	}
+	if len(addresses) == 0 {
+		return
+	}
+
+	state := p.scrape(ctx, cluster, addresses)
+	if state == nil {
+		return
```
deepakpunjabi, line 224: same as above for logging

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+	seen := make(map[types.NamespacedName]struct{})
+	for i := range clusters.Items {
+		p.pollCluster(ctx, &clusters.Items[i], now, seen)
+	}
+
+	// Drop backoff state for nodes that have gone away, so the map cannot grow
+	// without bound across cluster deletions and scale-in.
+	for key := range p.backoff {
+		if _, ok := seen[key]; !ok {
+			delete(p.backoff, key)
```
deepakpunjabi, line 162: should we add some comment in document or code to make these parts concurrency safe if someone in future tries to make parallel calls to each nodes in the cluster, to avoid any panic?

internal/controller/valkeycluster_controller.go
```diff
@@ -181,7 +181,18 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
-	if requeue, err := r.reconcileValkeyNodes(ctx, cluster, nodes, configHash); err != nil {
+	// One cluster-state snapshot per reconcile, shared throughout reconcile
+	operatorPassword, err := fetchSystemUserPassword(ctx, operatorUser, r.Client, cluster.Name, cluster.Namespace)
+	if err != nil {
+		log.Error(err, "failed to retrieve system user password")
+		setCondition(cluster, valkeyiov1alpha1.ConditionReady, valkeyiov1alpha1.ReasonSystemUsersAclError, err.Error(), metav1.ConditionFalse)
+		_ = r.updateStatus(ctx, cluster, nil)
```
deepakpunjabi, line 190: I am not sure if updating the status to nil is the right call here. We will lose information about the last known cluster state. Since we are already setting the condition, would it be better to avoid this operation or add a `ClusterStateUnknown`?

internal/controller/valkeynode_controller.go
```diff
@@ -668,12 +690,12 @@ func (r *ValkeyNodeReconciler) ensureConfigMap(ctx context.Context, node *valkey
 }
 
 // updateStatus updates the ValkeyNode status based on workload and Pod state.
-func (r *ValkeyNodeReconciler) updateStatus(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) error {
+func (r *ValkeyNodeReconciler) updateStatus(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode) (roleResolveFailed bool, err error) {
```
deepakpunjabi: reason about signature change

internal/controller/valkeynode_controller.go
```diff
@@ -912,13 +957,22 @@ func (r *ValkeyNodeReconciler) getValkeyRole(ctx context.Context, node *valkeyio
...
 	}
 
-	return parseValkeyRole(info)
+	if !parseClusterEnabled(info) {
+		return parseValkeyRole(info)
+	}
+
+	nodes, err := c.Do(ctx, c.B().ClusterNodes().Build()).ToString()
+	if err != nil {
+		log.Error(err, "failed to get cluster nodes")
```
deepakpunjabi, line 982: Not sure why this should not be a reconciler error. Is it because previous `c.Do(ctx, c.B().Info().Section("replication").Build()).ToString()` was not a reconciler error?

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+		if liveRole == node.Status.Role {
+			continue
+		}
+		// A disagreement the node controller cannot settle — it failed to resolve
+		// the role, or the pod is not ready — is re-emitted every tick until it
+		// clears. That is bounded to one reconcile per interval per node, and the
+		// workqueue's rate limiter collapses repeats while one is already queued.
+		log.V(1).Info("live role differs from status, triggering reconcile",
+			"node", node.Name, "live", liveRole, "status", node.Status.Role)
+		p.emit(node)
```
deepakpunjabi: Is there a possibility that valkey-server will start responding to `valkey-cli cluster nodes`while pod is not ready? In such cases we might want to avoid emit while pod waits for becoming ready(eg. pod is loading large aof/rdb) to cause extra operation between controller and poller.

internal/controller/valkeynode_controller.go
```diff
@@ -942,6 +996,53 @@ func (r *ValkeyNodeReconciler) applyLiveConfig(ctx context.Context, node *valkey
...
+			continue
+		}
+		flags := strings.Split(fields[2], ",")
+		if !slices.Contains(flags, "myself") {
+			continue
+		}
+		if slices.Contains(flags, "master") && len(fields) > 8 {
+			return RolePrimary
+		}
+		return RoleReplica
```
deepakpunjabi, line 1054: Do we need to check if this is a slot-less master before assigning it replica role, similar to `liveRoleForAddress`?

internal/controller/valkeynode_controller.go
```diff
@@ -942,6 +996,53 @@ func (r *ValkeyNodeReconciler) applyLiveConfig(ctx context.Context, node *valkey
...
+	for line := range strings.SplitSeq(clusterNodes, "\n") {
+		fields := strings.Fields(line)
+		if len(fields) < 8 {
+			continue
+		}
+		flags := strings.Split(fields[2], ",")
+		if !slices.Contains(flags, "myself") {
+			continue
+		}
+		if slices.Contains(flags, "master") && len(fields) > 8 {
```
deepakpunjabi, line 1051: Regarding `len(fields) > 8 {`, should a primary migrating slots to other node be considered primary? Does the staleness of status role primary have any effect on potential consumers?

Btw(for future), GetSlots() implements similar functionality. It would be better to parse the cluster nodes output into a struct and reuse the function?

internal/controller/rolepoller.go
deepakpunjabi: Should passed ctx in tick have more aggressive timeout? 
unreachable or partitioned nodes will have retries for each address, increasing the time for overall tick?

internal/controller/rolepoller.go
```diff
@@ -0,0 +1,278 @@
...
+	return scrapeClusterState(ctx, p.APIReader, cluster, addresses, operatorUser, password)
+}
+
+// emit pushes a reconcile trigger for the node, dropping it if the channel is
+// full. Dropping is correct: the backstop requeue still catches the change,
+// whereas blocking would stall every other cluster behind one slow consumer.
+func (p *RolePoller) emit(node *valkeyiov1alpha1.ValkeyNode) {
+	select {
+	case p.Events <- event.GenericEvent{Object: node.DeepCopy()}:
+	default:
```
deepakpunjabi, line 287: Should we add a metrics counter for default case?

bcfmtolgahan:
Some numbers for the problem this fixes, measured on main @ c33988f on EKS 1.35, in case they are useful as a before baseline.

Fresh cluster, 1 shard 1 replica, sampled every 8s. Ground truth from `valkey-cli ROLE` on each pod at the same instant:

```
t+8s    VALKEY: 0-0=master 0-1=slave     CR: 0-0=primary 0-1=primary
t+16s   cluster reports Ready            CR: 0-0=primary 0-1=primary
...
t+70s   VALKEY: 0-0=master 0-1=slave     CR: 0-0=primary 0-1=replica
```

Replication was established at t+8s and the CR caught up at t+70s, so 62s behind, 54s of it after the cluster had already gone Ready. `kubectl get valkeynodes` shows two primaries for a 1x1 cluster for most of that window.

Same pattern after an unplanned failover: Valkey promoted the replica at t+15s, the CR still had the old role at t+30s and corrected by t+60s, stable afterwards. Observed convergence delay across three clusters was roughly 40-70s, consistent with the 60s requeue in #261.

The transient-primary-on-restart behaviour you mention in the description matches what I saw at creation time too, so reading role from slot ownership rather than `INFO replication` should cover that case.

melancholictheory:
The detector and writer split is what makes this safe to add. The poller only decides when to wake the node controller, the controller still resolves from its own connection, so a false positive costs one reconcile and the single-writer invariant holds. The per-node backoff and the pruning of `seen` entries are both careful, and buffering `roleEvents` at 1024 means the drop path in `emit` is genuinely rare rather than the common case its comment might suggest.

On the polling cost you flagged for reviewers. The figure in the description, around 1.2 connects per second for a six-node cluster, is per cluster, but `tick` lists every ValkeyCluster with no filter and polls all of them on the same 5s cadence. So the real rate is the total node count across every cluster the operator watches, and with TLS each of those is a handshake, not just a connect. An operator holding a hundred six-node clusters is dialling 600 nodes every 5s, so 120 connections a second, permanently, with the interval fixed at compile time.

That may well be fine, and you have clearly thought about the single-cluster case. The question is whether 5s was picked with the fleet case in mind, since the cost lands on whoever runs the operator rather than on whoever set the field. Worth noting the seam is already there: `Interval` is a struct field and only `cmd/main.go` pins it to the constant, so scaling it by observed node count, or letting a deployment override it, needs nothing new in the poller.

This is also the strongest argument yet for #266. You pointed seaoferror at it in #385 a couple of days ago, where the question was whether per-reconcile connections are worth pooling. This PR moves the same connections from a 60s reconcile to a 5s tick, so the answer there changes by more than an order of magnitude.

Smaller, and only a consistency point: the poller does not look at `DeletionTimestamp`, while the cluster reconcile now returns early for a cluster being deleted since #374. During a foreground cascading delete the poller keeps dialling pods that are on their way out until they stop answering and backoff takes over. Bounded rather than harmful, but a cluster that is going away is one the poller has no reason to sample.

cmd/main.go
```diff
@@ -247,14 +252,30 @@ func main() {
...
+	if err := mgr.Add(&controller.RolePoller{
+		Client:    mgr.GetClient(),
+		APIReader: mgr.GetAPIReader(),
+		Interval:  controller.DefaultRolePollInterval,
+		Events:    roleEvents,
+	}); err != nil {
+		setupLog.Error(err, "Failed to add role poller")
+		os.Exit(1)
+	}
 	// +kubebuilder:scaffold:builder
```
bjosv: nit: I believe `// +kubebuilder:scaffold:builder` sits best right after `(ValkeyNodeReconciler).SetupWithManager` in case with add anything more using kubebuilder.

bjosv (commented):
Seems to work and LGTM

jdheyburn:
Placed back to draft while I address comments

internal/controller/valkeycluster_controller.go
jdheyburn: `updateStatus` skips the shard counts when state is nil:

```go
    if state != nil {
        current.Status.ReadyShards = r.countReadyShards(state, cluster)
        current.Status.Shards = int32(len(state.Shards))
    }
```

so the last known values are kept, nil just means "conditions only" and nothing is lost. It's also the existing convention rather than something new. the five error paths above this one (service, PDB, ACL, ConfigMap, node list) all call it the same way.

Fair enough on ClusterStateUnknown since a cluster we can't scrape currently keeps reporting shard counts with nothing marking them as stale. That would be against all error paths besides the one I'm adding here. Maybe it's worth raising an issue if it's desired.

internal/controller/valkeynode_controller.go
jdheyburn: It's deliberate. updateStatus turns it into roleResolveFailed, which keeps the last-known role instead of blanking a healthy node, and drops the backstop requeue from 30s to 10s so we retry sooner:

```go
    requeueAfter := 30 * time.Second
    if roleResolveFailed {
        requeueAfter = 10 * time.Second
    }
```

Returning an error would abort the rest of the reconcile; status, live config, ACL. All over a transient read on one field, and put the node into workqueue backoff, which is slower than the 10s we get now. Role discovery is best-effort by design and has its own retry.

We don't get the role from `info replication` because a restarted replica comes up with `role:master` for several seconds. `cluster nodes` gets us the master flag _and_ slot ownership.

internal/controller/valkeynode_controller.go
jdheyburn: On migration specifically: the source keeps its remaining slots through MIGRATESLOTS, they only leave range by range, so it stays primary until fully drained. By the time it hits zero the shard is being scaled in and the nodes are going away, so there's no real stale window.

Agree on GetSlots. Let me raise an issue for that: #406

internal/controller/rolepoller.go
jdheyburn: It's going through the context rather than an argument - logf.IntoContext on the way in, logf.FromContext at each use. That's controller-runtime's own pattern and what the rest of the repo does.

The reason it looks odd here is that this is the only IntoContext in the repo. Controllers get their context pre-seeded by controller-runtime so they only ever call FromContext. A Runnable doesn't, so it has to seed its own - that's what the `WithName("rolepoller")` is doing, so poller lines are greppable.

Struct field would work, but it'd diverge from the other controllers and lose the per-cluster WithValues that pollCluster adds on top.

internal/controller/rolepoller.go
jdheyburn: This would only be hit when there was an error retrieving the system password, which is already being logged there.

```
		logf.FromContext(ctx).V(1).Info("skipping role poll, operator password unavailable",
			"cluster", cluster.Name, "namespace", cluster.Namespace, "err", err)
```

internal/controller/rolepoller.go
jdheyburn: Yep good callout, I'll raise a commit that logs the count of skips:

```go
    log.V(1).Info("nodes skipped this pass",
        "dialled", len(addresses), "noPodIP", noPodIP, "backedOff", backedOff)
```

internal/controller/rolepoller.go
jdheyburn: Yep, same reply as here: https://github.com/valkey-io/valkey-operator/pull/320/changes/BASE..bff52dbdea146259b6a4a929e0ae5edce1576541#r3870260128

internal/controller/rolepoller.go
jdheyburn: I'd rather not for ValkeyCluster, the poller is not meant to write against anything, it should delegate writing conditions to the owning controllers. This is only meant to be a detector when the real state has changed from what the Status field says it is.

The question then is more so about observing when the poll isn't working; I've added metrics for triggers raised and dropped, and a log line on backed-off nodes.

That said if there's value in a condition, then it should be on the ValkeyNode side since that is what sets the Status.

internal/controller/valkeynode_controller.go
jdheyburn: They do disagree, and it's structural rather than a choice. liveRoleForAddress returns "" for a slot-less master because GetClusterState puts those in PendingNodes instead of giving them a shard, so there's no shard primary to compare against. The poller skips the node and nothing gets published either way.

I did consider returning "" from parseClusterNodesRole to match but I decided against it: "" already means "couldn't resolve" there, which sets roleResolveFailed and shortens the requeue, so a normal bootstrap would start looking like a failing read.

cmd/main.go
jdheyburn: Added this, thank you!

internal/controller/rolepoller.go
jdheyburn: I've wrapped each call in context.WithTimeout(ctx, interval) to solve unreachable or partitioned nodes.

internal/controller/rolepoller.go
jdheyburn: Agree, I changed it to pollOnce and pollOnceWithin.

internal/controller/rolepoller.go
jdheyburn: I added a log liine for any backed-off nodes.

internal/controller/rolepoller.go
jdheyburn: Agree. I changed it such that a blank role `""` is not treated as drift and the previous state remains the same.

internal/controller/rolepoller.go
jdheyburn: I added metrics counter for dropped triggers, and total triggers.

jdheyburn:
Addressed comments, back into ready for review

bjosv (approved):
Looks good!

jdheyburn:
Thanks for merging @bjosv! (I'm still catching up from OOO)

bjosv:
Ah great, I was lagging a bit regarding the merge..

PR #326, opened 2026-07-22
updated valkey version

Bloodraven21:
This PR closes <Issue #>

### Summary

Updated the valkey version from 9.0.0 -> 9.1.1

just the valkey image change no feat/behaviour change

### Checklist


bjosv (dismissed):

internal/controller/valkeycluster_controller.go
```diff
@@ -46,7 +46,7 @@ import (
 const (
 	DefaultPort           = 6379
 	DefaultClusterBusPort = 16379
-	DefaultImage          = "valkey/valkey:9.0.0"
+	DefaultImage          = "valkey/valkey:9.1.1"
```
sandeepkunusoth, line 49: so do we want the default image to be 9.1?   i thought operator should support versions>=9.0?

bjosv: I think it makes sense to use a recent valkey version as the default one.
Maybe we could have a CI step (merged to main/daily/weekly)that runs all tests with the minimal supported version?

sandeepkunusoth: i was using default image as 9.0.0 when version cannot be determined from spec and if 9.1.0+ not supported new directives are set by user in spec. i guess i need to create seperate constant for this not sure. https://github.com/valkey-io/valkey-operator/pull/307

bjosv:
I believe this PR depends on resolving #337, i.e. changing the default will roll cluster nodes (that uses the default) without any gating or ordering.

bjosv:
See https://github.com/valkey-io/valkey-operator/pull/307#discussion_r3939549783

PR #333, opened 2026-07-26
fix: re-introduce cluster members whose addresses changed after a full restart

### LINKED ISSUE #275: Persistent ValkeyCluster can fail after full cluster restart due to stale pod IPs in nodes.conf
## Summary

A persistent `ValkeyCluster` can fail to recover cleanly after a full Kubernetes cluster restart or broad Valkey pod restart when pod IPs change.

The persistent `/data/nodes.conf` retains old peer pod IPs, while the operator-generated Valkey pods announce the current ephemeral pod IP via `--cluster-announce-ip $(POD_IP)`. After all pods come back with new IPs, the Valkey cluster may keep trying to reach old pod IPs and remain stuck in `Reconciling` until the nodes are manually reintroduced with `CLUSTER MEET`.

## Environment

- valkey-operator chart: `0.2.1`
- valkey-operator image/app version: `v0.2.0`
- Valkey image: `valkey/valkey:9.1.0`
- Kubernetes: `v1.36.2`
- OS: Talos Linux 1.13.5
- Workload type: `StatefulSet`
- Persistence: enabled
- Storage: Longhorn-backed PVCs
- Cluster shape: 3 shards, 1 replica per shard, 6 pods total

## ValkeyCluster spec

```yaml
kind: ValkeyCluster
metadata:
  name: valkey
  namespace: valkey-system
spec:
  image: valkey/valkey:9.1.0
  shards: 3
  replicas: 1
  workloadType: StatefulSet
  persistence:
    size: 2Gi
    storageClassName: longhorn-valkey
    reclaimPolicy: Retain
  exporter:
    enabled: true
```

## Manual recovery

The cluster recovered after manually running CLUSTER MEET from all current pods to all current pod IPs:
```bash
for src in valkey-valkey-0-0-0 valkey-valkey-0-1-0 valkey-valkey-1-0-0 valkey-valkey-1-1-0 valkey-valkey-2-0-0 valkey-valkey-2-1-0; do
  for ip in 10.244.6.84 10.244.11.10 10.244.11.150 10.244.6.233 10.244.6.161 10.244.11.194; do
    kubectl exec -n valkey-system "$src" -c server -- valkey-cli cluster meet "$ip" 6379
  done
done
```

## Why this seems problematic

Persistence is useful for keeping Valkey node identity and data across restarts, but persisting nodes.conf with pod IPs makes full-cluster or broad pod restarts fragile in Kubernetes because pod IPs are ephemeral.
This can happen even with a highly available Kubernetes control plane if the Valkey pods are recreated and receive different IPs.

## Questions

1. Is a persistent ValkeyCluster expected to recover automatically after all pods restart with new pod IPs?
2. Should the operator use stable DNS identities for cluster announcement instead of pod IPs?
3. Should the operator detect current pod IPs and run CLUSTER MEET automatically when persisted topology contains stale addresses?
4. Is there a recommended configuration to make persistent ValkeyCluster safe across full Kubernetes cluster restarts?

matka12:
## Summary

Fixes #275 — a persistent `ValkeyCluster` never re-forms after a full restart that changes every pod IP; it stays in `Reconciling` until someone manually runs `CLUSTER MEET`.

**Root cause:** each node's persisted `nodes.conf` lists its peers' old pod IPs. A single restarted member is re-discovered through the survivors, but when *all* members restart at once there is no surviving gossip path — every node keeps dialing dead addresses and all peers stay `fail?`/`fail` forever. The operator's existing MEET phase only handles *isolated* nodes (`cluster_known_nodes <= 1`); these nodes know all their peers (at dead addresses), so nothing heals them.

## Changes

- **`ClusterState.FindStaleAddressPeers()`** (`internal/valkey/clusterstate.go`): detects peers flagged `fail`/`fail?`/`noaddr` in a node's table whose node ID belongs to a live, reachable member at a different address. `fail?` is included deliberately — pfail→fail promotion needs gossip between a majority of primaries, which is exactly what a cluster-wide address change breaks.
- **`healStaleAddressPeers()`** (`internal/controller/valkeycluster_controller.go`): new reconcile phase, runs after `promoteOrphanedReplicas` and before `forgetStaleNodes`. For each stale pair it issues `CLUSTER MEET <live-ip> <port>` from the viewer; the handshake carries the member's node ID, so the viewer rebinds the existing entry to the new address and gossip propagates from there. Emits a `StaleAddressesHealed` event and requeues.
- **`forgetStaleNodes` guard**: it matches failing entries to ValkeyNodes by pod IP, so a live member at a changed address looked like a dead node and could receive `CLUSTER FORGET` — a 60-second rejoin ban that actively fights the recovery. Entries whose ID matches a live scraped node are now skipped.

## Verification

Unit tests for the detector (full-restart, dead-peer, transient-pfail, gossip-in-progress, pending-node cases) plus live verification on a 3-shard / 1-replica cluster with persistence (k3d, `valkey/valkey:9.1.0`):

| | unpatched | patched |
|---|---|---|
| `kubectl delete pod --all` (all 6 pods, IPs change) | `cluster_state:fail`, stuck `Reconciling` 13+ min, needs manual `CLUSTER MEET` | `cluster_state:ok` + `Ready` in ~20s, no intervention |
| Data (AOF on PVC) | intact after manual heal | intact |
| Event trail | — | `StaleAddressesHealed: Re-introduced 30 peer link(s) whose addresses changed` |

The same failure is triggered by the 0.3.0 → 0.4.0 operator upgrade itself: the new auth env vars update all ValkeyNode StatefulSets in one reconcile pass, restarting every pod simultaneously — so this also makes operator upgrades on persistent clusters recover automatically.

`make test` passes (90 controller specs, valkey + api packages).


melancholictheory:
this is the right fix for the all-restart case, and it gets the two things that are easy to get wrong: sourcing the MEET target from the live cluster view (`live.Address`) rather than the stale table (the persisted `nodes.conf` is exactly what can't be trusted here), and guarding `CLUSTER FORGET` when the failing entry's node ID still belongs to a reachable member, otherwise cleanup would ban the very node you just re-introduced. the `address != peer.Address` check also keeps this from firing on a normal transient `fail?`, so including `fail?` for the all-restart case doesn't cost spurious MEETs during ordinary gossip hiccups.

one optional thing: on a full N-node restart every node's table lists every other peer as stale, so this issues on the order of N² MEETs before it settles. `CLUSTER MEET` propagates transitively, once any reachable viewer meets a node, gossip carries that node's new address to the rest, so meeting each live node once from a single viewer is enough and is gentler on a large cluster. not a correctness issue, the N² just makes more noise before it converges.

and where this sits relative to #296: the phase exists because nodes announce pod IPs, so `nodes.conf` holds addresses that die on restart. if hostname announcement lands and is in use, the persisted entries are resolvable names that survive the IP change and this heal path stops having anything to do. so it's the reactive fix for IP-mode clusters and #296 is the preventive one, complementary, and the phase stays correct (just inert) under hostname mode.

matka12:
Thanks @melancholictheory — good call on the N² MEETs. Pushed 5c88340: the heal now dedupes by the moved member's node ID, issuing one MEET per member instead of one per (viewer, peer) pair. A failed MEET doesn't mark the member as met, so another viewer retries it on the same pass. The first viewer in iteration order lists all moved peers, so the resulting MEET graph is effectively a star from that viewer — connected, and gossip converges the rest.

Re-verified on the k3d repro (3 shards / 1 replica, persistence, delete all 6 pods): 6 MEETs (was 30), still ~20s to `Ready` with data intact.

Agreed on #296: this phase is the reactive fix for IP-announcement clusters, hostname announcement is the preventive one — and if hostname mode lands, the detector's `address != peer.Address` guard means this phase simply never fires.

melancholictheory:
star from the first viewer is the clean way to do it, and 30 -> 6 on the repro is the payoff. one edge to keep in mind: if that first viewer is itself unreachable on a given pass, no star forms that round, but since an absent viewer just means the next reconcile picks another one, it self-heals on the requeue rather than getting stuck. so the only cost is an extra pass in that case, which is fine. nice turnaround.

matka12:
Thanks! Small note on that edge: it's actually cheaper than an extra pass. An unreachable viewer never makes it into the scraped state at all (`GetClusterState` drops nodes it can't connect to), so its stale pairs simply don't exist that round — the dedupe walks the remaining viewers' pairs, and since every reachable viewer's table lists all moved members, the next viewer in iteration order becomes the hub in the same pass. The requeue-and-retry case only kicks in for MEETs that fail mid-command, which don't mark the member as met.

jdheyburn:
Thanks for raising this @matka12. I want to validate it works properly with reproduction steps. You mention the upgrade from 0.3.0 -> 0.4.0, perhaps you could provide that? Would we then expect the issue to be resolved with an upgrade from 0.3.0 -> `this PR`?

matka12:
Sure — two repros, both on a scratch cluster (I used k3d, kind works the same). The minimal one first since it isolates the bug from the upgrade mechanics:

### Repro A — minimal (any operator version, ~2 min)

```bash
helm repo add valkey https://valkey.io/valkey-helm/
helm install valkey-operator valkey/valkey-operator --version 0.4.0 \
  -n valkey-operator-system --create-namespace

kubectl apply -f - <<'YAML'
kind: ValkeyCluster
metadata:
  name: test
spec:
  image: valkey/valkey:9.1.0
  shards: 3
  replicas: 1
  persistence:
    size: 1Gi
  config:
    appendonly: "yes"
YAML

# wait for state: Ready, then write a marker key
kubectl exec valkey-test-0-0-0 -c server -- valkey-cli -c set canary survives

# the trigger: every pod restarts at once, every pod IP changes
kubectl delete pod --all
```

**Unpatched result:** pods come back Ready, but every `nodes.conf` (persisted on the PVCs) lists the peers' old IPs. `CLUSTER NODES` on any pod shows all peers `fail?`/`fail` at dead addresses, `cluster_state:fail`, and the CR sits in `Reconciling` ("Waiting for replica sync") indefinitely — I gave it 13+ minutes. The existing MEET phase doesn't fire because no node is isolated (`cluster_known_nodes` = 6 on all of them; they know every peer, just at dead addresses). Manual `CLUSTER MEET <new-ip> 6379` per peer from any one pod recovers it — that's the workaround folks in #275 landed on.

**With this PR's image:** same kill, the controller logs `re-introducing peer whose address changed` and emits `StaleAddressesHealed: Re-introduced 6 peer link(s)`; `cluster_state:ok` and CR `Ready` ~20s after the pods are back, `canary` intact. No intervention.

### Repro B — the 0.3.0 → 0.4.0 upgrade as trigger

Same as A but install `--version 0.3.0` first (CR unchanged). Once Ready:

```bash
# per UPGRADE.md — CRDs first
kubectl apply --server-side -f https://raw.githubusercontent.com/valkey-io/valkey-operator/v0.4.0/config/crd/bases/valkey.io_valkeyclusters.yaml
kubectl apply --server-side -f https://raw.githubusercontent.com/valkey-io/valkey-operator/v0.4.0/config/crd/bases/valkey.io_valkeynodes.yaml

helm upgrade valkey-operator valkey/valkey-operator --version 0.4.0 -n valkey-operator-system
```

The v0.4.0 controller adds the system-users auth env (`PRIMARY_AUTH`, `VALKEY_USER`, ACL secret mounts) to every ValkeyNode's pod template in the same reconcile pass, so all six single-replica workloads roll simultaneously — which *is* repro A's trigger, no infrastructure event needed. Same deadlock follows.

### Your question — 0.3.0 → this PR directly

Yes, resolved. The simultaneous roll itself still happens (the pod-template change is inherent to the version bump), but the controller doing the rolling is the patched one, so the moment the pods are back it detects the stale pairs and re-MEETs them — the cluster re-forms on its own in seconds instead of wedging. Data on the PVCs (AOF) is untouched throughout. That was exactly my original failure case: this bug turned a routine operator upgrade on a persistent production cluster into a manual-recovery incident.

jdheyburn:
@greptile-apps

jdheyburn:
I was able to repo it and get it to work locally, so thank you!

It would be great to have an e2e test for this, but I don't want to block the PR any further since I'd like to get it out for 0.5. Can you raise an issue for that and get one created once this is merged in please?

Could you also take a look at the AI reviewer comments to see if its applicable?

matka12:
Done on both:

**AI reviewer comment** — valid catch, fixed in d3ab0c7. `CLUSTER NODES` renders IPv6 endpoints bracketed (`[fd00::2]:6379@16379`), and the last-colon slice kept the brackets, so a failing IPv6 peer at its *current* address never compared equal to the bare pod IP — misclassified as moved, spurious MEET + requeue. Endpoint host extraction now goes through `net.SplitHostPort` (after trimming `@cport` and the optional `,hostname`), which unbrackets IPv6 and leaves IPv4 untouched. Unit tests added for the endpoint parser and both IPv6 cases (current address → not stale; moved → stale).

**e2e test** — raised #352 with the scenario (create persistent cluster → kill all pods → assert `StaleAddressesHealed` + Ready + data intact, no manual MEET). I'll pick it up once this merges.

matka12:
Addressed both CodeRabbit findings in 19f641a:

- **noaddr entries skipped (major)** — correct catch: the `:0@0` placeholder parses to an empty host, so the `address != ""` guard I added for parse failures was also skipping noaddr — the one flag that by definition never has an address to compare. noaddr entries are now treated as stale whenever a live node with the same ID exists (no address comparison), with tests for both the live-match and genuinely-dead cases.
- **unbracketed IPv6 coverage (minor)** — `CLUSTER NODES` indeed reports IPv6 unbracketed (`fd00::2:6379@16379`); those already parse correctly via the last-colon fallback in `hostFromClusterNodesEndpoint` (`net.SplitHostPort` rejects them, fallback strips the port). Added parser cases and mirrored both stale-peer IPv6 scenarios with unbracketed endpoints, keeping the bracketed cases for defense in depth.

matka12:
The e2e failure here is inherited from main, not this branch: #341 granted `_operator` `+config|get` but the denied-commands e2e still expects `CONFIG GET *` to be NOPERM'd — [main's own run for that merge commit fails identically](https://github.com/valkey-io/valkey-operator/actions/runs/30701407296) (`Expected <int>: 5 to equal <int>: 6`, with the CONFIG GET output visible in the log).

Opened #355 with the test fix (move CONFIG GET to the allowed check, replace the denied slot with FLUSHALL). Once that merges I'll rebase/rerun here.

jdheyburn:
Will look to get this merged in after #338

jdheyburn (approved):
I did a test locally after #338 and it resolved from all the pods being deleted at once. Thank you for picking this up!

internal/valkey/clusterstate.go
```diff
@@ -375,6 +376,110 @@ func (n *NodeState) PrimaryIdFromSelf() string {
...
+	for _, node := range s.AllNodes() {
+		if node.Id == id {
+			return node
+		}
+	}
+	return nil
+}
+
+// hostFromClusterNodesEndpoint extracts the bare host from a CLUSTER NODES
+// endpoint field (<ip:port@cport[,hostname]>). IPv6 hosts appear bracketed
```
bjosv, line 401: I don't think CLUSTER NODES uses brackets by [default](https://github.com/valkey-io/valkey/blob/5018b12b0de2d2322a1bbf6b041c43740587c0f2/src/cluster_legacy.c#L6543), unlike valkeys logger?
Maybe it's when user sets in in `cluster-announce-ip`? (which seems to be blocked soon via https://github.com/valkey-io/valkey/pull/4055)

Maybe I'm wrong, and in that case we should reuse this function in `GetFailingNodes`

bjosv (commented):
LGTM, just a finding about the brackets.
<update> I now read in the comments that it was added as a defense

Issue #267, opened 2026-06-19
Skip the roll when a shard has no healthy replica to fail over to


melancholictheory:
Spun out of the #249 review so it doesn't block that PR.

### Context

When a primary needs a roll (config, image, or other spec change), the operator proactively fails it over to a synced replica first, then rolls the old primary. If the shard has no healthy synced replica, there is nothing to fail over to, so today the operator rolls the primary anyway: the shard loses its only writer for the duration of the roll, and recovery depends on an unplanned failover or the pod coming back.

### Proposal

When a shard has no healthy replica to fail over to, skip the roll for that primary instead of taking it down, and requeue. Proceed once a healthy replica is available. This is what @jdheyburn's Sentinel setup does today, and it matches the "don't take down the last writer" instinct.

### Why

- Avoids a self-inflicted write outage on a shard that can't fail over.
- A roll is a planned action, so deferring it until it is safe is cheaper than an unplanned single-writer takedown.
- For a single-node shard (no replica at all), it is the difference between choosing to take writes down and waiting until it is safe.

### Open questions

- How to surface the wait: a cluster-level Progressing / blocked condition with a reason and the shard(s) being held, so a deferred roll is visible rather than silent (same surfacing discussed on #228).
- Is there a case where you would still want to force the roll, for example a config that is actively crashing the primary where rolling is the fix? If so, an escape hatch or timeout.
- Definition of "healthy replica": synced (master_link_status:up) is the floor. #249 established that such a replica already reports offset >= 1, so no extra offset gate is needed.

### Refs

- #249 (proactive replica selection, where this came up)
- #265 (placement / failover epic)

melancholictheory:
following up on my own issue: most of this is already in main, i'd missed it when i filed. reconcileValkeyNode returns nodeDeferred when a node that needs a roll is the primary of its shard and has no synced replica (with replicas > 0), and reconcileValkeyNodes handles that by MEETing isolated nodes and replicating pending replicas to bring a replica back, then requeues. so the roll is skipped and retried once a replica is healthy, which is what i was after.

the one piece that's still generic is surfacing: a deferral shows up as Progressing=True / UpdatingNodes, the same as a normal roll, so from the cluster status you can't tell "waiting for a healthy replica in shard X" from "rolling normally". happy to narrow this to a specific condition reason for the deferral if that's useful, or close it as already covered, your call.

jdheyburn:
[...]

PR #338, opened 2026-07-27
feat: stage pod template rolls via Spec.WorkloadRevision

### LINKED ISSUE #337: [BUG]: Pod template changes restart every cluster node at once
### Pre-flight checks

### Bug Description

ValkeyCluster already has careful sequencing when it updates ValkeyNode specs: one node at a time, replicas before primaries, and a failover attempt before rolling a primary.

That care stops at the ValkeyNode object. Each ValkeyNode owns a single-pod StatefulSet (or Deployment) and rewrites the pod template as soon as the computed template differs from what is live. There is no coordination across nodes at that layer.

So when every node needs a new pod template at the same time, every pod restarts together. A common case is changing the Valkey image on the cluster. The same full restart also shows up when an operator upgrade changes how the pod template is built (for example defaults, probes, or sidecars), even if the ValkeyCluster spec did not change.

On a multi-node cluster that is a full outage, not a rolling update.

### Steps to Reproduce

1. Deploy a multi-node ValkeyCluster that is healthy (several shards and replicas).
2. Change something that updates every node's pod template at once. Easiest cases:
   - bump the Valkey image on the ValkeyCluster, or
   - upgrade the operator to a version that builds a different pod template for the same ValkeyNode spec.
3. Watch the StatefulSets and pods for all ValkeyNodes in that cluster.

### Expected Behaviour

Nodes come up on the new pod template in a staged way, consistent with the existing cluster roll behaviour: limited concurrency, replicas before primaries, and no simultaneous restart of the whole cluster.

### Actual Behaviour

Every ValkeyNode applies its StatefulSet or Deployment template update in the same wave. All pods restart together. The cluster loses capacity across shards at once.

### Operator version

main / recent releases (behaviour is in the ValkeyNode workload apply path; not limited to one tag)

### Kubernetes version

Reproduced on kind and on multi-node clusters (server 1.28+)

kind (repro) and production-like multi-node clusters

### Additional context

The gap is the split between:

1. **ValkeyCluster → ValkeyNode spec updates** — already staged (one at a time, replica-first, proactive failover before primary).
2. **ValkeyNode → StatefulSet/Deployment pod template** — applied immediately whenever the computed template differs, with no cluster-wide permit or sibling check.

Image changes are the example operators feel first, because the pod template image is what people change on purpose. Operator upgrades that rewrite the built template hit the same path without any ValkeyCluster spec edit.

I have a fix in progress that stages pod template applies behind a cluster-granted permit (same spirit as the existing one-at-a-time Spec rolls) and can open a PR once this issue is filed.

daanvinken:
This PR closes #337

### Summary

ValkeyCluster already stages ValkeyNode Spec updates carefully (one node at a time, replicas first, failover before primary). That care stopped at the ValkeyNode object: each node rewrote its single-pod StatefulSet/Deployment as soon as the computed pod template differed, so operator upgrades and other builder-only template changes could restart every pod at once.

This PR makes the authorized pod template a Spec field the cluster controller owns.

- New `ValkeyNode.spec.workloadRevision`: hash of the fully built pod template, set by the ValkeyCluster controller.
- Cluster-owned nodes apply a rolling template update only when the hash they compute matches `spec.workloadRevision`.
- Template advances become normal Spec rolls (same one-at-a-time / replica-first / proactive failover path). A future `rollingStrategy` only needs to change how many Specs advance per reconcile.
- First-time backfill of an empty `workloadRevision` (operator upgrade onto this feature) does **not** run proactive failover; it is bookkeeping only.
- Standalone ValkeyNodes still apply template updates immediately.
- Create path unchanged.
- `WorkloadRollPending` / `AwaitingWorkloadRevision` when the node is waiting for Spec to catch up.
- API-server probe defaults include period/timeout so user probe patches do not thrash templates.

### Implementation

- Shared builders compute the revision (`computeWorkloadRevision` / `buildNodePodTemplate`).
- Cluster sets `WorkloadRevision` on desired Spec; scrape topology only when a failover-aware roll is needed.
- Node gates rolling STS/Deployment applies on Spec match; still syncs non-template fields when the template already matches.

### Limitations

- No new user-facing roll strategy API yet.
- Stuck-holder observability (slow roll vs jammed node) is follow-up.
- Full e2e for staged template rolls is follow-up (tracking issue).

### Testing

- Unit/envtest: revision gate, preflight, revision-only no failover, probe defaults, WorkloadRollPending.
- `go test ./internal/controller/` and `make lint` pass.
- Manual kind / multi-node repro of parallel vs staged rolls described on #337.

### Checklist


internal/controller/valkeynode_controller.go
```diff
@@ -315,33 +391,249 @@ func (r *ValkeyNodeReconciler) ensureDeployment(ctx context.Context, node *valke
...
+	}
+	log.Info("updated Deployment pod template", "name", dep.Name, "desiredHash", desiredHash)
+	r.Recorder.Eventf(node, nil, corev1.EventTypeNormal, "WorkloadRollApplied", "ApplyWorkloadRoll",
+		"Applied pod template update (hash %s)", desiredHash)
+	return r.markWorkloadRollApplied(ctx, node, desiredHash)
+}
+
+// syncDeploymentWithoutRoll updates labels, owner, and Spec when the pod
+// template would not roll.
+func (r *ValkeyNodeReconciler) syncDeploymentWithoutRoll(ctx context.Context, node *valkeyiov1alpha1.ValkeyNode, dep *appsv1.Deployment, desired *appsv1.Deployment) error {
```
jdheyburn: There's a lot of similarity with syncStatefulSetWithoutRoll. Are we able to DRY these? Something like syncWorkloadWithoutRoll, that might help cut the LOC down.

jdheyburn (commented):
I did a quick scan on the review, but I don't know if I'm sold on the design yet. I would rather understand what is causing the ValkeyCluster controller to push the change out to so many nodes at once. I spent a bit of time on the original PRs to implement safe sequential rolls, so I'm interested to see what's happening.

melancholictheory:
this is the right extension. the sequencing the cluster does at the CR level was being undone the moment each node rewrote its own template, so gating the template apply behind the same permit closes a real gap. and it's good that `maybeProactiveFailoverBeforeRoll` fires for the workload path too, not just spec rolls: without that, staging alone would still restart a primary's pod as primary and drop writes. the failover-before-primary is the part that actually matters here, and you kept it.

it also composes well with #317: now that the operator owns the api-server defaults, the template only differs on genuine changes (image, real builder changes), so this stages the changes that should be staged rather than defaulting churn. the two together are the full fix for #337.

one thing to make sure is observable: the workload permit is a single cluster-wide token, so a node whose new pod never becomes Ready (bad image, failing probe) holds it indefinitely and no other node's workload roll proceeds. that's the safe behaviour, you don't want to cascade a broken rollout, but it means one stuck node wedges every other node's template roll silently. the `AwaitingRollPermit` condition helps, but it's worth distinguishing "waiting its turn" from "the holder is stuck" (the same updating-normally vs stuck-updating distinction from #267), so an operator can tell a slow rollout from a jammed one. a bounded wait or a stuck-holder event would do it.

(the DRY point on `syncStatefulSetWithoutRoll` seems right too, but that's cosmetics next to the above.)

internal/controller/valkeynode_controller.go
daanvinken: Fair! We can for sure

daanvinken:
Let's take the discussion on the bug itself to https://github.com/valkey-io/valkey-operator/issues/337#issuecomment-5104238704

internal/controller/valkeycluster_controller.go
```diff
@@ -541,7 +541,15 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNodes(ctx context.Context, clus
 	// primary last within each shard, and after an update we requeue
 	// immediately, re-scraping fresh state before any further rolls.
 	var clusterState *valkey.ClusterState
-	if anyNodeRequiresRoll(cluster, nodes, configHash) {
+	// Scrape topology when a Spec roll (including WorkloadRevision) is needed so
+	// proactive failover still runs before killing a primary.
+	aclSecret, err := r.getClusterACLSecret(ctx, cluster)
+	if err != nil {
+		// Bootstrap: secret may not exist yet; hash without ACL annotations.
```
daanvinken, line 545: When the ACL secret isn't ready the cluster hashes without template annotations, but the gate only engages when podTemplateWouldRoll is treu. On create the node bypasses the gate. It converges once the secret exists.

daanvinken (commented):
E2E as a follow up I'd say.

(Meta) comment: A big chunk of the node-controller diff isn't new behavior, it's restructuring `ensureStatefulSet`/`ensureDeployment` and the cluster-controller extractions (maybeProactiveFailoverBeforeRoll, handleUnchangedValkeyNode). 

We could split that out if this is found hard to review, let me know

I'll squash commits once this is approved.

jdheyburn:
@greptile-apps

api/v1alpha1/valkeynode_types.go
```diff
@@ -181,13 +190,22 @@ const (
 	// of Spec.Config has been successfully applied via CONFIG SET. The cluster
 	// controller blocks one-at-a-time progress until this condition is True.
 	ValkeyNodeConditionLiveConfigApplied = "LiveConfigApplied"
+	// ValkeyNodeConditionWorkloadDrift indicates the desired pod template built
+	// by the operator differs from the live StatefulSet/Deployment template, and
+	// Spec.WorkloadRevision has not yet authorized that template. Status True
+	// means a rolling update is deferred until the ValkeyCluster controller
+	// advances Spec.WorkloadRevision (one node at a time today).
+	ValkeyNodeConditionWorkloadDrift = "WorkloadDrift"
```
jdheyburn: Can you document this in docs/status-conditions.md please?

internal/controller/workload_roll.go
```diff
@@ -0,0 +1,124 @@
...
+	case valkeyiov1alpha1.WorkloadTypeDeployment:
+		dep, err := buildValkeyNodeDeployment(node)
+		if err != nil {
+			return corev1.PodTemplateSpec{}, err
+		}
+		if aclSecret != nil {
+			dep.Spec.Template.Annotations = buildPodTemplateAnnotations(node, aclSecret)
+		}
+		return dep.Spec.Template, nil
+	default:
```
jdheyburn: Can we case on StatefulSet, then error on default?

jdheyburn (commented):
I did a local test and it worked great, so thank you! I had a couple of comments.

- Can you update the PR title and description with the new setup?
- Can you check up on some of the AI code review comments?
- Can you check on the failing e2e test?
- I agree that we can park an e2e test for this, but let's capture an issue for it so we don't lose track of it

api/v1alpha1/valkeynode_types.go
daanvinken: Done!@

internal/controller/valkeycluster_controller.go
```diff
@@ -609,47 +617,37 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNode(ctx context.Context, clust
...
+	if err := r.Get(ctx, client.ObjectKeyFromObject(node), current); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nodeUnchanged, err
 		}
+		currentExists = false
+	}
+
+	needsSpecRoll := currentExists && nodeRequiresRoll(current, desired)
+
+	if deferred := r.maybeProactiveFailoverBeforeRoll(ctx, cluster, clusterState, current, needsSpecRoll); deferred {
```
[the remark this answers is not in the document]
daanvinken: Done!

daanvinken:
Thanks @jdheyburn , sorry was a bit rushed on Thursday.

- PR title/body updated for Spec.WorkloadRevision.
- e2e: ACL-hash pod queries now filter by cluster label
- Created https://github.com/valkey-io/valkey-operator/issues/362 for E2E

internal/controller/valkeycluster_controller.go
```diff
@@ -541,7 +541,15 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNodes(ctx context.Context, clus
...
-	if anyNodeRequiresRoll(cluster, nodes, configHash) {
+	// Scrape topology when a Spec roll (including WorkloadRevision) is needed so
+	// proactive failover still runs before killing a primary.
+	aclSecret, err := r.getClusterACLSecret(ctx, cluster)
+	if err != nil {
+		// Bootstrap: secret may not exist yet; hash without ACL annotations.
+		log.V(1).Info("ACL secret not ready for roll preflight, hashing without it", "err", err)
+		aclSecret = nil
+	}
+	if anyNodeRequiresFailoverAwareRoll(cluster, nodes, configHash, aclSecret) {
```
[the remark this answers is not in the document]
daanvinken: Fixed: empty WorkloadRevision only skips failover-aware handling when the live
STS/Deployment template hash already matches the authorized revision.

internal/controller/valkeycluster_controller.go
```diff
@@ -666,38 +666,91 @@ func (r *ValkeyClusterReconciler) reconcileValkeyNode(ctx context.Context, clust
...
+	if shard != nil {
+		log.Info("proactive failover before rolling primary",
+			"name", current.Name, "address", current.Status.PodIP,
+			"syncedReplicas", len(replicas))
+		if err := proactiveFailover(ctx, r.Recorder, cluster, shard, replicas); err != nil {
+			log.Info("proactive failover did not complete, proceeding with roll",
+				"name", current.Name, "err", err)
+		}
+		return false
+	}
```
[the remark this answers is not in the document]
daanvinken: Fixed: if proactiveFailover errors or times out we defer the Spec update
(requeue) instead of proceeding with CreateOrUpdate on that primary.

internal/controller/valkeynode_controller.go
```diff
@@ -146,6 +147,17 @@ func (r *ValkeyNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// Re-read after ensureWorkload may have written WorkloadDrift status.
+	if err := r.Get(ctx, req.NamespacedName, node); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	// Waiting for Spec.WorkloadRevision: rely on watches when the cluster advances
+	// Spec, with a long backoff so waiters do not spam the API.
+	if meta.IsStatusConditionTrue(node.Status.Conditions, valkeyiov1alpha1.ValkeyNodeConditionWorkloadDrift) {
+		log.V(1).Info("ValkeyNode awaiting Spec.WorkloadRevision, requeuing", "name", node.Name)
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
```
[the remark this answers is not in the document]
daanvinken: WorkloadDrift early requeue, so a node waiting on WorkloadRevision can still
clear live-config state for the cluster controller.

internal/controller/failover.go
```diff
@@ -141,15 +141,34 @@ func nodeRequiresRoll(current *valkeyiov1alpha1.ValkeyNode, desired *valkeyiov1a
...
+	c, d := current.Spec, desired.Spec
+	c.WorkloadRevision, d.WorkloadRevision = "", ""
+	c.Config, d.Config = nil, nil
+	if !equality.Semantic.DeepEqual(c, d) {
+		return true
+	}
+	// Only WorkloadRevision differs.
+	// Empty current revision is first-time backfill, not a pod template roll.
+	if current.Spec.WorkloadRevision == "" {
+		return false
```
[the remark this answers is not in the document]
daanvinken: Fixed.

api/v1alpha1/valkeynode_types.go
bjosv: Maybe we have settled with `Drift` already, but to me drift can be seen as something going wrong and is uncontrolled, while this is expected.
I believe conditions are states (Ready/Applied..) and this is a temporary state until the roll has be finalized, so an alternative condition type/name could be `WorkloadRollPending`  (or `WorkloadRollDeferred`). WDYT?

bjosv (commented):
This PR looks great to me, just a comment about the condition.
(there is also something with ACL-roll in the e2e, I have no clue yet though)

api/v1alpha1/valkeynode_types.go
jdheyburn: +1 to `WorkloadRollPending`

daanvinken: That's a good point, I agree that `drift` has a negative connotation. This is an expected, temporary wait until the cluster authorizes the template.

I'm going to renmae to `WorkloadRollPending`+ reason `AwaitingWorkloadRevision`. cc @jdheyburn

daanvinken: Oh your comment didn't load for me yet, hence the cc.

daanvinken:
About the E2E; creates a Valkey Cluster deployment (end of the happy path):

1. Read pod annotation valkey.io/internal-acl-hash
2. Delete internal-…-system-passwords
3. Secret comes back, operator logs fallback
4. Wait up to 5 minutes for that hash on items[0] to change
   → times out: hash never changes

___
Before `WorkloadRevision`, the node controller rewrote every STS template as soon as the ACL secret hash changed. Pods all rolled; the assertion passed.

Now an ACL hash change is a template change. Apply waits on Spec.WorkloadRevision, which the cluster advances one node at a time, with failover-aware handling when live template ≠ authorized hash.

Fixing by keeping the secret recreated + fallback log checks. Drop the assumption of an immediate flip on a single arbitrary pod (items[0]). Assert that ACL hash progresses for the cluster (any or all pods under the cluster
label) under staged rolls.

daanvinken:
@jdheyburn @bjosv I think we're good here now.

bjosv (approved):
LGTM and works fine in manual tests (like upgrading the operator with new default)

jdheyburn (approved):
Great thank you!

Issue #318, opened 2026-07-15
[Epic] ValkeyCluster spec.networking


daanvinken:
### Context

One place on ValkeyCluster for how clients and peers reach nodes.

Today TLS lives under `spec.networking.tls` (merged in #339). Discovery and external access still need a clear home under the same tree.

### Target shape

```yaml
spec:
  networking:
    clusterDomain: cluster.local   # default; on networking root
    tls: ...                       # present today; more detail in #360
    discovery: ...                 # in cluster endpoint announce
    external: ...                  # clients outside the cluster
```

Leave `spec.scheduling` alone. Placement is a different domain.

### Decisions

* Keep one tree: `networking.tls`, `networking.discovery`, `networking.external`.
* Put `clusterDomain` on the networking root (default `cluster.local`). CR field for now; operator-global default later is optional.
* Per-node Services are an optional primitive, not discovery DNS and not only available when `external` is set. External may reuse them.
* Do not put a full networking object on ValkeyNode. The cluster owns the API. Nodes only get the low level fields they need (for example TLS material or announce flags).
* Do not add `external.enabled`. Omit `external` when it is off. When present, use a mode enum (exact modes stay with the external design).
* Top level `spec.tls` is gone. Use `spec.networking.tls`. Migrate objects before you roll the CRD. If you leave top level `tls` only, the field is dropped and the cluster can come up without TLS (plaintext). Say that clearly in the release notes for the cut that shipped #339.
* Discovery and external both affect how nodes announce addresses. `CLUSTER SLOTS` returns one client facing view. Prefer one hostname and split DNS when external is on. If that is not available, document that in cluster clients may go through the external path (hairpin). Do not treat a simple precedence rule as a full fix.
* Gateway API can front nodes. For cluster mode, clients must still reach each node after `MOVED`. Prefer one gateway or address with per node routing (for example SNI per node hostname) and announce that matches. Do not put all shards behind one shared backend. Single primary or replication is simpler and can land first. Details stay with the external design (#276).

### Work tracking

* [x] **TLS under networking** — #339
* [x] **Discovery + clusterDomain** — #365. Related: #297, #296. DNS via STS `serviceName` → cluster headless; no per-node Services required for discovery.
* [ ] **Per-node Services (opt-in primitive)** — stable ClusterIP, Gateway backends, and other use cases. **API must not nest this only under `networking.external`.** External (#367) may consume the Services. Explore field placement; not required for #365 discovery.
[...]

PR #339, opened 2026-07-27
refactor(api)!: move ValkeyCluster TLS under spec.networking

daanvinken:
## Summary

Phase 1a of #318: introduce `spec.networking` and move TLS to `spec.networking.tls`.

**Hard break (no dual-read):** top-level `spec.tls` is removed from the API. `GetTLS()` only reads `spec.networking.tls`. There is no fallback and no deprecation warning path. When TLS is set, `certificate.secretName` is required.

```yaml
spec:
  networking:
    tls:
      certificate:
        secretName: valkey-tls
```

## Breaking upgrade / release note

Same approach as the scheduling move (v0.4.0): document in **release notes** + docs. For TLS the failure mode is worse than scheduling (security downgrade), so the wording must state the consequence, not only that the field moved.

**Migrate every ValkeyCluster to `spec.networking.tls` before rolling the new CRD/operator.**

After the CRD drops top-level `tls`, the API server prunes that field from objects already in etcd. The operator then sees no TLS config and reconciles the cluster **without TLS (plaintext)**. Clients that expected TLS fail. That is intentional 0.x behaviour; the release note is the protection, not a dual-read.

### Draft for the next operator GitHub release

Copy into the release body (do not use a soft “fields moved” line alone):

```markdown
### Breaking: TLS under `spec.networking`

- Top-level `spec.tls` is **removed**. Use `spec.networking.tls`.
- **Upgrade order:** migrate every ValkeyCluster to `spec.networking.tls`
  **before** applying the new CRD / operator image.
- **If you upgrade with only top-level `spec.tls` still set, TLS is turned
  off.** The unknown field is dropped from the object; the cluster comes
  back up serving **plaintext**. This is not a silent rename of a working
  field—migrate first, then roll CRDs.

```yaml
# before
spec:
  tls:
    certificate:
      secretName: valkey-tls

# after
spec:
  networking:
    tls:
      certificate:
        secretName: valkey-tls
```
```

Also in-repo: `docs/valkeycluster.md` (TLS section) has the same upgrade-order warning.

## Behaviour

| Spec after this change | Result |
|---|---|
| `networking.tls` set | TLS used |
| only old top-level `tls` (pre-migration) | pruned after CRD roll → **no TLS** |
| neither | no TLS |

`ValkeyNode` still uses top-level `node.Spec.TLS` as an **internal** primitive filled from the cluster (not a mirrored `NetworkingSpec`).

## Out of scope

- Hostname discovery (`networking.discovery`, `clusterDomain`) — #296 / follow-up under #318
- External access (`networking.external`) — #276
- Split-TLS / mTLS under `networking.tls` — design follow-up
- Dual-read of top-level `spec.tls` (explicitly not done; see review thread)

## Testing

- `go test ./api/v1alpha1/ ./internal/controller/`
- `make generate manifests`
- e2e TLS manifest uses `networking.tls`

Umbrella #318 stays open for discovery/external (sub-issues next).

### CODE REVIEW

File: api/v1alpha1/valkeycluster_types.go [hunk 1]
```diff
@@ -261,7 +262,14 @@ type ValkeyClusterSpec struct {
...
+	// Networking groups how clients and peers reach cluster nodes (TLS today;
+	// discovery and external access land in follow-ups under this object).
+	// +optional
+	Networking *NetworkingSpec `json:"networking,omitempty"`
+
+	// TLS is deprecated: use networking.tls. Kept for one release so existing
+	// objects are dual-read and not silently pruned on CRD upgrade. Cannot be
+	// set together with networking.tls (CEL). Will be removed in a later release.
 	// +optional
 	TLS *TLSConfig `json:"tls,omitempty"`
```
sandeepkunusoth: i think we can remove these old fields within same PR. as 0.* version is still not mature and causes breaking changes between operator upgrades?

jdheyburn: Agree, we can remove them. I did the same for the scheduling changes I made.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -287,6 +304,27 @@ type TLSConfig struct {
 	Certificate CertificateRef `json:"certificate,omitempty"`
```
jdheyburn: Do you mind adding some validation here, until we have additional TLS fields added in future?


```suggestion
	// +kubebuilder:validation:Required
	Certificate CertificateRef `json:"certificate,omitempty"`
```

api/v1alpha1/valkeycluster_types.go
jdheyburn: Similar to a previous comment, we want this to be mandatory if its defined:


```suggestion
    // +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=253
	SecretName string `json:"secretName,omitempty"`
```

melancholictheory:
this is the migration i flagged on #318, done the careful way, and for TLS specifically i'd argue against the "just remove it, we're 0.x" that came up in the thread.

the reason is the failure mode. for the scheduling move, dropping the top-level field on an already-stored cluster meant pods reschedule, annoying but recoverable. for tls it means the cluster comes back up with tls off: an encrypted, authenticated cluster silently becomes plaintext on a CRD roll, and every client that expected tls fails. that's a security downgrade triggered by an operator upgrade, not a reschedule, so it's exactly the field where the silent prune is worth avoiding.

the dual-read + CEL-rejects-both + deprecation warning you've got is the right shape for that, and it also sets the pattern for the external-access move coming next, which has the same breaking-reshape problem. if the decision is still to remove top-level tls now, the mitigation has to be loud: a release note that says migrate the CR before rolling the CRD, not just the deprecation event, since the event only helps people who are still applying manifests, not the objects sitting in etcd.

api/v1alpha1/valkeycluster_types.go
bjosv: Also agree, it will simplify this PR a lot. We give a warning/heads up in the release notes instead of in the docs as previously done then.

daanvinken:
Yeah I should have clarified but that's exactly the reason why I kept it, it's a strong security regression due to a simple operator upgrade.

I do understand the point of @jdheyburn and @sandeepkunusoth though that this operator is not meant for critical workloads yet.

I'd be also in favor of keeping backwards compatibility for at least one version and then clean up. This makes upgrades a lot easier and it's nice for us maintainers to gain some mileage with these kind of migrations (hoping we never need them ;)). 
Happy to discuss further.

api/v1alpha1/valkeycluster_types.go
daanvinken: Ok let's go with democracy. Updating PR :)

daanvinken:
See [thread](https://github.com/valkey-io/valkey-operator/pull/339#discussion_r3659212666) - moving forward with cleanup of old spec in this PR.

melancholictheory:
makes sense for 0.x, no objection to going with the removal. the one thing i'd ask of the release note then: have it say plainly that a cluster with top-level `spec.tls` set comes back up with TLS off after the CRD rolls, so it reads as "this turns TLS off, migrate first" rather than "a field moved". the failure mode is silent and security-facing, so the wording is what does the protecting here once the dual-read is gone. and worth stating the ordering explicitly: migrate the CR to `spec.networking.tls` before rolling the new CRD, since after that the value is already gone from etcd's view.

jdheyburn:
@melancholictheory Would a message in the release notes as we did for scheduling be sufficient? https://github.com/valkey-io/valkey-operator/releases/tag/v0.4.0

api/v1alpha1/valkeycluster_types.go
```diff
@@ -284,13 +294,25 @@ type TLSConfig struct {
...
+	Certificate CertificateRef `json:"certificate"`
+}
+
+// GetTLS returns the cluster TLS config from spec.networking.tls, or nil.
+func (c *ValkeyCluster) GetTLS() *TLSConfig {
+	if c == nil || c.Spec.Networking == nil {
+		return nil
+	}
+	return c.Spec.Networking.TLS
 }
```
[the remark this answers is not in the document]
jdheyburn: @greptile-apps If the user switches `spec.tls` to `spec.networking.tls`, things would continue as normal - right?

melancholictheory:
release notes are the right place, yes, but the scheduling wording isn't quite enough as a template for this one. that note says "these fields moved to spec.scheduling", which is accurate for scheduling because the failure mode is cosmetic, pods reschedule. for tls the same "it moved" sentence hides the actual effect: an upgrade with top-level `spec.tls` still set doesn't move anything, it drops it, and the cluster comes back up serving plaintext with auth off.

so the note needs one extra sentence the scheduling one didn't: not just "tls moved to spec.networking.tls", but "if you upgrade with top-level spec.tls still set, TLS is disabled on the cluster, migrate the field before rolling the new CRD". same format as v0.4.0, just with the consequence spelled out and the ordering made explicit. that's the whole ask, the mechanism (release note) is fine, it's the wording that has to name the security effect rather than describe a field move.

and to your inline question (switching `spec.tls` -> `spec.networking.tls` continues as normal): yes, an explicit migration is seamless, the gap is only the upgrade-without-migrating path, which is exactly what the note covers.

docs/valkeycluster.md
```diff
@@ -263,19 +263,22 @@ Mixing strengths (one `Preferred`, the other `Required`), or leaving one of them
...
-`tls` enables TLS for all cluster communication. The Secret must contain:
+`networking.tls` enables TLS for all cluster communication. When set, `certificate.secretName` is required. The Secret must contain:
 
 | Key | Description |
 |---|---|
 | `ca.crt` | Certificate authority |
 | `tls.crt` | Server certificate (or chain) |
 | `tls.key` | Private key for the certificate |
 
+> **Breaking (alpha):** top-level `spec.tls` is removed. Use `spec.networking.tls`. Migrate every ValkeyCluster **before** rolling the CRD; the API server prunes unknown fields on existing objects, so an unmigrated CR quietly loses TLS after the CRD upgrade.
```
jdheyburn: I think we can put this in the release notes instead, or we will remove this in 0.6?

api/v1alpha1/valkeycluster_types.go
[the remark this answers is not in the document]
daanvinken: Resolved on the description: dual-read was dropped by design after review.
GetTLS only reads networking.tls. Protection is migrate-before-CRD + release notes (stating TLS turns off if you don't migrate), not a fallback path.

docs/valkeycluster.md
daanvinken: Yeah I'm on board with stripping at 0.6.

jdheyburn (approved):
LGTM, thank you!

bjosv (approved):
Nice!

Issue #299, opened 2026-07-10
[epic] Implementation of Scheduling API


jdheyburn:
## Summary

A forward-looking `SchedulingSpec` API at `ValkeyCluster.Spec.Scheduling` (and the proposed `ValkeyCell` #226) addressing:

1. Spread pods of a shard across nodes (never colocate a primary with its replica)
2. Balance primaries across nodes (#247)
3. Balance all pods across nodes
4. Spread/balance shard pods, primaries, and all pods across availability zones
5. Pin pods to deterministic zones

This design has been revised from the [original proposal in #284](https://github.com/valkey-io/valkey-operator/discussions/284) following review. Key changes from that first draft are called out in **Changes from the original proposal** below.

## Motivation

Users want to dictate topology and placement of pods. Agreeing the long-term API shape up front lets us ship it in small, behaviour-compatible slices without repainting the CRD later.

## Design

Two fixed axes, `node` and `zone`. Each carries the same spread trio; `zone` additionally supports pinning. The controller renders the curated config into plain scheduling primitives on each ValkeyNode — ValkeyNode never receives a `scheduling` block.

```yaml
spec:
  scheduling:

    # ── escape hatches (relocated from top-level spec; passed through VERBATIM) ──
    affinity: {}
    nodeSelector: {}
    tolerations: []
    topologySpreadConstraints: []
    priorityClassName: ""              # pod priority is a scheduling input (preemption, queue order)

    # ── node axis (topologyKey: kubernetes.io/hostname, fixed) ──
    node:
      spread:
        shards:    {mode: Preferred}   # same-shard pods on distinct nodes   (default Preferred)
        primaries: {mode: Preferred}   # primaries balanced across nodes     (default Preferred)
        pods:      {mode: Disabled}    # ALL cluster pods balanced           (default Disabled)

    # ── zone axis (topologyKey: topology.kubernetes.io/zone, fixed) ──
    zone:
      spread:
        shards:    {mode: Disabled}    # same-shard pods across zones        (default Disabled)
        primaries: {mode: Disabled}    # primaries balanced across zones     (default Disabled)
        pods:      {mode: Disabled}    # ALL cluster pods across zones        (default Disabled)
      pinning:                         # nil = off; mutually exclusive with non-Disabled zone.spread.*
        zones: [az1, az2, az3]         # ordered round-robin sequence; min 1, unique; immutable while set
```

[...]

Issue #311, opened 2026-07-15
[FEATURE]: Failback primaries back to node-index=0


jdheyburn:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.

### Problem / Use case

The pod labelled node-index=0 is created as the primary. Over time during updates and pod deschedules, the pod serving as the primary may change to a node index >0.

We have been working on a number on enhancements that have a dependency on the assumption of node-index=0 being the primary:

- Pod scheduling: https://github.com/valkey-io/valkey-operator/issues/299
  - We introduced topology spread constraints on primaries at node-index=0
  - We are going to be introducing zone pinning, expecting the primary to be scheduled (and running from) the desired AZ

### Proposed solution

The check to determine if we need to failback (`node-index=0 role != primary`) can retrieve the role from the pod live as opposed to the enhancements below which look to improve role observability:

- https://github.com/valkey-io/valkey-operator/issues/261
- https://github.com/valkey-io/valkey-operator/issues/303

The failback should happen at the end of the reconcile loop, after all other maintenance has completed. Ideally failbacks should happen after some quiet period of no changes (30 mins? 60 mins?).

### Scope of change

Medium (new CRD field group / new controller flow)

### API / behaviour impact

- [ ] This change is backwards-compatible with existing ValkeyCluster CRs.
- [ ] This change introduces a breaking API change in `v1alpha1`.

### Contribution

- [ ] I would like to implement this feature myself.
- [x] I would like to help review or test this feature.

melancholictheory:
this is the piece that makes the node-index=0 assumption safe to lean on, good to see it split out. the shape (live role check, run it at the end of reconcile, gate on a quiet period) matches what we landed on. a few gotchas from doing this:

the failback itself is a graceful `CLUSTER FAILOVER` initiated from the node-0 pod, not FORCE or TAKEOVER. the default variant pauses the current primary, waits for the target to catch up, then promotes, so it's a clean handoff with no lost writes. FORCE/TAKEOVER skip that and can drop data or bump the slot epoch, they're never what you want for a planned failback.

it only works if node-0 is currently a synced replica of the shard's primary, link up and caught up. that's the same readiness check as the highest-offset failover in #249, just pinned to node-0 instead of "best replica". if node-0 is lagging or still syncing after a rejoin, the failover gets refused or stalls, so it's a precondition, not fire-and-forget.

[...]

PR #340, opened 2026-07-27
feat: Add zone scheduling axis

### LINKED ISSUE #314: [FEATURE]: Implement zone scheduling axis
### Pre-flight checks

### Problem / Use case

Implement the zone scheduling axis as outlined in phase 1c in the epic below:

- #299

### Proposed solution

The epic outlines the implementation further:

```yaml
spec:
  scheduling:
    # ── zone axis (topologyKey: topology.kubernetes.io/zone, fixed) ──
    zone:
      spread:
        shards:    {mode: Disabled}    # same-shard pods across zones        (default Disabled)
        primaries: {mode: Disabled}    # primaries balanced across zones     (default Disabled)
        pods:      {mode: Disabled}    # ALL cluster pods across zones        (default Disabled)
```

### Scope of change

Medium (new CRD field group / new controller flow)

### API / behaviour impact

### Contribution

### Additional context

_No response_

jdheyburn:
This PR closes #314.

### Summary

Implements `spec.scheduling.zone` as defined in the epic #299.

```yaml
spec:
  scheduling:
    node:
      spread:
        shards:
          mode: Disabled
        primaries:
          mode: Disabled
        pods:
          mode: Disabled
```

- Adds spec.scheduling.zone.spread.{shard, primaries, pods}, the availability-zone counterpart to node.spread.
- Opt-in: every dimension defaults to Disabled, so an operator upgrade adds no constraints and triggers no rolling restart.
- The node and zone axes are independent — they key on different topology labels (kubernetes.io/hostname vs topology.kubernetes.io/zone), so both can be enabled at once.
- Unlike the node axis, zone.spread.shard renders as a topology spread constraint (balancing), not pod anti-affinity. Anti-affinity would cap a shard's members at one per zone, making Required unsatisfiable whenever a shard has more members than there are zones; a balancing TSC spreads them as evenly as the zone count allows without that cap.

### Implementation

All three dimensions render as topology spread constraints on topology.kubernetes.io/zone:

- shard — label-selects each shard (valkey.io/shard-index), applied to every pod, so a shard's members balance across zones.
- primaries — label-selects primaries only (node-index=0), so each shard's primary-at-creation balances across zones.
- pods — label-selects every cluster pod.

Required → whenUnsatisfiable: DoNotSchedule, Preferred → ScheduleAnyway, Disabled → nothing rendered. Curated constraints are appended to any user topologySpreadConstraints (passed through verbatim), never mutating the user's list.

### Limitations

- Kubernetes allows only one topology spread constraint per {topologyKey, whenUnsatisfiable}. Because zone shard is also a TSC, all three zone dimensions compete for the two slots per zone, so at most one may be Required and one Preferred. CEL validation on ValkeyClusterSpec rejects violations on manifest apply.
- A passthrough topologySpreadConstraints entry on topology.kubernetes.io/zone that collides with an enabled zone dimension of the same whenUnsatisfiable is likewise rejected at admission.
- zone.pinning (deterministic per-pod zone assignment) is not included in this PR — it is deferred to a later phase.

### Testing

Unit tests cover the rendering helpers (effectiveZoneSpread, zoneSpreadTSCs) and the controller wiring in buildClusterValkeyNode. envtest cases cover the CEL validation (the two-slot rules, passthrough-vs-curated collisions, and the omitted-field fallback).

Manually tested on a kind cluster with zone-labelled workers:

```yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
  - role: control-plane
  - role: worker
    labels:
      topology.kubernetes.io/zone: az-a
  - role: worker
    labels:
      topology.kubernetes.io/zone: az-b
  - role: worker
    labels:
      topology.kubernetes.io/zone: az-c
```

**Applied below to spread pods in a shard across zones:**

```yaml
kind: ValkeyCluster
metadata:
  name: cluster-sample-topology-spread
spec:
  shards: 3
  replicas: 1
  scheduling:
    zone:
      spread:
        shard:
          mode: Required
```

Templates on the pods:

```yaml
  topologySpreadConstraints:
  - labelSelector:
      matchLabels:
        valkey.io/cluster: cluster-sample-topology-spread
        # Differs per shard
        valkey.io/shard-index: "0"
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
```

**Applied below to spread primaries across zones:**

```yaml
kind: ValkeyCluster
metadata:
  name: cluster-sample-topology-spread
spec:
  shards: 3
  replicas: 1
  scheduling:
    zone:
      spread:
        primaries:
          mode: Required
```

Created these pods:

```
NAME                                                  READY   STATUS    RESTARTS   AGE     IP           NODE                                NOMINATED NODE   READINESS GATES
valkey-cluster-sample-topology-spread-0-0-0           2/2     Running   0          21s     10.244.2.4   valkey-operator-zone-test-worker2   <none>           <none>
valkey-cluster-sample-topology-spread-0-1-0           2/2     Running   0          21s     10.244.1.5   valkey-operator-zone-test-worker3   <none>           <none>
valkey-cluster-sample-topology-spread-1-0-0           2/2     Running   0          21s     10.244.3.5   valkey-operator-zone-test-worker    <none>           <none>
valkey-cluster-sample-topology-spread-1-1-0           2/2     Running   0          21s     10.244.3.4   valkey-operator-zone-test-worker    <none>           <none>
valkey-cluster-sample-topology-spread-2-0-0           2/2     Running   0          21s     10.244.1.6   valkey-operator-zone-test-worker3   <none>           <none>
valkey-cluster-sample-topology-spread-2-1-0           2/2     Running   0          21s     10.244.2.5   valkey-operator-zone-test-worker2   <none>           <none>
```

Templates on the primary pods:

```yaml
  topologySpreadConstraints:
  - labelSelector:
      matchLabels:
        valkey.io/cluster: cluster-sample-topology-spread
        valkey.io/node-index: "0"
    maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
```

### Checklist


melancholictheory:
this looks right, and the two calls that matter are the ones you made: rendering the zone dimensions as topology spread constraints (balancing) rather than anti-affinity, so a shard larger than the zone count degrades to sharing a zone instead of going Pending, and keying primaries on node-index=0, same schedule-time reasoning as the node axis.

two things worth a line in the docs rather than the code:

primaries-across-zones is a placement-time property, not a maintained one. the constraint is evaluated when the pod is scheduled, so after a failover the actual primary can be node-index>0 in whatever zone it landed, and the spread reflects where primaries were placed, not where they currently are, until failback (#311) realigns. same caveat as the node axis, worth stating so nobody reads zone primaries spread as a live guarantee.

and the cross-AZ cost. spreading a shard's pods across zones puts the primary and its replicas in different AZs, so every replicated write crosses a zone boundary: added write latency and inter-AZ transfer cost. it's the right default for availability but it isn't free, worth documenting so people enable it deliberately.

test/e2e/valkeycluster_test.go
```diff
@@ -1480,6 +1480,152 @@ spec:
...
+
+			zones := []string{"e2e-az-a", "e2e-az-b"}
+			for i := 0; i < 2; i++ {
+				w := strings.TrimSpace(workers[i])
+				c := exec.Command("kubectl", "label", "node", w,
+					fmt.Sprintf("topology.kubernetes.io/zone=%s", zones[i]), "--overwrite=true")
+				o, e := utils.Run(c)
+				Expect(e).NotTo(HaveOccurred(), fmt.Sprintf("Failed to label node %s: %s", w, o))
+				defer func(node string) {
+					c := exec.Command("kubectl", "label", "node", node, "topology.kubernetes.io/zone-", "--overwrite=true")
```
bjosv: nit: not sure we need an overwrite when removing label

test/e2e/valkeycluster_test.go
```diff
@@ -1480,6 +1480,152 @@ spec:
...
+				"--selector=!node-role.kubernetes.io/control-plane",
+				"-o", "go-template={{ range .items }}{{ .metadata.name }}{{ \"\\n\" }}{{ end }}")
+			out, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to list worker nodes: %s", out))
+			workers := utils.GetNonEmptyLines(out)
+			Expect(len(workers)).To(BeNumerically(">=", 2), "expected at least two worker nodes")
+
+			zones := []string{"e2e-az-a", "e2e-az-b"}
+			for i := 0; i < 2; i++ {
+				w := strings.TrimSpace(workers[i])
```
bjosv, line 1498: Maybe overkill, but we could be nice and keep any existing zone labels
```
original, _ := utils.Run(exec.Command("kubectl", "get", "node", w, "-o", "jsonpath={.metadata.labels['topology.kubernetes.io/zone']}"))
```

then in defer
```
if original != "" {
    exec.Command("kubectl", "label", "node", node, "topology.kubernetes.io/zone="+original, "--overwrite=true")
} else {
    exec.Command("kubectl", "label", "node", node, "topology.kubernetes.io/zone-")
}
```

test/e2e/valkeycluster_test.go
```diff
@@ -1480,6 +1480,152 @@ spec:
...
+			eligibleNode, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to get worker node: %s", eligibleNode))
+			Expect(eligibleNode).NotTo(BeEmpty(), "expected at least one worker node")
+
+			cmd = exec.Command("kubectl", "label", "node", eligibleNode,
+				fmt.Sprintf("%s=true", eligibleNodeLabelKey), "--overwrite=true")
+			output, err := utils.Run(cmd)
+			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("Failed to label worker node: %s", output))
+			defer func() {
+				c := exec.Command("kubectl", "label", "node", eligibleNode, eligibleNodeLabelKey+"-", "--overwrite=true")
```
bjosv, line 1603: nit: --overwrite=true should not be needed

test/e2e/valkeycluster_test.go
```diff
@@ -1480,6 +1480,152 @@ spec:
...
+
+			By("waiting for the ValkeyCluster to become ready despite the unsatisfiable preferred spread")
+			Eventually(func(g Gomega) {
+				cr, err := utils.GetValkeyClusterStatus(zonePreferredClusterName)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+				g.Expect(cr.Status.ReadyShards).To(Equal(int32(1)))
+			}, 5*time.Minute, 2*time.Second).Should(Succeed())
+		})
 	})
```
bjosv, line 1640: We don't have a e2e test covering the node- and the zone-axis at the same time.
Maybe the unittests cover it enough?

api/v1alpha1/valkeycluster_types.go
```diff
@@ -193,6 +231,10 @@ type NodeScheduling struct {
...
...one.spread) || !( (((has(self.scheduling.zone.spread.shard) && has(self.scheduling.zone.spread.shard.mode)) ? self.scheduling.zone.spread.shard.mode : 'Disabled') == 'Required' && ((has(self.scheduling.zone.spread.primaries) && has(self.scheduling.zone.spread.primaries.mode)) ? self.scheduling.zone.spread.primaries.mode : 'Disabled') == 'Required') || (((has(self.scheduling.zone.spread.shard) && has(self.scheduling.zone.spread.shard.mode)) ? self.scheduling.zone.spread.shard.mode : 'Disabled') == 'Required' && ((has(self.scheduling.zone.spread.pods) && has(self.scheduling.zone.spread.pods.mode)) ? self.scheduling.zone.spread.pods.mode : 'Disabled') == 'Required') || (((has(self.scheduling.zone.spread.primaries) && has(self.scheduling.zone.spread.primaries.mode)) ? self.scheduling.zone.spread.primaries.mode : 'Disabled') == 'Required' && ((has(self.scheduling.zone.spread.pods) && has(self.scheduling.zone.spread.pods.mode)) ? self.scheduling.zone.spread.pods.mode : 'Disabled') == 'Required') )",message="at most one of zone.spread.shard, zone.spread.primaries, zone.spread.pods may be Required: they render duplicate topology.kubernetes.io/zone DoNotSchedule topology spread constraints"
```
bjosv, line 244: All these validations rules are hard to read in my view, maybe could help if we have a short comment above each, not sure

bjosv (commented):
LGTM and mergable in my view, just some nits/questions

test/e2e/valkeycluster_test.go
jdheyburn: The unit test `TestBuildClusterValkeyNode_MergesNodeAndZoneCuration` covers it, if there's value in having an e2e test I'll include it - lmk.

- https://github.com/jdheyburn/valkey-operator/blob/546d582ffa76d360827542967cfe6486847e0a45/internal/controller/valkeynode_resources_test.go#L918-L944

api/v1alpha1/valkeycluster_types.go
jdheyburn: I agree they are exotic and complex, I don't know any other way round it other than building out a Validation webhook for the CR - which I want to avoid unless absolutely necessary.

I'll add comments to explain what they're doing.

test/e2e/valkeycluster_test.go
jdheyburn: True, I included it for idempotency so that its makes reruns robust. I left it in but equally happy to remove it too.

test/e2e/valkeycluster_test.go
jdheyburn: Agree, removed

test/e2e/valkeycluster_test.go
jdheyburn: I've added it for good measure

bjosv (approved):
Nice!

jdheyburn:
Thanks both, I'll get this merged in. @melancholictheory lmk if there is anything else to add and I'll get it in for the next PR to implement zone pinning.

melancholictheory:
nice, glad it landed. one thing for the pinning PR, mostly about where the knob sits.

pin the cluster to a set of zones, not a shard to a zone. a shard still needs its primary and replicas spread across zones to survive a zone outage, so a knob that pins a whole shard into one zone trades zonal HA away without meaning to. "these are the zones this cluster may use" as required nodeAffinity, with the spread axis from this PR distributing within that set, composes cleanly: the pin narrows the candidate nodes, the spread balances across the pinned zones.

required is the right primitive here (a pin that's preferred isn't a pin, it spills out under pressure), so the thing worth naming is the failure mode: lose a pinned zone and the pods that were there can only reschedule onto the remaining pinned zones, or go Pending if none are left. that's the same schedule-time-only property as the spread axis, so it pairs with the degrade-to-sharing behaviour you already have.

single-zone pinning is a real want too (data locality, cross-zone traffic cost), but it's HA-off by construction, so i'd make it an explicit opt-in rather than something you back into by listing one zone, and say as much in the field docs.

Issue #335, opened 2026-07-27
[BUG]: ACL misconfiguration: _operator user can CONFIG SET all parameters


tkarger:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.
- [x] I am running a recent version of the operator (latest release or `main`) and the issue still reproduces.

### Bug Description

The _operator user's ACL grants permission to execute CONFIG SET for arbitrary parameters — including security-critical ones like requirepass and protected-mode — but explicitly denies CONFIG GET.
A user with write-but-not-read access to config can silently strip the instance's password (requirepass "") or disable protected-mode, effectively opening the server to unauthenticated access, without ever being able to confirm or audit the config state itself.
A priviledge escalation could be possible.

Mentioned this issue in #245 

### Steps to Reproduce

1. Deploy a valkey cluster
```bash
kubectl apply -f config/samples/v1alpha1_valkeycluster-with-user.yaml
```
2. Login to the instance as `_operator` user
```bash
root@valkey-cluster-sample-with-users-0-0-0:/data# valkey-cli --user _operator --pass TblTVUNQ1XDqiHpG5urKLJCRuM
````
3. Run commands to change config e.g.
```bash
127.0.0.1:6379> CONFIG GET *
(error) NOPERM User _operator has no permissions to run the 'config|get' command
127.0.0.1:6379> CONFIG SET appendonly yes
OK
127.0.0.1:6379> CONFIG SET requirepass ""
OK
127.0.0.1:6379> CONFIG SET protected-mode no
OK
127.0.0.1:6379>
```

### Expected Behaviour

* The `_operator` user's permitted `CONFIG` operations should be symmetric: if it cannot `GET` config, it should not be able to `SET` config either
* At a minimum, security-critical parameters (requirepass, protected-mode, masterauth, dir, dbfilename) should require an explicitly elevated permission level

### Actual Behaviour

* `_operator` is denied all CONFIG GET access.
* `_operator` is permitted unrestricted CONFIG SET access, including the ability to clear requirepass and disable protected-mode — both of which can result in the instance becoming unauthenticated and network-exposed.

### ValkeyCluster / ValkeyNode resource

```yaml

```

### Operator logs

```shell

```

### Valkey pod logs

```shell

```

### Operator version

main

### Kubernetes version

v1.33.1

### Kubernetes distribution / environment

kind

### Additional context

Would be happy to contribute on this topic

sandeepkunusoth:
[...]

PR #341, opened 2026-07-28
feat(acl): allow config|get for auditability

tkarger:
### Summary

The `_operator` user had only the permission to `CONFIG SET`, not to `CONFIG GET`. Due to this limitation the user wasn't able to retrieve the current configurations of the cluster and audit the state.

The `_operator` user can check the applied settings of the valkey instance and audit changes

### Implementation

The permisson for `config|get` was added to the `_operator` user.

### Testing

Tested by e2e tests

### Checklist

### DIFF
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -63,6 +63,7 @@ var (
 			"+cluster|migrateslots",      // migrate slots between shards
 			"+cluster|set-config-epoch",  // set epoch on new nodes
 			"+config|set",                // apply live config changes
+			"+config|get",                // verify applied config / audit current state
 			"+info",                      // node info and replication status
 			"+role",                      // current replication role
 		}, " "),


jdheyburn:
Thanks for raising - I left a comment on the issue on what underlying bug we're trying to solve.

- https://github.com/valkey-io/valkey-operator/issues/335#issuecomment-5101712508

melancholictheory:
this one's independently correct regardless of where the broader #335 discussion lands: read-only, no new write surface, and it closes the write-without-read audit gap. it's also useful to the operator itself, the live-config apply path can read a key back after `CONFIG SET` to confirm it took, which the current `-config|get` blocks. worth landing on its own.

jdheyburn (approved):
Thank you!

bjosv (commented):
Make sense to allow config get while we allow set.
Should we remove the part in the  top-comment that `This PR closes 335` ?

tkarger:
> Make sense to allow config get while we allow set. Should we remove the part in the top-comment that `This PR closes 335` ?

@bjosv I removed the comment

jdheyburn (approved):

PR #342, opened 2026-07-31
feat: observability into cluster scheduling state

### LINKED ISSUE #313: [FEATURE]: Observability when pods are unschedulable due to scheduling conflicts
### Pre-flight checks

### Problem / Use case

As a user who deploys ValkeyCluster, I would like the operator to inform me when the `spec.scheduling.node` configuration I have applied has caused pods to be stuck in pending state.

The operator can only do so much to prevent misconfigured scheduling requirements, see the CEL validation in #310. That PR did not manage the observability for the node axis (in order to keep the PR minimal).

### Proposed solution

We already have `PodUnschedulable` condition. We should ensure this is reused.

Perhaps consider a `SchedulingSatisfied` condition to state everything is fine.

### Scope of change

### API / behaviour impact

### Contribution

### Additional context

_No response_

jdheyburn:
This PR closes #313 

### Summary

Adds a new condition `SchedulingSatisfied` which is True when all pods have been scheduled. 

Exposes a new generic metric `valkey_operator_cluster_condition` which reports on the status of conditions.

- New `SchedulingSatisfied` status condition on `ValkeyCluster`: `True` (`AllPodsScheduled`) when every in-scope pod is scheduled, `False` (`PodsPendingScheduling`) while any pod is stuck with the scheduler's `Unschedulable` reason
- New operator metric `valkey_operator_cluster_condition{type, status}` exporting all cluster conditions in the kube-state-metrics shape (three 0/1 series per condition type)
  - Docs include an example alert for `SchedulingSatisfied`
    - the alert is configurable by the user to determine after how long they should be alerted on SchedulingSatisfied=False

### Implementation

- The condition is set in `handlePodSchedulingIssues`, which already computes the signal for `Degraded`/`Ready` so that there is no new detection logic
- The metric is a projection of `status.conditions` in `updateClusterMetrics` which by a new `ClusterConditionTypes` slice (same pattern as `ClusterStates` / `cluster_state_info`)
  - A condition absent from status reads 0 on all three series, so series identities stay stable — no per-update series deletion
- Open question from the spec: `SchedulingSatisfied` overlaps with `Ready`/`Degraded`'s `PodUnschedulable` reason
  - This PR leaves those untouched, a cluster with an unschedulable pod genuinely isn't ready

### Limitations

- The condition reports instantaneous state only; no in-operator debounce and no `AwaitingCapacity` vs `Unschedulable` distinction (deliberately: that's time-based and belongs in the alert rule)
- Alerting on `status="false"` won't fire for a cluster whose reconcile fails before the scheduling check (condition unreported = all-zeros)
  - general breakage is covered by `cluster_state_info{state="Degraded"}`
- e2e asserts the condition end-to-end but doesn't scrape `/metrics`; the condition→metric projection is covered by unit tests.

### Testing

Unit and e2e tests added.

### Checklist

@melancholictheory It would be great to get your POV on this. I'm happy to change direction if this isn't the right call. I tried to think about how it could be extended on in the future regarding the generic metric.

### DIFF
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -417,7 +417,27 @@ const (
 	// considers risky, for example a terminationGracePeriodSeconds too short for
 	// graceful failover.
 	ConditionConfigurationWarning = "ConfigurationWarning"
+	// ConditionSchedulingSatisfied reports whether every in-scope pod in the
+	// cluster is scheduled. It is False while any pod is Pending with the
+	// scheduler's Unschedulable reason. The operator reports the instantaneous
+	// state only; the "stuck for how long" threshold for alerting is left to
+	// the consumer's Prometheus rule (via its `for:` clause).
+	ConditionSchedulingSatisfied = "SchedulingSatisfied"
 )
+
+// ClusterConditionTypes lists the ValkeyCluster condition types exported by the
+// valkey_operator_cluster_condition metric. Keep in sync with the constants
+// above; a type missing from this list is still exported once present in
+// status.conditions, but its series are not pre-created at zero.
+var ClusterConditionTypes = []string{
+	ConditionReady,
+	ConditionProgressing,
+	ConditionDegraded,
+	ConditionClusterFormed,
+	ConditionSlotsAssigned,
+	ConditionConfigurationWarning,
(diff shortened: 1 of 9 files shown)


rlunar:
@greptileai

Issue #265, opened 2026-06-18
[epic] ValkeyCluster scheduling & placement (shard spread, primary balance, zones, failover)


jdheyburn:
### Purpose

Umbrella for **where `ValkeyCluster` pods run** and how that ties to **HA and failover**: shard-level placement, **even distribution of primaries across workers**, zone-aware behaviour, defaults, surfacing scheduling failures, and integration with **proactive failover** replica selection.

Use this issue as the **parent / tracking hub**; link implementation issues here as they are opened.

### Existing GitHub issues (anchor this epic to these)

| Track | Issue | Role |
|-------|-------|------|
| Shard-aware placement **design** | [#146](https://github.com/valkey-io/valkey-operator/issues/146) | Open design: API (e.g. `topologySpreadConstraints` vs operator policy), defaults, failure surfacing, interaction with user `affinity` / future zone work. |
| Primaries balanced across workers | [#247](https://github.com/valkey-io/valkey-operator/issues/247) | Cluster-level: avoid too many primaries on one worker; rebalance after node events; **complementary to #146** per that issue. |
| Proactive failover replica selection | [#249](https://github.com/valkey-io/valkey-operator/issues/249) | Replica choice for rolls / write-pause; natural tie-in for **zone-aware** primary preference later. |

### Workstreams (file or link sub-issues when ready)

1. **Close / summarize #146** — record a written decision (Option + defaults + how strict failures surface) before treating any implementation spec as final.
2. **Shard-aware placement implementation** — labels, optional default soft spread, `spec.placement` surface, conditions, docs, E2E — *blocked on #146 outcome*.
3. **Primary balance across workers (#247)** — scheduling and/or controlled failovers to rebalance; coordinate with (2) so policies do not fight.
4. **Zone-affinity for primaries** (future issue) — preferred AZ, integration with #249 for `CLUSTER FAILOVER` target selection, optional fail-back — *depends on #146 not painting the API into a corner; often ships after (2)*.

### Cross-cutting

- PDBs, single-shard / tiny-cluster behaviour, CI-friendly defaults.
- Discussion [#228](https://github.com/valkey-io/valkey-operator/discussions/228) for zone / failback context where relevant.

PR #344, opened 2026-08-01
feat: add scheduling.zone.pinning

### LINKED ISSUE #343: [FEATURE]: Implement zone pinning
### Pre-flight checks

### Problem / Use case

Implement the zone pinning axis as outlined in phase 2 in the epic below:

- #299

### Proposed solution

The epic outlines the implementation further:

```yaml
spec:
  scheduling:
    # ── zone axis (topologyKey: topology.kubernetes.io/zone, fixed) ──
    zone:
     pinning:                         # nil = off; mutually exclusive with non-Disabled zone.spread.*
        zones: [az1, az2, az3]         # ordered round-robin sequence; min 1, unique; immutable while set
```

### Scope of change

Medium (new CRD field group / new controller flow)

### API / behaviour impact

### Contribution

### Additional context

_No response_

jdheyburn:
This PR closes #343 

### Summary

Adds `spec.scheduling.zone.pinning`, an ordered list of zones that assigns every pod a zone outright instead of asking the scheduler to balance across them. Phase 2 of #265, after the node axis (#310) and the zone axis (#340). Covers the zone placement half of #146.

It's opt-in. A cluster that doesn't set it renders exactly as it did before, so upgrading the operator won't reschedule anything.

User docs are in [`docs/valkeycluster.md`](docs/valkeycluster.md) under Scheduling.

`spec.scheduling.zone.pinning.zones` takes 1 to 32 unique zone names. A pod's zone is `zones[(shardIndex + nodeIndex) % len(zones)]`, which lands on the pod as a `topology.kubernetes.io/zone` nodeSelector entry.

Each shard starts at a different offset, so a shard's pods spread across consecutive zones. Primaries end up in separate zones too, as long as you have at least as many zones as shards.

Adding shards or replicas won't move existing pods, since their indices don't change.

Admission rejects duplicate or empty zone names, pinning combined with any non-`Disabled` `zone.spread`, an in-place edit of the list, and a passthrough `nodeSelector` that sets the zone key itself.

### Implementation

Same idea as #310 and #340. The cluster controller renders the curated config into plain scheduling primitives on the ValkeyNode, and ValkeyNode itself doesn't change.

- `api/v1alpha1/valkeycluster_types.go` gets the `ZonePinning` type and a `Pinning` pointer on `ZoneScheduling`, plus four schema bounds and four CEL rules
- `internal/controller/scheduling.go` gets three helpers sitting next to the existing `effectiveZoneSpread`/`zoneSpreadTSCs`. `buildClusterValkeyNode` changes by two lines.

I picked `nodeSelector` over `nodeAffinity` since we would have to insert the new condition into every `nodeAffinity` entry. It would simpler for me to pick `nodeSelector` since that gets ANDed to everything - plus it is just one label.

`withZonePin` returns its input untouched when pinning is off, which keeps a nil `nodeSelector` nil. `DeepEqual` treats nil and an empty map as different, so returning an empty map would roll every pod on upgrade and undo #317.

Length bounds needed on `zones` for the CEL cost estimator, I don't anticipate something needing 32 zones but it's some maximum limit. I am OK to lower this. MinLength=1 to validate that we're going to be adding a label value.

Also updated the docs for `scheduling.affinity` field doc to what really happens (labels are ANDed together).

### Limitations

- If you use persistence, pin at creation time
  - Zonal volumes can't follow a pod, so switching pinning on for an existing persistent cluster strands every pod whose volume sits in the wrong zone
  - The operator can't catch that at admission because it has no idea where your PVs are
  - Removing pinning gets you back with no data loss
  - You also need a `WaitForFirstConsumer` StorageClass
  - With `Immediate` binding the volume is provisioned before the scheduler picks a node, so pinning can't work at all, even on a fresh cluster
  - The docs walk through the migration runbook and flag the case where a shard has no replicas and you'd genuinely lose data.

- The zone list is immutable while set
  - Changing it means removing pinning, reconciling, then adding it back, which on a persistent cluster is the hazard above.
  - We can always loosen this in the future

- Primaries only land in distinct zones while `shards <= len(zones)`
  - A primary is node-index 0, so its zone is `zones[shardIndex % len(zones)]`, and six shards over three zones puts 0 and 3 together
  - This is intended

Two things admission can't see, so they aren't validated: a user `affinity`, or a passthrough `topologySpreadConstraints` on the zone key, that contradicts the pin. Both show up at runtime as `Degraded` with reason `PodUnschedulable`.

### Testing

Unit tests in `scheduling_test.go` and `valkeynode_resources_test.go`. Five new functions covering the round robin at every shard/node position in a 3x2x3 example, the single zone case, pinning off, nil preservation, that the caller's map isn't mutated, key precedence, and the rendered ValkeyNode with and without pinning.

envtest in `scheduling_validation_test.go`. 14 cases against a real API server, one per rule plus the accept cases, including remove-then-re-add, an omitted `spread` block to exercise the CEL fallback, and pinning alongside `node.spread` to confirm the axes stay independent.

`make test` passes, `make lint` is clean, and `make manifests generate` produces no drift.

No e2e in this one. It would need a multi zone cluster or relabelled nodes, and the rendering is a map assignment the unit tests already cover, with envtest exercising the validation against a real API server which was the same decision in #340.

### Checklist


jdheyburn:
@greptile-apps

jdheyburn:
@coderabbitai

jdheyburn:
@greptile-apps

api/v1alpha1/valkeycluster_types.go
```diff
@@ -251,6 +296,15 @@ type ZoneScheduling struct {
...
+//
+// zone.pinning: the zone list is immutable while set (changing it reassigns nearly every pod); removal and re-adding are allowed.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.scheduling) || !has(oldSelf.scheduling.zone) || !has(oldSelf.scheduling.zone.pinning) || !has(self.scheduling) || !has(self.scheduling.zone) || !has(self.scheduling.zone.pinning) || self.scheduling.zone.pinning.zones == oldSelf.scheduling.zone.pinning.zones",message="zone.pinning.zones is immutable while set: changing it reassigns nearly every pod. Remove zone.pinning, reconcile, then re-add it with the new list"
+//
+// zone.pinning: reject a passthrough nodeSelector that sets the zone key the pinning render owns.
+// +kubebuilder:validation:XValidation:rule="!has(self.scheduling) || !has(self.scheduling.zone) || !has(self.scheduling.zone.pinning) || !has(self.scheduling.nodeSelector) || !('topology.kubernetes.io/zone' in self.scheduling.nodeSelector)",message="scheduling.nodeSelector cannot set topology.kubernetes.io/zone while zone.pinning is set: pinning renders that key itself, and the curated value would overwrite yours"
```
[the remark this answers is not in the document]
jdheyburn: I don't think we can promise everything to the user - they would need to have some understanding of the conflicting requests they are making.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -108,7 +108,9 @@ type SchedulingSpec struct {
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 
-	// Affinity to apply to the pods, overrides NodeSelector if set.
+	// Affinity to apply to the pods. Kubernetes ANDs nodeAffinity with
```
bjosv, line 111: nit: I got confused with lowercase `nodeAffinity` vs uppercase `NodeSelector`, but its the same as
 `Affinity to apply to the pods. Its nodeAffinity is ANDed with nodeSelector rather than one overriding the other: ...`
I guess.

internal/controller/scheduling.go
```diff
@@ -192,3 +193,43 @@ func zoneSpreadTSCs(clusterName string, shardIndex, nodeIndex int, shard, primar
...
+}
+
+// withZonePin returns a copy of base with the pinned zone added. base may be
+// nil and is never mutated. An empty zone returns base unchanged, so an
+// unpinned cluster keeps a nil nodeSelector rather than gaining an empty map,
+// which would re-render every pod on upgrade. If base already carries the
+// zone key, the curated value wins; admission rejects a passthrough
+// nodeSelector that sets it while pinning is on (the scheduling.nodeSelector
+// CEL rule on ValkeyClusterSpec), so that collision should be unreachable here.
+func withZonePin(base map[string]string, zone string) map[string]string {
```
bjosv: nit: Maybe add that base is a map of node-label requirements, I didn't get what base meant when just looking at the function

bjosv (dismissed):
LGTM!
I have not fully tested it manually but should be fine, added some notes/nits while reading through. Can be ignored

api/v1alpha1/valkeycluster_types.go
jdheyburn: I guess here its saying that the top-level NodeSelector on L109 can conflict with Afffinity.nodeAffinity. Let me know if I could make that clearer.

internal/controller/scheduling.go
jdheyburn: Fair - fixed in here to make it clearer: https://github.com/valkey-io/valkey-operator/pull/344/changes/046e3062cca89411edc2904243edc9b61d898dfd

daanvinken (approved):
LGTM, nice

nit: docs could say pinning follows creation-time node index (primary = node 0),
not live primary after failover; which I think is the same idea as `node.spread.primaries`.

This would be solved by https://github.com/valkey-io/valkey-operator/issues/311

bjosv (approved):
LGTM Thanks!

PR #345, opened 2026-08-01
feat: Add args configuration support to exporter

### LINKED ISSUE #331: [FEATURE]: Add args configuration support to exporter
### Pre-flight checks

### Problem / Use case

I think it would be useful to have ability to pass options to redis-exporter via `exporter: args: []`.

This allows to configure different options, like streams monitoring, etc. By default it's not required, so it shouldn't be breaking change I think.

### Proposed solution

```
exporter:
  args:
    # custom parameters, example:
    - -append-instance-role-label
```

### Scope of change

### API / behaviour impact

### Contribution

### Additional context

I can implement PR if idea looks good.

nantiferov:
This PR closes #331

### Summary

As discussed in #331, I'm moving here default args for exporter to ENV variables, so they could be overwritten if needed.

As well added ExporterSpec.Args to pass args: to exported spec.

No behaviour changes. Existing args passed to exporter replaced with ENV variable with same values.

### Implementation

* Moves current two args passed to exporter to ENV variables
* Adds ExporterSpec.Args to pass args to exporter parameters
* Adds getEnvVar to `internal/controller/valkeynode_resources_test.go` to test new functionality 
* Deepcopy and CRD files updated accordingly

### Limitations

None

### Testing

Unit tests passed

### Checklist


docs/valkeycluster.md
```diff
@@ -69,12 +69,16 @@ containers:
...
   image: oliver006/redis_exporter:v1.80.0
+  args: # optional command line arguments to exporter
+    - -ping-on-connect
   resources:
     requests:
       memory: "64Mi"
       cpu: "50m"
 ```
 
+NOTE: `exporter.args:` have higher priority than default ENV vars passed to exporter, allowing override them if needed.
```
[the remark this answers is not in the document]
nantiferov: Updated https://github.com/valkey-io/valkey-operator/commit/f75193f856dfdad667bb1ba444b6118616e63469.

docs/valkeycluster.md
```diff
@@ -69,12 +69,16 @@ containers:
...
   image: oliver006/redis_exporter:v1.80.0
+  args: # optional command line arguments to exporter
+    - -ping-on-connect
   resources:
     requests:
       memory: "64Mi"
       cpu: "50m"
 ```
 
+NOTE: `exporter.args` has higher priority than the default exporter environment variables, so it can override them when needed.
```
jdheyburn: Might be worth adding that this is the behaviour of redis_exporter and not the operator.

jdheyburn (commented):
I tested this locally and works great! Nice tidy up too.

I had a nit on the documentation, otherwise LGTM

docs/valkeycluster.md
bjosv: nit: We could mention that this is exporter image specific, or just mention oliver006/redis_exporter which this is valid for, other exporters might behave differently.

bjosv (commented):
LGTM! added a nit for discussion

nantiferov:
Thank you for review.

Clarified usage in https://github.com/valkey-io/valkey-operator/pull/345/commits/b7b93166c6bbd81420abfeed99f3582242771e1d, PTAL

jdheyburn (approved):
Thank you for raising this!

bjosv (approved):
Great, thanks!

PR #359, opened 2026-08-03
fix: grant _replication user +sync for dual-channel replication

### LINKED ISSUE #358: [BUG]: _replication system user missing +sync ACL grant, breaks dual-channel replication with NOPERM loop
## Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate. (Related but distinct: #305 / fixed by #306, which granted `+select +@write` to `_replication` for atomic slot migration. This report is about a different missing permission needed for a different replication path.)
- [x] I am running a recent version of the operator (latest release, v0.4.0) and the issue still reproduces.

## Bug Description

The operator's built-in `_replication` system user ACL grants `+psync` but not `+sync`. When `dual-channel-replication-enabled yes` is set on the Valkey server, the RDB-only side channel used by dual-channel replication issues a legacy `SYNC` command (not `PSYNC`) to start its transfer. Because `_replication` isn't granted `+sync`, this fails with `-NOPERM`, and the replica retries the full handshake from scratch roughly every second, forever — it never completes a full sync.

The relevant ACL is generated here:

https://github.com/valkey-io/valkey-operator/blob/main/internal/controller/users.go#L72-L78

```go
replicationUser: strings.Join([]string{
    "-@all +psync +replconf +ping", // the ACL rawstring for replication is taken from Valkey documentation: https://valkey.io/topics/acl/#acl-rules-for-sentinel-and-replicas
    "+cluster|syncslots",           // required for atomic slot migration
    // Today, Atomic slot migration streams the snapshot as a command stream
    // (SELECT + type-specific write commands from the AOF-rewrite)
    "+select +@write ~* -flushall -flushdb -swapdb",
}, " "),
```

`sync` is missing from the initial `-@all +psync +replconf +ping` grant.

## Steps to Reproduce

1. Install valkey-operator v0.4.0.
2. Create a ValkeyCluster with replicas, e.g.:
   ```yaml
   kind: ValkeyCluster
   metadata:
     name: valkey-cluster
   spec:
     shards: 3
     replicas: 1
   ```
3. Set `dual-channel-replication-enabled yes` on the Valkey server config (via the operator's server config passthrough).
4. Trigger any full resync of a replica (e.g. add a new replica, or force a replica to lose its cached primary — rolling restart, primary failover, or a network blip past `repl-backlog-size` all work).
5. Watch the replica's logs.

## Expected Behaviour

The replica completes dual-channel full sync and catches up to the primary.

## Actual Behaviour

The replica loops forever, retrying the handshake roughly every second, and never completes sync:

```
server 1:S 22 Jul 2026 15:22:07.571 * Dual channel replication: PSYNC is not possible, initialize RDB channel.
server 1:S 22 Jul 2026 15:22:07.574 # Dual channel replication: Received unexpected $ENDOFF response: -NOPERM User _replication has no permissions to run the 'sync' command
server 1:S 22 Jul 2026 15:22:07.574 # Dual channel sync failed with error -NOPERM User _replication has no permissions to run the 'sync' command
server 1:S 22 Jul 2026 15:22:07.574 * Dual channel replication: Aborting dual channel sync
server 1:S 22 Jul 2026 15:22:08.578 * Connecting to PRIMARY 10.44.99.13:6379
server 1:S 22 Jul 2026 15:22:08.579 * PRIMARY <-> REPLICA sync started
server 1:S 22 Jul 2026 15:22:08.581 * Partial resynchronization not possible (no cached primary)
server 1:S 22 Jul 2026 15:22:08.582 * Dual channel replication: PSYNC is not possible, initialize RDB channel.
server 1:S 22 Jul 2026 15:22:08.585 # Dual channel replication: Received unexpected $ENDOFF response: -NOPERM User _replication has no permissions to run the 'sync' command
server 1:S 22 Jul 2026 15:22:08.585 # Dual channel sync failed with error -NOPERM User _replication has no permissions to run the 'sync' command
server 1:S 22 Jul 2026 15:22:08.585 * Dual channel replication: Aborting dual channel sync
```
(repeats indefinitely, ~1s cadence)

For a cluster taking meaningful write throughput, this is a hard outage for that replica/shard — it can never catch up.

## ValkeyCluster / ValkeyNode resource

```yaml
kind: ValkeyCluster
metadata:
  name: valkey-cluster
spec:
  shards: 3
  replicas: 1
```
(Minimal repro shape)

## Valkey pod logs

See the log block above under "Actual Behaviour."

## Operator version

v0.4.0

## Additional context

- Sibling issue: #305 (fixed by #306) hit the same class of bug for the `select` command needed during atomic slot migration. Same root cause pattern (the `_replication` system user ACL not covering every command a given replication mode needs), this time for `sync`, needed by dual-channel replication.
- Suggested fix: add `+sync` to the initial command grant on `internal/controller/users.go` line 73, e.g. `"-@all +psync +replconf +ping +sync"`.
- Workaround in the meantime: don't set `dual-channel-replication-enabled yes` until this is fixed — plain `PSYNC`-based single-channel replication works fine since `_replication` already has `+psync`.

itsyuvalcohen:
This PR closes #358

### Summary

The `_replication` system user's ACL grants `+psync` but not `+sync`. When `dual-channel-replication-enabled yes` is set on the Valkey server, the RDB-only side channel issues a legacy `SYNC` (not `PSYNC`) to start its transfer, which fails with `-NOPERM`. The replica then retries the full handshake from scratch roughly every second, indefinitely, and never completes sync.

No new features or CRD changes. This only widens the built-in `_replication` system user's ACL to include `+sync`, matching the maintainer-suggested grant order from the linked issue.

### Implementation

Single-line fix in `internal/controller/users.go`: added `+sync` to the initial `-@all +psync +sync +replconf +ping` grant for `replicationUser`. No other behaviour changes.

### Limitations

The upstream Valkey ACL documentation (referenced in the code comment) doesn't currently call out `+sync` as a requirement for dual-channel replication specifically — that's a separate doc fix outside this operator repo, noted by @jdheyburn on the linked issue.

### Testing

- `make test`: all packages pass, including `internal/controller` (existing ACL/user tests unaffected since none assert the literal ACL string content).
- `make lint`: could not run to completion in my local environment — `golangci-lint` was built against go1.25 but a transitive dependency (`sigs.k8s.io/controller-tools`) now requires go1.26, causing a pre-existing toolchain version mismatch (`can't load config: the Go language version (go1.25) used to build golangci-lint is lower than the targeted Go version (1.26)`). This reproduces on a clean checkout of `main` with no changes, so it's unrelated to this PR. `go vet ./...` (part of the `make test` prerequisite chain) passes cleanly.
- Did not stand up a live cluster to reproduce end-to-end in this session; the fix is verified against the exact `-NOPERM` log signature reported in #358, and the grant matches what @jdheyburn confirmed would resolve it.

### Checklist

- [ ] Tests are added or updated. (No new test added — happy to add a regression test asserting `+sync` is present in the `_replication` ACL if maintainers want one; didn't want to presume the preferred test shape.)
- [x] Documentation files are updated. (N/A for this repo; see Limitations re: upstream Valkey docs.)
- [ ] I have run pre-commit locally (`pre-commit run --all-files` or hooks on commit) — not installed in my environment; ran `go fmt`/`gofmt` and `make test` manually instead.

### DIFF
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -71,8 +71,8 @@ var (
 		exporterUser: "-@all +@connection +memory -readonly +strlen +config|get +xinfo +pfcount -quit +zcard +type +xlen -readwrite -command +client -wait +scard +llen +hlen +get +eval +slowlog +cluster|info +cluster|slots +cluster|nodes -hello -echo +info +latency +scan -reset -auth -asking",
 
 		replicationUser: strings.Join([]string{
-			"-@all +psync +replconf +ping", // the ACL rawstring for replication is taken from Valkey documentation: https://valkey.io/topics/acl/#acl-rules-for-sentinel-and-replicas
-			"+cluster|syncslots",           // required for atomic slot migration
+			"-@all +psync +sync +replconf +ping", // the ACL rawstring for replication is taken from Valkey documentation: https://valkey.io/topics/acl/#acl-rules-for-sentinel-and-replicas; +sync is required for dual-channel replication
+			"+cluster|syncslots",                 // required for atomic slot migration
 			// Today, Atomic slot migration streams the snapshot as a command stream
 			// (SELECT + type-specific write commands from the AOF-rewrite)
 			"+select +@write ~* -flushall -flushdb -swapdb",


sandeepkunusoth (approved):

jdheyburn (approved):
Thank you for raising this!

PR #368, opened 2026-08-08
test(e2e): updated e2e tests asserting pod template rolls and version upgrade rolls are happening on one node at a time

### LINKED ISSUE #362: [TEST]: e2e coverage for staged WorkloadRevision template rolls
### Pre-flight checks

### Problem / Use case

#338 stages cluster-owned pod template updates via `ValkeyNode.spec.workloadRevision` so operator upgrades and other builder-only template changes do not restart every node at once. Behaviour is covered by unit/envtest and manual kind repro; there is no automated e2e that asserts pods do not all restart together when the template changes under a fixed ValkeyCluster Spec.

### Proposed solution

Add an e2e (or extend an existing suite) that:

1. Creates a multi-node ValkeyCluster and waits until Ready.
2. Forces a pod-template-only change without editing ValkeyCluster Spec fields the user would change for data (e.g. operator image / builder delta, or an equivalent controlled template hash change).
3. Asserts StatefulSet/Deployment template applies (or pod restarts) are staged: limited concurrency, not all nodes replaced in one wave.
4. Optionally asserts `WorkloadDrift` / `spec.workloadRevision` progress while rolling.

### Scope of change

Medium (new controller flow coverage)

### Contribution

- [ ] I would like to implement this feature myself.

### Additional context

Follow-up from review of https://github.com/valkey-io/valkey-operator/pull/338 (park e2e for the feature PR, track separately). Related: #337.

sandeepkunusoth:
### Summary

Added e2e tests asserting pod template rolls are staged one node at a time in both of these bwlo scenarios.
- exporter args change
- valkey cluster image version upgrade.

### Changes

- updated existing e2e test Context("rolliing update")

### Testing

tested locally on kind cluster

### Checklist


test/e2e/valkeycluster_test.go
```diff
@@ -1871,7 +1872,268 @@ spec:
 	Context("rolling update", func() {
 		const clusterName = "valkeycluster-rolling-e2e"
 
+		createReadyCluster := func(manifest string, readyShards int32) {
+			cmd := exec.Command("kubectl", "delete", "valkeycluster", clusterName, "--ignore-not-found=true", "--wait=true", "--timeout=5m")
+			_, _ = utils.Run(cmd)
```
bjosv, line 1877: Do we need a `--timeout` in kubectl delete here, but not in `deleteCluster` below?
In that case, should we check err for this Run(cmd)?

test/e2e/valkeycluster_test.go
```diff
@@ -1881,28 +2143,13 @@ spec:
...
-			}()
-
-			By("waiting for the cluster to become Ready")
-			Eventually(func(g Gomega) {
-				cr, err := utils.GetValkeyClusterStatus(clusterName)
-				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
-				g.Expect(cr.Status.ReadyShards).To(Equal(int32(2)))
-			}, 10*time.Minute, 5*time.Second).Should(Succeed())
+			createReadyCluster(manifest, 2)
```
bjosv, line 2146: Previously we used a 10min timeout for this (2-shard) cluster and now its 2 minutes (default),
is it enough for below 3-shard clusters on a slow CI day? Should be?

test/e2e/valkeycluster_test.go
```diff
@@ -1871,7 +1872,268 @@ spec:
...
+				return updated == expectedPods && restarted == expectedPods
+			}
+
+			Eventually(func(g Gomega) {
+				updated := updatedSTs(g)
+				if updated > 0 && updated < expectedPods {
+					sawPartial = true
+				}
+
+				cmd := exec.Command("kubectl", "get", "pods",
```
bjosv, line 2060: This code looks similar to `podUIDs`, can it be used here?

test/e2e/valkeycluster_test.go
```diff
@@ -1871,7 +1872,268 @@ spec:
...
+
+		countStatefulSetsWithServerImage := func(g Gomega, image string) int {
+			return countStatefulSetContainers(g, func(name, img string, _ []string) bool {
+				return name == "server" && img == image
+			})
+		}
+
+		assertStagedRoll := func(baselineUIDs map[string]string, expectedPods int, baselineRevs map[string]string, updatedSTs func(g Gomega) int) {
+			maxConcurrentRestarts := 0
+			sawPartial := false
```
bjosv, line 2049: alternative name for `sawPartial` can be `stagedRollObserved`?

daanvinken:
#362 targets a roll where the `ValkeyCluster` spec stays fixed and only the built pod template changes (operator upgrade or an equivalent builder change). That is the case `workloadRevision` is meant to stage.

This PR patches `spec.exporter.args` and `spec.image`. Those are user spec edits. The cluster controller already writes them onto one `ValkeyNode` at a time, and `workloadRevision` goes out in the same write ([`reconcileValkeyNode`](https://github.com/valkey-io/valkey-operator/blob/6fbed7fa7442fd97074eb0b09b31b8b49e9e0db4/internal/controller/valkeycluster_controller.go#L642-L695)). The node never has to wait on `AwaitingWorkloadRevision` I think.

I still think this is a valuable E2E, but these tests can pass even if the revision gate always allows the apply.

test/e2e/valkeycluster_test.go
sandeepkunusoth: added back timeout values during create

sandeepkunusoth:
> #362 targets a roll where the `ValkeyCluster` spec stays fixed and only the built pod template changes (operator upgrade or an equivalent builder change). That is the case `workloadRevision` is meant to stage.
> 
> This PR patches `spec.exporter.args` and `spec.image`. Those are user spec edits. The cluster controller already writes them onto one `ValkeyNode` at a time, and `workloadRevision` goes out in the same write ([`reconcileValkeyNode`](https://github.com/valkey-io/valkey-operator/blob/6fbed7fa7442fd97074eb0b09b31b8b49e9e0db4/internal/controller/valkeycluster_controller.go#L642-L695)). The node never has to wait on `AwaitingWorkloadRevision` I think.
> 
> I still think this is a valuable E2E, but these tests can pass even if the revision gate always allows the apply.

you are right this PR is not completely fixing the issue https://github.com/valkey-io/valkey-operator/issues/362. removed it from PR description. will post discussion on issue as it depends on whether we want to test operator upgrades in E2e test.

removed AwaitingWorkloadRevision check on exisiting tests added for spec changes version upgrade

Issue #367, opened 2026-08-06
networking: external access under spec.networking.external


daanvinken:
### Context

Epic #318. Clients outside the Kubernetes cluster cannot use a sharded ValkeyCluster today.

Nodes announce pod IPs. After `MOVED` or `ASK`, the client opens a new connection to that announced address. Pod IPs are not reachable from outside the cluster, so the next hop fails.

#276 is the full design for how to fix that: Services per shard, what address Valkey announces, NodePort read back, optional Gateway, and how to keep node to node gossip on internal IPs. This issue does **not** replace #276.

This issue is the **phase** for external client access under `spec.networking.external`. It records the API path and on/off rules from #318. Behaviour, field detail, and PR order stay in #276. Implementation lands under this path as stacked PRs.

### Goal

Expose external client access as `spec.networking.external`. Follow the behaviour designed in #276.

### Target API root

```yaml
spec:
  networking:
    clusterDomain: cluster.local   # #365
    discovery: ...                 # #365
    tls: ...                       # #339 / #360
    external:                      # this issue; leave out when off
      mode: ...                    # required when external is set; values from #276
      # more fields from #276 (Services, hostnames, Gateway, ...)
```

`clusterDomain` and `discovery` are owned by #365. This issue only adds `external`.

### Per-node Services (decoupled from external)

External modes (Gateway, NodePort, LoadBalancer, etc.) often need a **Service that selects exactly one node** as a backend. That primitive is useful **without** external access too (for example stable ClusterIP for in-cluster clients).

* **Creating** per-node Services must not require `spec.networking.external` to be set. Prefer an independent opt-in in the networking API (exact field name TBD; explore with #276 / maintainers).
* When `external` is on, it may **reuse** those Services as backends rather than inventing a second set of backends.
* Discovery (#365) does not create per-node Services; headless `serviceName` covers Hostname DNS.

This path is `spec.networking.external`. Do not add top level `externalAccess`. Open PRs that still use that name need retarget or replace.

### How on and off work

* **Off:** do not set `spec.networking.external` at all. No boolean `enabled: false`.
* **On:** set `spec.networking.external` and set `mode` to an enum value from #276.
* When `external` is set, `mode` is required (CEL or equivalent). Exact enum values come from #276.
* Do **not** add `external.enabled`.

### Decisions from #318 (do not reopen here)

[...]

Issue #276, opened 2026-06-25
[Design] External cluster access for client connections outside Kubernetes


scrothers:
## Summary

This proposes an `externalAccess` configuration block on ValkeyCluster that lets clients outside Kubernetes connect to a sharded cluster. The operator exposes each shard through its own Service, optionally announces a stable per-shard hostname, and (when asked) directs clients to those external endpoints using Valkey's client-facing announce settings. Throughout, node-to-node traffic (gossip and replication) stays on internal pod IPs, so the operator's existing reconciliation, which correlates `CLUSTER NODES` addresses with pod IPs, is unaffected. The work is delivered as four small, stacked PRs. Each one is independently useful and fully backward compatible: a cluster that doesn't opt in renders exactly as it does today.

## Motivation

Today a ValkeyCluster is reachable only from inside the Kubernetes cluster. Every node announces its pod IP, and the only Service is a headless `ClusterIP`. Cluster clients follow MOVED/ASK redirects, and in cluster mode those redirects carry the target node's announced address, which is a pod IP that isn't routable from outside. So even if you put a LoadBalancer in front of the pods, the first cross-shard redirect sends the client to an unreachable address and the connection fails.

This blocks a real and common deployment: applications running outside the Kubernetes cluster (a different cluster, a VM fleet, on-prem services) that need to use a Valkey cluster hosted in Kubernetes. Users have asked for it, and other Redis/Valkey operators support some form of it.

This need has been raised before: discussion [#77](https://github.com/valkey-io/valkey-operator/discussions/77) opened the idea (LoadBalancer, Ingress, NodePort, and proxy options, plus TLS termination), and design proposal [#259](https://github.com/valkey-io/valkey-operator/discussions/259) sketched per-shard Services, reading back allocated ports, publishing endpoints in status, and optional Gateway/TLSRoute SNI routing. This RFC follows that direction and fills in the details: how to keep the cluster bus internal so the reconciler is untouched, how the announce settings are rendered per node, the NodePort allocation and read-back mechanics, and the optional Gateway API path. It starts with TCPRoute rather than TLSRoute/SNI, and leaves TLSRoute as a follow-up.

[...]

PR #378, opened 2026-08-14
feat: discovery preferredEndpointType and headless serviceName

### LINKED ISSUE #365: networking: discovery preferredEndpointType and clusterDomain
### Context

Epic #318. In-cluster clients need stable announce addresses after `CLUSTER SLOTS`, especially with TLS and DNS-only certs.

Today nodes still announce like the old IP path unless we add discovery fields under `spec.networking`. Hostname work started in #296 against an older API shape. That needs to land under the tree in #318.

### Goal

Add discovery settings under `spec.networking` so operators can choose IP or hostname announce, and set the cluster DNS domain once.

### Target API

```yaml
spec:
  networking:
    clusterDomain: cluster.local   # default; on networking root
    discovery:
      preferredEndpointType: IP    # IP | Hostname; default IP
    tls: ...                       # already shipped (#339)
```

### Defaults when fields are omitted

| What is omitted | Effective behaviour |
|---|---|
| `networking` omitted | IP announce; same as today. No mass roll for discovery alone on operator upgrade. |
| `discovery` omitted | `preferredEndpointType` = `IP` |
| `preferredEndpointType` omitted | `IP` |
| `clusterDomain` omitted | `cluster.local` |
| `workloadType` omitted | `StatefulSet` (existing default). Hostname is allowed. |

### Behaviour

Announced hostname FQDN (Hostname mode only):

`<podName>.<headlessServiceName>.<namespace>.svc.<clusterDomain>`

Example: `valkey-mycluster-0-0-0.valkey-mycluster.default.svc.cluster.local` (exact resource names follow existing operator naming).

#### How per-pod DNS is obtained

Cluster-owned StatefulSets must set `spec.serviceName` to the **cluster headless** Service name (`headlessServiceName(cluster)`), not the ValkeyNode / STS resource name.

Today `serviceName` is the node resource name and is not useful for DNS (#239). Multiple 1-pod StatefulSets that share one headless as `serviceName` get resolvable per-pod names under that Service. That is enough for in-cluster Hostname announce. Do **not** rely on hand-setting `pod.spec.subdomain` as the main lever (the StatefulSet controller owns hostname / subdomain behaviour from `serviceName`).

`serviceName` is immutable on an existing StatefulSet. Changing it requires delete with orphan cascade, then recreate on the next reconcile. Document that for upgrades.

Per-node Services are **not** required for discovery DNS and are **not** created in this issue. They are a separate optional primitive (stable ClusterIP, Gateway backends, and other use cases), not something that only exists when `networking.external` is set. External (#367) may *consume* them; the API must not force external on to get them. Exact field placement is tracked on the epic / #367 follow-up, not here.

| `preferredEndpointType` | Controller does |
|---|---|
| `IP` (default) | Announce pod IP (`--cluster-announce-ip` from pod IP). Do not set managed `cluster-preferred-endpoint-type` to hostname. |
| `Hostname` | Ensure STS `serviceName` is the cluster headless Service (see above). Announce the FQDN via `--cluster-announce-hostname`. Set managed config `cluster-preferred-endpoint-type hostname`. |

`clusterDomain` must match the kubelet cluster domain. It is not hard-coded only to `cluster.local` in code paths that build FQDNs.

Whether `serviceName` is corrected for all StatefulSet nodes (recommended durable identity) or only when Hostname is selected is an implementation detail: Hostname requires the headless `serviceName`; applying the fix for all STS is preferred if the one-time orphan recreate cost is accepted and documented.

### Rolls

* IP → Hostname rolls pods (announce flags / template change as implemented).
* Hostname → IP also rolls.
* Operator upgrade with default IP: no discovery-driven mass roll for announce mode alone. Correcting immutable `serviceName` is a separate, documented one-time STS recreate if applied on upgrade.
* Existing clusters switching to Hostname need working per-pod DNS under the headless Service first (via correct `serviceName`).

### Why nest under `discovery` (not flat on `networking`)

We considered `spec.networking.preferredEndpointType` on the root next to `tls` and `clusterDomain`.

We keep it under `networking.discovery` for now so the tree stays extensible and the split stays clear:

* **`discovery`** is in-cluster announce and client rediscovery after `CLUSTER SLOTS` (this field, and later announce knobs that are not external Services).
* **`external`** is how clients outside the cluster reach shards (#276 / #367).
* **`tls`** is encryption and cert material (#339, #360).

`preferredEndpointType` is an in-cluster announce choice, not a TLS field and not external access. Nesting keeps the networking root from becoming a flat list of mixed knobs, and leaves room for more discovery fields without another rename. `clusterDomain` stays on the root because many features reuse it (FQDNs for announce, certs, later external hostnames).

If discovery never grows past one field, we can still flatten later. Prefer nest first while the API is young.

### Constraints

* `Hostname` requires `workloadType: StatefulSet` (stable pod names). Reject with CEL if not. Unset `workloadType` counts as StatefulSet, so Hostname is allowed.
* Do not put a full `networking` object on ValkeyNode. #296 mirrored networking onto the node; that is **not** the target. The cluster owns the API. The controller writes only node primitives (announce flags, STS `serviceName`, config).
* Discovery is for **in-cluster** announce and client rediscovery. External client addressing stays under `networking.external` (#276 / #367).
* For Hostname + TLS, server certs must cover the announced pod FQDNs. Typical SAN: `*.<headlessServiceName>.<namespace>.svc.<clusterDomain>`. How secrets are split or named is #360. This issue only needs docs and e2e that use a cert that matches those names.
* Hostname without TLS is allowed. Soft guard applies only when TLS is set and preferred endpoint is IP.
* Do not create per-node Services in this issue.

### Related: stale pod IPs (#275)

#275 is about persisted `nodes.conf` keeping old peer **pod IPs** after mass restart, so the cluster does not re-form until `CLUSTER MEET` (heal work in #333).

Hostname announce can give stable DNS names for clients and may help long term, but it does **not** close #275. Recovery when addresses change stays on #275 / #333. Link both ways; do not merge the issues.

### Discussion: TLS + IP announce, and soft guard

Pod IPs change. Certs that only list DNS names do not match a re-dial to a pod IP after `CLUSTER SLOTS`. Certs that list pod IPs as IP SANs break when the IP changes unless you reissue.

Hostname announce plus DNS SANs on the cert stays valid when the pod IP changes. That is the durable path for TLS in Kubernetes.

Default announce stays **IP** so clusters without TLS (and existing IP-based setups) do not change behaviour.

**Do not hard reject** `networking.tls` with `preferredEndpointType: IP` (or default IP). Valid escapes exist: clients with skip verify, rare IP SAN setups, migration windows.

**Do soft guard** instead:

* When `networking.tls` is set and preferred endpoint is IP (including default when the field is omitted), set a status condition. Do **not** set Ready to False for this reason alone. Optionally emit an Event; condition is the source of truth if both are used.
* Clear the condition when TLS is removed or preferred endpoint is Hostname.
* Message idea: with TLS and IP announce, clients that re-dial announced IPs often fail certificate name checks; prefer `preferredEndpointType: Hostname` and cert SANs for the pod FQDNs.
* Still allow apply and reconcile.

Hard CEL only for Hostname without StatefulSet (unset workloadType counts as StatefulSet).

Do not parse cert SANs from the Secret in this phase. Document the SAN expectation; e2e uses a matching cert.

### Decided: `clusterDomain` placement

* **CR field** on `networking` (default `cluster.local`), as in the target API above.
* Operator-level config as a later default when the CR omits the field is optional follow-up, not required for this phase.
* The API cannot validate that the value matches kubelet `--cluster-domain`. Document the requirement; wrong values surface as non-resolving announced names.

### Acceptance criteria

* [ ] CRD fields as above; defaults match the omit matrix
* [ ] Cluster-owned StatefulSets use headless Service as `spec.serviceName` (not the ValkeyNode resource name); Hostname path depends on that DNS
* [ ] Immutable `serviceName` change path documented (orphan delete + recreate); covered in upgrade notes if applied on existing clusters
* [ ] Controller implements the IP vs Hostname levers table (announce flags, managed preferred endpoint type, FQDN from headless + `clusterDomain`)
* [ ] Announced Hostname FQDN is `<pod>.<headless>.<namespace>.svc.<clusterDomain>`; `clusterDomain` is not hard-coded only to `cluster.local`
* [ ] CEL: Hostname + Deployment (or other non-StatefulSet) rejected; unset workloadType allows Hostname
* [ ] Soft guard: TLS + IP (default counts) sets a non-blocking condition; Ready may stay True; warning clears on Hostname or TLS off
* [ ] No per-node Services created for discovery
* [ ] Unit tests: FQDN uses non-default `clusterDomain`; CEL Hostname + Deployment; soft guard set and clear; `serviceName` is headless name
* [ ] Docs: FQDN shape, `serviceName` / DNS mechanism, wildcard SAN expectation, TLS + IP warning, omit defaults, upgrade recreate note
* [ ] e2e: multi-shard TLS cluster with Hostname; cert SANs match announced names; clients re-dial by hostname without skip verify
* [ ] #296 closed or reshaped onto this API
* [ ] Progresses #297; does **not** close NetworkPolicy or headless Service annotation work under #297

### Out of scope

* External Services, LoadBalancer, Gateway (#276 / #367)
* Creating or specifying per-node Services (opt-in primitive for ClusterIP / backends / external consumers). Not only under `networking.external`; see epic #318 and #367 notes. This issue must not create them for discovery alone.
* Multi-certificate TLS layout (#360)
* Healing stale addresses in `nodes.conf` after full restart (#275 / #333)
* NetworkPolicies and headless Service annotations or labels (remain under #297 follow-ups)
* Closing #297 as a whole
* Dual announce for internal bus vs external clients beyond what #318 already decided (hairpin / split DNS docs when external lands)

### Links

* Epic: #318
* Older umbrella: #297
* Prior PR: #296
* TLS under networking: #339
* External phase: #367
* Stale pod IPs after restart: #275
* Stale address heal PR: #333
* Split TLS secrets: #360
* External access design: #276

daanvinken:
Closes #365 (implementation; e2e still open at https://github.com/valkey-io/valkey-operator/pull/383 which I'll rebase after - but the point is proven).

Add `spec.networking.discovery.preferredEndpointType` (`IP` default, `Hostname`) and `spec.networking.clusterDomain` (default `cluster.local`). Hostname announce uses per-pod DNS under the cluster headless Service. Cluster-owned StatefulSets set `serviceName` to that headless Service so multi 1-pod STS get real FQDNs.

Hostname is the discovery mode switch: announce hostname, managed `cluster-preferred-endpoint-type`, CEL requiring StatefulSet, and clearing the non-blocking TLS+IP warning. No per-node Services. No full `networking` object on ValkeyNode (announce primitives only).

`serviceName` is immutable. The operator orphan-deletes and recreates the STS with the **live** pod template so WorkloadRevision still stages real template rolls. If the STS is gone, a pod is still there, and WorkloadRevision does not match, we requeue instead of creating the full desired spec. Orphan keeps pods; per-pod DNS is reliable after a later pod replace, not after the STS-only recreate.

Operator TLS `ServerName` uses `clusterDomain` (was hardcoded `cluster.local`).

## Testing

- `go test ./api/v1alpha1/ ./internal/controller/`
- Unit coverage for announce FQDN, headless `serviceName`, managed config, discovery fields on `buildClusterValkeyNode`, live-template migrate helper, and refuse-desired-create after failed migrate
- e2e multi-shard TLS Hostname rediscovery is not in this PR

Signed-off-by: daanvinken <daanvinken@tythus.com>

This was a shared effort of Grok 4.6 and a human


internal/controller/valkeynode_controller.go
```diff
@@ -411,7 +411,17 @@ func (r *ValkeyNodeReconciler) ensureStatefulSet(ctx context.Context, node *valk
 		if client.IgnoreNotFound(err) != nil {
 			return err
 		}
-		// Create path: no live pod to protect.
+		// After a failed serviceName recreate the STS is gone but the pod remains.
```
daanvinken, line 428: Additionally, this PR also refuses Create of the full desired STS when the object is gone, a pod is still there, and WorkloadRevision does not match (requeue). That covers Delete-ok / Create-fail on the next turn. We do not rebuild a live-template STS from the pod in that case.

internal/controller/valkeynode_controller.go
```diff
@@ -423,6 +433,33 @@ func (r *ValkeyNodeReconciler) ensureStatefulSet(ctx context.Context, node *valk
 		return r.clearWorkloadRollPending(ctx, node)
 	}
 
+	// serviceName is immutable. Orphan-delete and recreate with the live pod
```
daanvinken, line 451: This is important; `serviceName` is immutable, so we orphan-delete the StatefulSet and recreate it.

A first cut recreated with the full desired spec on the next reconcile (Get NotFound → Create desired). That skips gateRollingWorkloadUpdate. On operator upgrade every cluster-owned STS changes serviceName at once. If the desired template also changed, every pod can roll together with no staged failover.

This PR recreates with the live pod template and the new serviceName in the same reconcile, then lets WorkloadRevision gate any real template roll.

internal/controller/valkeycluster_controller.go
```diff
@@ -948,7 +971,7 @@ func (r *ValkeyClusterReconciler) getValkeyClusterState(ctx context.Context, clu
 	}
 	var tlsConfig *tls.Config
 	if tlsSpec := cluster.GetTLS(); tlsSpec != nil && tlsSpec.Certificate.SecretName != "" {
-		serverName := fmt.Sprintf("%s.%s.svc.cluster.local", headlessServiceName(cluster.Name), cluster.Namespace)
+		serverName := headlessServiceFQDN(cluster.Name, cluster.Namespace, cluster.GetClusterDomain())
```
daanvinken: Announce FQDNs use `spec.networking.clusterDomain`. Operator clients used to hardcode `…svc.cluster.local` for TLS `ServerName`.

daanvinken:
Tested locally on Kind:

- Default IP cluster: STS `serviceName` is the cluster headless Service; `--cluster-announce-ip`; no `cluster-preferred-endpoint-type` in config
- Hostname patch: `--cluster-announce-hostname $(POD_NAME).<headless>.<ns>.svc.cluster.local`; config `cluster-preferred-endpoint-type hostname`; in-cluster DNS resolves; `CLUSTER SLOTS` announces the hostname
- CEL rejects Hostname + Deployment
- TLS + IP sets `TLSEndpointWarning`; switching to Hostname clears it

Will add e2e in a follow-up PR.

daanvinken:
@greptile-apps

api/v1alpha1/valkeycluster_types.go
```diff
@@ -389,15 +392,47 @@ type ValkeyClusterSpec struct {
...
+)
+
 // NetworkingSpec groups connectivity configuration for the cluster.
-// Phase 1 of the networking API (#318): TLS only. Discovery (in-cluster
-// endpoint announcement) and external access will nest here later.
 type NetworkingSpec struct {
+	// ClusterDomain is the DNS suffix kubelet publishes Service DNS under
+	// (kubelet --cluster-domain). Used when building Hostname announce FQDNs.
+	// Must match the cluster; the API cannot validate that. Default cluster.local.
+	// +kubebuilder:default="cluster.local"
```
jdheyburn, line 416: Should the default be `cluster.local.` (period at the end) to mark it as the end of the lookup? 

> Ending a domain name with a period (.) in Kubernetes turns it into an absolute Fully Qualified Domain Name (FQDN). This forces the pod’s DNS resolver to skip appending local cluster search paths (like svc.cluster.local), preventing massive DNS latency and unnecessary lookup noise caused by the default ndots:5 setting

internal/controller/valkeynode_controller.go
```diff
@@ -423,6 +433,33 @@ func (r *ValkeyNodeReconciler) ensureStatefulSet(ctx context.Context, node *valk
...
+		}
+		if err := r.Create(ctx, recreated); err != nil {
+			return err
+		}
+		r.Recorder.Eventf(node, nil, corev1.EventTypeNormal, "StatefulSetServiceNameChange", "EnsureStatefulSet",
+			"Recreated StatefulSet %s (orphan) to change serviceName from %q to %q; pod template left unchanged until WorkloadRevision allows a roll",
+			sts.Name, from, to)
+		// Template rolls still go through the gate below on the next reconcile
+		// (or immediately if we continue). Prefer continuing so WR pending is set.
+		sts = recreated
```
jdheyburn, line 458: Could we extract part of this block into a `orphanAndRecreateStatefulSet()` function or similar?

internal/controller/valkeynode_resources.go
```diff
@@ -154,6 +154,88 @@ func mergePatchContainers(base, patches []corev1.Container) ([]corev1.Container,
...
+			},
+		},
+	}
+	if node.Spec.PreferredEndpointType != valkeyiov1alpha1.PreferredEndpointTypeHostname {
+		return []string{"--cluster-announce-ip", "$(POD_IP)"}, []corev1.EnvVar{podIPEnv}
+	}
+
+	clusterDomain := node.Spec.ClusterDomain
+	if clusterDomain == "" {
+		clusterDomain = "cluster.local"
```
jdheyburn: Not sure if because it's Friday and my brain isn't working: Doesn't this get set via `+kubebuilder:default=` already? I'm just thinking about having the default string duplicated in a couple of places, since I had a previous comment about having a period suffixed.

internal/controller/valkeynode_resources.go
```diff
@@ -154,6 +154,88 @@ func mergePatchContainers(base, patches []corev1.Container) ([]corev1.Container,
...
+	if clusterName := node.Labels[LabelCluster]; clusterName != "" {
+		return headlessServiceName(clusterName)
+	}
+	return valkeyNodeResourceName(node)
+}
+
+// headlessServiceFQDN is the Service DNS name used for TLS ServerName and similar.
+func headlessServiceFQDN(clusterName, namespace, clusterDomain string) string {
+	if clusterDomain == "" {
+		clusterDomain = "cluster.local"
```
jdheyburn: This is duplicated elsewhere too.

internal/controller/valkeynode_resources.go
```diff
@@ -154,6 +154,88 @@ func mergePatchContainers(base, patches []corev1.Container) ([]corev1.Container,
...
+	clusterName := node.Labels[LabelCluster]
+	if clusterName == "" {
+		// Standalone Hostname is not supported; fall back to IP.
+		return []string{"--cluster-announce-ip", "$(POD_IP)"}, []corev1.EnvVar{podIPEnv}
+	}
+	fqdn := fmt.Sprintf("$(POD_NAME).%s.%s.svc.%s",
+		headlessServiceName(clusterName),
+		node.Namespace,
+		clusterDomain,
+	)
```
jdheyburn: Could you reuse headlessServiceFQDN?

daanvinken:
Appreciate the thorough review, thank you

melancholictheory (commented):
Nice, the serviceName-immutability handling is the part I was most curious about and it holds together. Orphan-deleting so the pods survive, recreating the STS with the live template, and gating the real roll on WorkloadRevision means the switch doesn't restart pods on its own. The announce stays consistent through it too: `--cluster-announce-hostname` lives in the pod template, so the pod that comes up under the new `serviceName` gets both the new subdomain (from the STS) and the hostname arg in the same replace. Until that roll the pod keeps its IP announce, so there's never a window where a hostname is announced that DNS can't resolve yet. That property falling out of the ordering is what makes the migration safe.

Two things:

- The STS-missing-but-pod-exists recovery creates the full desired template (when `refuseDesiredSTSCreate` is false) rather than the live-template recreate. So a crash landing between the orphan delete and the create would apply the desired template directly. It looks like `refuseDesiredSTSCreate` gates that on WorkloadRevision so it stays staged, but a line confirming a crash mid-serviceName-change can't apply an unauthorized or un-staged template there would put it to rest.

- The IP to Hostname migration has a lot of steps (serviceName change, orphan recreate, then the staged roll that actually lands the FQDN). An e2e that flips an existing cluster and asserts the pods keep their identities through the recreate and only pick up the FQDN after the roll would be worth having. You already flagged the e2e as open, so mostly noting it as the highest-value one to write.

Un-hardcoding `clusterDomain` for the TLS ServerName, and leaving the announce fields empty on IP so existing nodes don't churn, both look right.

internal/controller/valkeynode_controller.go
daanvinken: Done!

daanvinken:
@melancholictheory thanks!

On a crash between orphan delete and create: `refuseDesiredSTSCreate` blocks Create of the desired STS while a pod exists and `WorkloadRevision` does not match the desired hash. We requeue instead. An unstaged template cannot land there 👍 

If `WorkloadRevision` already equals that hash, Create desired is the same authorization as after a successful recreate (we would apply that template in the same reconcile anyway).

Agreed the highest-value e2e is flip an existing cluster IP → Hostname: pod identities stay through the STS recreate, FQDN/announce-hostname only after the staged roll. I’ll do that in the follow-up shortly.

melancholictheory:
That clears it up, thanks. `refuseDesiredSTSCreate` gating on `pod != nil && isClusterOwned(node) && !workloadRevisionAllows(node, desiredHash)` is the guarantee I was after, and the second case is the same authorization either way.

One thing for the follow-up e2e, because it decides whether the test proves the property or only looks like it does. Pod names in a StatefulSet are stable, so the same pod name is present whether the pod survived the recreate or was replaced by a new one under that name. The observable that separates those is the pod UID, or creationTimestamp if that reads better in the assertion. Checking that the ownerReference now points at the new StatefulSet UID is worth having too, since that is what shows the orphan was adopted rather than left behind.

The reason the property holds is `statefulSetAfterServiceNameChange` copying `live.Spec.Template` onto the desired object. `serviceName` sits outside the pod template, so the recreated StatefulSet computes the same pod-template hash and its controller has no reason to replace the adopted pod. If a later change ever pulls something serviceName-derived into the template, a UID assertion is what catches it.

daanvinken:
Let me stack a PR on top with E2E 👍 

Edit: Hmm I can't stack if this branch is only on my own repo, I'll create a second PR with both changes so I can rebase when this one is merged but still validate e2e.

internal/controller/valkeycluster_controller.go
```diff
@@ -180,6 +180,18 @@ func (r *ValkeyClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		removeConditionIfReason(&cluster.Status.Conditions, valkeyiov1alpha1.ConditionConfigurationWarning, valkeyiov1alpha1.ReasonGracePeriodTooShort)
 	}
 
+	// Soft guard: TLS with IP announce (including default IP). Non-blocking.
```
bjosv, line 177: Do we need a runtime warning for this or would it be enough to have in the docs?
If a user really wants to use IP and then skip hostname verification in its own clients this emits noise and might scare unaware observers.

An alternative would be that we skip the warning when IP is explicitly configured.

bjosv (commented):
Oh, Github responds again..  Looks good, a comment about the warning though

erkules:
Help me to understand please. This PR introduces cluster-announce-hostname, which I was hoping to use for external access. While - misusing for my purpose -  spec.networking.clusterDomain can be set, the generated FQDN would  expose Kubernetes-specific DNS labels.  I would like to set the FQDN to something  like <pod>.valkey.example.org, 
Afaik does that approach also needs a service for every Pod on the top.

melancholictheory:
Your reading is right, and this PR will not get you there.

The hostname this builds is the per-pod DNS name under the cluster's headless Service, `<pod>.<headless>.<namespace>.svc.<clusterDomain>`. `clusterDomain` only replaces the last part, for clusters that do not use `cluster.local`, so pointing it at `valkey.example.org` still leaves `<headless>.<namespace>.svc` in the middle. It is a knob for the cluster suffix rather than a way to announce a domain of your own. What this gives is in-cluster discovery by name instead of by pod IP, which is what makes TLS certificates verifiable, and that is the whole scope here.

External access is a separate phase, #367, with the full design in #276: Services per shard, which address the nodes announce, NodePort read back, an optional Gateway, and keeping node to node gossip on internal IPs. That last part is why it cannot just be this field with a different domain, since the announced address is what other nodes use to talk to each other as well as what clients follow after a MOVED.

There is already code in flight for it, from @scrothers: #278 for the per-shard Services, #279 for announcing per-shard hostnames, #280 for pointing clients at the external endpoints, #281 for Gateway API TCPRoutes. Your question would land better on #367 or on those, where the people working on it will see it.

One thing worth knowing, since you were bracing for it: the design goes per shard rather than per pod, so it is one Service per shard instead of one per node.

erkules:
Thx for clarifying. And yes indeed the per-shard model does not fit to my approach.

api/v1alpha1/valkeycluster_types.go
daanvinken: IIUC the ndots issue is about lookups of the announced name, not about the this field. Announce args and TLS ServerName already build an absolute FQDN and add the trailing dot ([headlessServiceFQDN](https://github.com/valkey-io/valkey-operator/pull/378/changes#diff-4335bd43eee11330a0301e5799668b129821638703bbe7c310e6de4b04a67b4bR201)).

internal/controller/valkeycluster_controller.go
daanvinken: I think we should keep the runtime condition, including when IP is set on purpose.

The failure is the same either way: TLS is on, `CLUSTER SLOTS` still hands out pod IPs, clients that re-dial often fail name checks. Writing `preferredEndpointType: IP` does not makethat safer. It makes the choice explicit.

The escape is already there: status Read=true stays. Skip-verify clients still work. The condition is a warning likely worth having, instead of hiding it, where docs-only is easy to miss until a client breaks.

daanvinken:
https://github.com/valkey-io/valkey-operator/pull/383 proves functionality through E2E. I plan to rebase that once this is merged.
cc @bjosv @jdheyburn @melancholictheory 
Thanks again all

melancholictheory:
Greptile flagged this on #383 and it belongs here, since this is the PR that lands the code. It also sits in the spot I called right in my first pass, so I am correcting that too.

`buildClusterValkeyNode` leaves `ClusterDomain` empty unless `PrefersHostnameAnnounce()`, to avoid a needless spec churn. The node controller reads exactly that field to build the TLS ServerName: `buildNodeClientOption` calls `headlessServiceFQDN(clusterName, node.Namespace, node.Spec.ClusterDomain)`, and `headlessServiceFQDN` substitutes `DefaultClusterDomain` when the domain is empty. So a TLS cluster with `clusterDomain: corp.local` and the default IP discovery verifies against `<headless>.<ns>.svc.cluster.local.` while its certificate is issued for the configured domain.

The two controllers end up disagreeing about the same cluster. `getValkeyClusterState` builds its ServerName from `cluster.GetClusterDomain()` and gets the configured domain, while the node controller falls back without saying so.

Propagating `ClusterDomain` unconditionally looks cheap. It only reaches the pod template through the Hostname announce argument, so in IP mode it changes the ValkeyNode spec without changing the template hash: one spec write per node, no roll. That is a smaller price than a ServerName that cannot match the certificate. The gate still makes sense for `PreferredEndpointType`, which is the field that actually decides what the pod announces.

daanvinken:
@melancholictheory thanks
Done in [`dab09c2`](https://github.com/valkey-io/valkey-operator/commit/dab09c24b61d6c933ad5e011c14ae13683ad21cc): ClusterDomain is always copied onto the node, including IP announce.

After merging #409, operator TLS verify uses `tls.serverName` when set, else the headless Service FQDN under that clusterDomain. No trailing dot . The node client uses `ValkeyNode.spec.tls.serverName`, so cluster scrape and node client stay aligned

internal/controller/valkeynode_controller.go
```diff
@@ -463,6 +492,37 @@ func (r *ValkeyNodeReconciler) ensureStatefulSet(ctx context.Context, node *valk
...
+	policy := metav1.DeletePropagationOrphan
+	if err := r.Delete(ctx, live, &client.DeleteOptions{PropagationPolicy: &policy}); err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			return nil, err
+		}
+	}
+	if err := controllerutil.SetControllerReference(node, recreated, r.Scheme); err != nil {
+		return nil, err
+	}
+	if err := r.Create(ctx, recreated); err != nil {
```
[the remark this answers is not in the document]
daanvinken: narrow race, should fix, not in this PR.

docs/valkeycluster.md
```diff
@@ -346,16 +346,38 @@ If a pod cannot be placed in its pinned zone — no capacity, a `nodeSelector`/`
...
+#### Discovery (in-cluster announce)
+
+`networking.discovery.preferredEndpointType` controls how nodes advertise themselves after `CLUSTER SLOTS`:
+
+| Value | Behaviour |
+|---|---|
+| `IP` (default) | Announce pod IPs. Same as historical behaviour. |
+| `Hostname` | Announce `<pod>.<headlessService>.<namespace>.svc.<clusterDomain>`. Requires `workloadType: StatefulSet` (or omit for the default). |
+
+Cluster-owned StatefulSets use the **cluster headless Service** as `spec.serviceName` so multi 1-pod STS get real per-pod DNS under that Service.
```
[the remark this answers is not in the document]
daanvinken: nit

internal/controller/valkeynode_controller.go
```diff
@@ -883,7 +943,7 @@ func (r *ValkeyNodeReconciler) buildNodeClientOption(ctx context.Context, node *
 		secretName := node.Spec.TLS.Certificates.Server.SecretName
 		serverName := ""
 		if clusterName, ok := node.Labels[LabelCluster]; ok {
-			serverName = fmt.Sprintf("%s.%s.svc.cluster.local", headlessServiceName(clusterName), node.Namespace)
+			serverName = headlessServiceFQDN(clusterName, node.Namespace, node.Spec.ClusterDomain)
 		}
```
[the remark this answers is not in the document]
daanvinken: stale after #409

internal/controller/valkeynode_controller.go
```diff
@@ -463,6 +492,37 @@ func (r *ValkeyNodeReconciler) ensureStatefulSet(ctx context.Context, node *valk
...
+			return nil, err
+		}
+	}
+	if err := controllerutil.SetControllerReference(node, recreated, r.Scheme); err != nil {
+		return nil, err
+	}
+	if err := r.Create(ctx, recreated); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return nil, errTransientRequeue
+		}
```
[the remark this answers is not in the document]
daanvinken: `AlreadyExists` here is the old StatefulSet still terminating. Next reconcile `Get`s that object and retries the orphan.

If Delete completed and Create never landed, `Get` is NotFound and `refuseDesiredSTSCreate` requeues. That is intentional: we will not Create the full desired template and skip WorkloadRevision. The pod stays. Migration continues when WorkloadRevision matches.

Not fixing that race in this PR.

internal/controller/valkeynode_controller.go
```diff
@@ -417,7 +425,18 @@ func (r *ValkeyNodeReconciler) ensureStatefulSet(ctx context.Context, node *valk
...
+		// Do not Create full desired (that skips WorkloadRevision). Requeue instead.
+		pod, err := r.getPod(ctx, node)
+		if err != nil {
+			return err
+		}
+		desiredHash := podTemplateRollHash(desired.Spec.Template)
+		if refuseDesiredSTSCreate(node, pod, desiredHash) {
+			log.Info("StatefulSet missing while pod exists; requeue without applying unauthorized template",
+				"name", desired.Name, "desiredHash", desiredHash, "specRevision", node.Spec.WorkloadRevision)
+			return errTransientRequeue
```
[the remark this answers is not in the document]
daanvinken: Same as https://github.com/valkey-io/valkey-operator/pull/378#discussion_r3914669223

`AlreadyExists` is the old StatefulSet still terminating. Next reconcile `Get`s it and retries the orphan.

If the StatefulSet is already gone, `refuseDesiredSTSCreate` requeues on purpose. We will not Create the full desired template and skip WorkloadRevision. The pod stays. Migration continues when WorkloadRevision matches.

Not fixing that race in this PR.

@greptile-apps you are repeating the same comment.

bjosv (approved):
LGTM, we can address any additional comments or findings in trailing PRs

melancholictheory (commented):
Re-reviewed at `a24286d`. The race on the orphan window is already covered above and you have called it deferred, so I am not adding to that. Two things I checked instead, since the e2e proving this lives in #383 and is not merged yet.

**The migration really is non-disruptive.** I reproduced the mechanism on its own, outside the operator: a StatefulSet orphan-deleted and recreated with a different `serviceName` and a byte-identical template.

```
                          before              after
pod uid                   e0101a4c            e0101a4c      not replaced
controller-revision-hash  snmig-85dcc844d9    snmig-85dcc844d9
sts updateRevision        snmig-85dcc844d9    snmig-85dcc844d9
sts currentRevision                           snmig-85dcc844d9
pod owner                 old sts             new sts (adopted)
```

The single ControllerRevision survived the orphan and was re-adopted, and `readyReplicas` came back to 1 without the pod restarting. So copying the live template in `statefulSetAfterServiceNameChange` does what the comment claims: the revision name is derived from the template, the template does not move, so the adopted pod is already on the revision the new StatefulSet wants and nothing rolls.

**The clusterDomain and TLS split holds end to end.** Computed values on this head:

```
tlsServerName default        valkey-mycluster.valkey.svc.cluster.local
tlsServerName corp.local     valkey-mycluster.valkey.svc.corp.local
tlsServerName "corp.local."  valkey-mycluster.valkey.svc.corp.local
announce FQDN                valkey-mycluster.valkey.svc.corp.local.
```

Then through the CRD that #409 merged, which is where this could have bitten:

```
serverName: valkey-mycluster.valkey.svc.corp.local    accepted
serverName: valkey-mycluster.valkey.svc.corp.local.   rejected
```

So announce keeps the absolute form and the verify name is a DNS-1123 subdomain, a trailing dot on `clusterDomain` input is normalised away, and the operator writes a value its own CRD accepts. That was the collision I was worried about on #409 and it is closed.

One note for whenever the deferred race does get its own issue, since it is cheap to record now: in that window the pod is not merely unmanaged, it has no owner at all. After the orphan delete its `ownerReferences` is empty, so "the pod stays" holds only as long as nothing removes it. I deleted one in that state and nothing brought it back, which is the expected behaviour rather than a surprise, but it is the property that decides how urgent the follow-up is.

daanvinken:
Follow-ups from this PR:

- #424 recover live-template STS Create after serviceName migrate race
- #425 announced Hostname FQDN should not use a trailing dot

Migrate e2e is #383.

Issue #371, opened 2026-08-10
Improve CI turnaround time


bjosv:
### Context
Our CI feedback loop takes some time. The two longest jobs are E2E Tests and Publish, both largely parallelizable or cacheable. This tracks a series of proposed incremental improvements.

### Measured baseline

| Job | Time |
|---|---|
| Publish | ~25 min |
| E2E Tests (serial) | ~20 min |
| Tests | ~3 min |

E2E per-run breakdown: Kind create ~1m23s, operator docker-build ~1m32s, kind load + deploy ~30s, Go compile (cold) ~89s, test execution ~11 min (in serial).

Publish: arm64 go build under QEMU emulation ~23 min vs ~2.5 min native; emulation is the whole critical path.

#### Step 1 - Native-arch publish  (#366)

Build each arch on its native runner (amd64 on ubuntu-24.04, arm64 on ubuntu-24.04-arm) in parallel, merge by digest, instead of emulating arm64.

Measured: 3m 58s, down from ~25 min.

#### Step 2 - Cache Go build outputs + drop redundant go mod tidy  (proposed, #372)

`actions/setup-go`'s built-in cache saves only once per go.mod hash and won't update it on later runs unless go.mod changes, so it never keeps a warm build cache.
Can be replaced with an `actions/cache` step (keyed on .go-version + go.sum, 
github.sha so it saves every run, restore-keys to reuse the latest). We can also drop `go mod tidy` from the test/e2e jobs since tidiness is already verified by the mod-tidy job in `lint.yml`.

PoC measured (with warm cache):
| Job | Before | After |
|---|---|---|
| Tests | 3m 00s | 1m 54s (~37%) |
| E2E | 20m 38s | 18m 31s (~10%) |

Caching mainly helps the compile-bound Tests job; E2E is dominated by Kind setup, spec runtime, and docker-build, which caching doesn't touch. Workflow-only change.

#### Step 3 — Parallelize E2E via a label matrix + build the image once  (proposed, #373)

The bulk of E2E is ~11 min of serial test execution plus per-run setup.
Proposed changes:

1. **Label matrix**
Run the suite as a GitHub Actions matrix, each job covering a subset of tests via Ginkgo `--label-filter`, so subsets run in parallel on separate runners. The label buckets are defined in one place, and labels not specified are run in a catch-all job for all remaining test.
We should also include some verification that no test lands in two buckets, and that none are missed.
2. **Build the operator image once**
[...]

PR #379, opened 2026-08-14
test(e2e): assert cluster teardown against the cluster the spec deletes

melancholictheory:
Found while looking at the e2e suite for #371. There is no separate issue for it, so I can open one if the project prefers a PR per issue.

### Summary

The teardown spec in the `ValkeyCluster` Ordered container asserts on a different cluster than the one it deletes, so it passes while the cluster it claims to have removed is still running.

Test only. No operator behaviour changes.

### Implementation

"deletes the Valkey Cluster deployment" deletes `config/samples/v1alpha1_valkeycluster.yaml`, which is `cluster-sample`, then asserts on the shared `valkeyClusterName`. That variable is declared once for the whole Ordered container and reassigned by the scale-out and scale-in specs that run before it (lines 669 and 792 on `main`), so by the time the teardown spec runs it holds `valkeycluster-scalein`. The scale-in spec already deleted that cluster in its own `defer` with `--wait=false`.

All four `Eventually` blocks (CR, Service, ConfigMap, ValkeyNodes) therefore check a cluster that is already gone, and removal of `cluster-sample` is never verified. Under a label-filtered run, which #373 is heading toward, the variable would be empty and the assertions would pass vacuously again.

This names the cluster in the spec so the assertions describe what the spec actually deletes.

### Limitations

This fixes the one spec. The wider coupling is still there: `valkeyClusterName` is shared across the container and reassigned by three specs, which is what makes the container impossible to split by label. I wrote that up in #371 and can follow up with the decoupling if the approach looks right.

### Testing

Verified on a local k3d cluster, running the assertions exactly as the spec does.

With `cluster-sample` present and its Service, ConfigMap and six ValkeyNodes all live, the old assertions (against `valkeycluster-scalein`) pass 4 out of 4, which is the bug: the spec is green while the cluster it should be checking is still up.

After `kubectl delete -f config/samples/v1alpha1_valkeycluster.yaml`, the new assertions (against `cluster-sample`) pass 4 out of 4.

The cascade does not depend on the operator: the ValkeyCluster carries no finalizers, and the Service, ConfigMap and ValkeyNodes all have `ownerReferences` pointing at it, so the API server garbage collects them.

### Checklist


jdheyburn (approved):
Good find, thank you!

aamirpatel (approved):
test/e2e/valkeycluster_test.go: Consider deriving the cluster name from the sample YAML rather than hard-coding "cluster-sample" in the test. This will prevent drift if the sample name changes (e.g., parse the metadata.name field from config/samples/v1alpha1_valkeycluster.yaml or add a small helper to extract it).

bjosv (approved):
Good finding! We should restructure the tests in valkeycluster_test.go in the near future, to avoid these kind of problems.

PR #380, opened 2026-08-16
test(e2e): cluster re-forms after a full simultaneous pod restart

### LINKED ISSUE #352: e2e test: cluster re-forms after full simultaneous pod restart (stale-address heal)
Follow-up from #333 (requested by @jdheyburn in the #333 review).

#333 adds a reconcile phase that heals a persistent cluster whose members all restarted with new pod IPs (#275): stale peer entries are detected by node ID against the live scraped state and re-introduced with CLUSTER MEET.

This issue tracks an e2e test covering that path, to be added once #333 merges:

1. Create a ValkeyCluster (3 shards, 1 replica, persistence enabled, appendonly).
2. Wait Ready; write a marker key.
3. `kubectl delete pod --all` in the namespace — all pods restart simultaneously, every pod IP changes while each nodes.conf still holds the old peer IPs.
4. Assert: a `StaleAddressesHealed` event is emitted, the CR returns to Ready with `cluster_state:ok` within a bounded window (~1 min), and the marker key is still readable — without any manual CLUSTER MEET.

Optionally also assert the negative: no `NodeForgetFailed`/forget of live members during recovery (the #333 forget guard).

I'll pick this up after #333 lands.

matka12:
## Summary

The e2e test promised in #352, covering the stale-address heal merged in #333 (#275).

## Scenario

1. Create a 3-shard / 1-replica **persistent** ValkeyCluster (new sample manifest) and wait for Ready.
2. Write a canary key.
3. `kubectl delete pod -l valkey.io/cluster=...` — all six pods restart at once, every pod IP changes while each persisted `nodes.conf` still holds the old peer addresses. No node is isolated, so only the heal phase can re-form the cluster.
4. Assert:
   - a `StaleAddressesHealed` event is emitted for the cluster,
   - the CR returns to `Ready` with 3 ready shards, without any manual `CLUSTER MEET`,
   - `cluster_state:ok` and the canary key survived (AOF on the PVCs),
   - **no** `StaleNodeForgotten` event for the cluster during recovery — the forget guard from #333 (a forget here would 60s-ban the rejoining member).

## Verification

Ran the exact scenario against an image built from current main (`5676fb9`) on a local cluster: heal event fired (`Re-introduced 6 peer link(s) whose addresses changed`), Ready ~40s after the kill, canary intact, zero forget events.

Closes #352


matka12:
Addressed both review findings in 2d53da5 — same root issue from two angles: event assertions weren't isolated from retained events of an earlier same-named cluster.

- Both event queries now match `involvedObject.uid` of the CR created by this run, so events from a previous `cluster-persistent-restart` object can neither satisfy the heal assertion nor fail the forget assertion.
- Both assertions are additionally scoped to the restart window via pre-kill counts: `StaleAddressesHealed` must **increase** after the pod deletion (an event from initial formation can't pass it), and `StaleNodeForgotten` must stay **unchanged** (an unrelated earlier forget can't fail it).

config/samples/kustomization.yaml
```diff
@@ -2,4 +2,5 @@
 resources:
 - v1alpha1_valkeycluster.yaml
 - v1alpha1_valkeycluster-persistent.yaml
+- v1alpha1_valkeycluster-persistent-restart.yaml
```
bjosv: I'm not sure why we even kept the file kustomization.yaml in the repo.
We haven't added most sample manifests here anyway, so adding this is not needed in my view

config/samples/v1alpha1_valkeycluster-persistent-restart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: valkey.io/v1alpha1
```
bjosv: Could the manifest be inlined in the test rather than adding a new sample file? (See [example](https://github.com/valkey-io/valkey-operator/blob/7da59cbb5d8b4258876b03969825826a1913fec0/test/e2e/valkeycluster_test.go#L1405))

This manifest is somewhat test specific and not needed as a user-facing example.

matka12: Done in c32c8bd — manifest inlined in the test (same pattern as the persistence-mutation specs), sample file and kustomization entry dropped.

config/samples/kustomization.yaml
matka12: Reverted — kustomization.yaml untouched now.

test/e2e/valkeycluster_restart_test.go
```diff
@@ -0,0 +1,167 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			return len(utils.GetNonEmptyLines(output))
+		}
+		healEventsBefore := countEvents("StaleAddressesHealed")
+		forgottenEventsBefore := countEvents("StaleNodeForgotten")
+
+		By("deleting all pods at once so every pod IP changes")
+		cmd = exec.Command("kubectl", "delete", "pod",
+			"-l", podSelector, "--wait=false")
+		_, err = utils.Run(cmd)
```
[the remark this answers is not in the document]
matka12: Fixed in 5102171 — the test now snapshots the 6 pod UIDs before the kill and, right after the delete, waits until 6 pods exist that all carry NEW UIDs before any recovery assertion runs. A partial restart can no longer pass.

test/e2e/valkeycluster_restart_test.go
```diff
@@ -0,0 +1,167 @@
...
+  persistence:
+    size: 1Gi
+  config:
+    appendonly: "yes"
+`, clusterName)
+		cmd := exec.Command("kubectl", "delete", "valkeycluster", clusterName, "--ignore-not-found=true")
+		_, _ = utils.Run(cmd)
+		cmd = exec.Command("kubectl", "apply", "-f", "-")
+		cmd.Stdin = strings.NewReader(manifest)
+		_, err := utils.Run(cmd)
```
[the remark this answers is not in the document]
matka12: Fixed in 5102171 — setup now deletes any leftover CR and matching PVCs with --wait plus error checks before applying the manifest (retained PVCs would seed the run with old nodes.conf/canary data), and AfterAll waits and asserts on both deletions instead of discarding errors. Also applied the callback-scoped Gomega point from the outside-diff comment: countEvents/getPodUIDs now return errors and the Eventually callbacks assert via g.Expect, so transient kubectl failures retry instead of aborting the spec.

matka12:
@coderabbitai resolve

All three findings from the last round are addressed in 5102171:
- pod-UID snapshot + full-replacement gate before any recovery assertion (greptile P1)
- deterministic setup/teardown: leftover CR + PVCs deleted with --wait and error checks (CodeRabbit major)
- countEvents/getPodUIDs return errors and Eventually callbacks assert via the callback-scoped g.Expect (CodeRabbit outside-diff)

CI is green on that commit including the e2e suite.

jdheyburn (approved):
LGTM, thank you!

jdheyburn:
memo: Update branch and enable auto-merge

PR #381, opened 2026-08-17
refactor: Remove serverConfigHash from ValkeyNode

### LINKED ISSUE #264: [chore] Remove serverConfigRollHash from public ValkeyNode CRD spec
### Problem

`spec.serverConfigRollHash` on `ValkeyNode` is an internal implementation detail used to trigger pod rolls when the server config changes. It is set by the `ValkeyCluster` controller and was never intended to be part of the public CRD API.

Exposing it in the spec means it appears in `kubectl describe`, `helm` templates, and any tooling that introspects the CRD schema. This leaks implementation detail, confuses users, and creates an accidental API contract.

This was noted as a limitation in PR #209.

### Proposed fix

Move the config hash into an annotation on the ValkeyNode (or on the managed StatefulSet's pod template) rather than a spec field. The `ValkeyCluster` controller writes the annotation; the `ValkeyNode` controller reads it to determine whether a roll is needed.

This removes `serverConfigRollHash` from the CRD spec entirely. The change must be backwards-compatible (existing clusters must not trigger spurious rolls on upgrade).

### Acceptance criteria

- [ ] `serverConfigRollHash` removed from `ValkeyNode` spec
- [ ] Config hash stored as an annotation on the ValkeyNode or StatefulSet pod template
- [ ] Existing clusters do not trigger spurious pod rolls after upgrading the operator
- [ ] `ValkeyNode` CRD schema updated; migration note added to release notes

### References

- PR #209 (limitations section)

### Related GitHub issues (live configuration)

Roll triggers and “what must go through `CONFIG SET` vs a pod restart” are the same product surface. These issues are the open **live config** workstream:

- **#210** — Apply minimum set of configs live without rolls
- **#211** — Revert to default when managed live config is removed
- **#212** — Apply ACL changes live

**Sequencing note:** Moving `serverConfigRollHash` off the public spec (e.g. to an annotation) can stay a **small API cleanup** if roll semantics stay unchanged. If you instead want one coherent “config propagation” design (live apply + roll fallback + no leaked internals), treat this as part of the same design pass as #210–#212 and coordinate there.

jdheyburn:
This PR closes #264 

### Summary

This is a refactor to remove `serverConfigHash` from ValkeyNode.

Since #338 introduced a `WorkloadRevision` to ValkeyNode, there is duplicated functionality which means we no longer need a `serverConfigHash`. 

- ValkeyNode.spec.serverConfigHash is removed from the CRD (internal, operator-managed field; nothing outside the operator reads it).
- No runtime behaviour changes: config-change rolls, live-config apply (CONFIG SET), and ACL live-apply are unchanged. A frozen pin test (internal/controller/config_rollhash_test.go) guarantees the derived hash is byte-identical to the previously stamped one, so upgrading the operator rolls zero pods.

### Implementation

Simply removing a field that has had its function duplicated elsewhere.

- The ValkeyNode pod-template builders now derive the config-hash annotation themselves (nodeServerConfigRollHash), gated on `spec.serverConfigMapName`
- The render core in config.go is parent-agnostic so cluster and node sides produce the same output bytes
- The cluster controller no longer computes or threads a config hash, shrinking the parent→ValkeyNode contract: set the real inputs, call setDesiredWorkloadRevision.
- `spec.workloadRevision` remains the single roll-control field

### Limitations

During a mixed state, its possible that the operator would cause some rolls and failovers, however once CRDs and operator is synced up then this is not expected.

### Testing

Beyond unit tests, tested on a kind cluster to verify that:

- change in maxmemory-policy is applied live and does not cause a pod roll
- change in io-threads causes a pod roll

### Checklist


jdheyburn:
@greptile-apps @coderabbitai

bjosv (dismissed):
LGTM! The added testcase with the old render logic was smart.

sandeepkunusoth (approved):

PR #382, opened 2026-08-17
feat: gate ACLApplied on an ACL revision user so permission edits are honest

### LINKED ISSUE #369: ACLApplied does not reflect enabled/permission changes (applied live, not signalled)
Follow-up from #319.

`ACLApplied` compares the user set and each user's password hashes, so adding or removing a user or a password moves it through `PendingPropagation` back to `True`. A change to a user's `enabled` flag or permissions still takes effect on the server (the reload is unconditional), but the condition does not report a transient `PendingPropagation` for those fields, so it is not a signal to wait on for them. This is documented under `ACLApplied` in `docs/status-conditions.md`.

Why permissions are not compared directly: `ACL GETUSER` returns Valkey's normalized rendering of the rules, while the operator only holds the aclfile text, so a field-by-field comparison would mean reimplementing Valkey's ACL parser.

### Proposed fix

Append a revision sentinel to the aclfile: a disabled user (`_operator_acl_sentinel off`) whose only password hash is a hash of the whole managed ACL. It flows through the existing user and hash comparison, so `ACLApplied` reads `True` only once the running server has loaded the current revision, permissions included. I prototyped this and verified it live: a permission-only edit holds `ACLApplied=False` until the reload lands, then flips to `True`.

Longer term, a native `ACL DIGEST` command (proposed in valkey-io/valkey#4355) would remove the need for the sentinel user entirely.

melancholictheory:
This PR closes #369

### Summary

`ACLApplied` compares the user set and each user's password hashes, so a permission-only edit leaves it reading `True` for the whole window before the mounted aclfile refreshes, while the granted or revoked command is not yet live on the server.

The aclfile gains a trailing disabled user, `_operator_acl_revision`, whose only password hash is a hash of the managed ACL above it. It appears in `ACL LIST` and `ACL USERS`. It is `off` with `-@all` and no keys or channels, so it cannot authenticate and holds no capability.

`ACLApplied` now reads `True` only once the running server has loaded the current revision, permission edits included.

### Implementation

Comparing permissions directly is not practical: `ACL GETUSER` returns Valkey's normalized rendering of the rules while the operator only holds the aclfile text, so a field-by-field comparison would mean reimplementing Valkey's ACL parser and keeping it in step with the server.

The revision user sidesteps that. Its password hash covers every managed entry above it, so any edit changes it, and it then flows through the user and password-hash comparison that `aclObservablyInSync` already does. It needs no new comparison logic, and the condition stops depending on which fields happen to be comparable.

`reconcileUsersAcl` is the only place the aclfile is assembled, and the revision user has to be appended in every version written there. A path that wrote the file without it would leave a stale revision on the server and hold every node at `ACLApplied=False` forever, so there is a comment on that function saying to keep aclfile assembly in one place.

### Limitations

The revision user is a workaround for the server having no way to report which ACL revision it holds. valkey-io/valkey#4355 proposes `ACL DIGEST` for exactly this, and the TSC vote on it is currently 3 core-team approvals with no objections. Once that command exists, the operator can compare the server's digest directly and the bookkeeping user can be dropped, behind a version gate of the kind #307 is adding.

### Testing

Unit tests cover the revision user being appended last and its hash covering the managed entries above it, plus the sync comparison.

Verified on a local k3d cluster, both directions, using a permission-only edit that leaves users and passwords untouched.

On this branch, `ACLApplied` held `False` with reason `PendingPropagation` for about 40 seconds, the mounted-secret refresh window, then flipped to `True` once the reload landed. The revision hash changed with the edit. `ACL GETUSER _operator_acl_revision` on the server reported `off`, `-@all`, no keys and no channels.

On a build of `main` with the same cluster and the same class of edit, starting from a stable `True`, the condition stayed `True` for the full 60 seconds and never signalled, which is the behaviour #369 describes.

### Checklist

### DIFF
--- a/docs/status-conditions.md
+++ b/docs/status-conditions.md
@@ -206,19 +206,17 @@ This condition is set once the node is ready and mounts an ACL Secret. An edit r
 
 | Status | Meaning |
 |---|---|
-| `True` | The desired user set and their password hashes are live on the server. |
-| `False` | The reload ran but the running ACL does not yet match the desired users and passwords (for example the mounted aclfile has not refreshed yet), or the reload failed. |
+| `True` | The desired user set, their password hashes, and the current ACL revision are live on the server. |
+| `False` | The reload ran but the running ACL does not yet match the desired revision (for example the mounted aclfile has not refreshed yet), or the reload failed. |
 
 Common reasons when `ACLApplied=False`:
 - `PendingPropagation` – the reload read a stale mounted aclfile; the node retries until the projected volume catches up.
 - `ApplyFailed` – `ACL LOAD` or the follow-up read returned an error. The message field contains the exact error.
 
 Common reasons when `ACLApplied=True`:
-- `Applied` – the desired users and passwords are live.
+- `Applied` – the desired ACL revision is live.
 
-> **Note:** `ACLApplied` compares the user set and password hashes exactly, which is what a password rotation waits on. It is informational and does not block the cluster controller's one-at-a-time progress. On a failed apply the node controller emits a `LiveACLApplyFailed` warning event and retries with backoff.
->
-> **Limitation:** Only the user set and password hashes drive the condition, so only those changes move it through `PendingPropagation` back to `True`: adding or removing a user or a password. A change to a user's `enabled` flag or permissions still takes effect on the server (the reload is unconditional), but the condition does not report a transient `PendingPropagation` for it, so it is not a signal to wait on for those fields. Comparing rules directly would mean reimplementing Valkey's normalized ACL rendering; making the condition track every field is left as a follow-up.
+> **Note:** The operator appends a disabled bookkeeping user, `_operator_acl_revision`, to the aclfile. Its only password is a hash of the whole managed ACL, so a node reports `ACLApplied=True` only once the running server has loaded that exact revision. This keeps the condition honest for permission-only edits, which leave every user and password unchanged but still change the revision hash. The user is disabled (`off`) and cannot authenticate; it is expected to appear in `ACL LIST`. The condition is informational and does not block the cluster controller's one-at-a-time progress. On a failed apply the node controller emits a `LiveACLApplyFailed` warning event and retries with backoff.
 
 #### `WorkloadRollPending`
 Indicates that a rolling pod-template update is intentionally deferred: the ValkeyNode controller has built a pod template that differs from the live StatefulSet or Deployment, but `spec.workloadRevision` has not yet authorized that template. This is expected staging while the cluster advances rolls one node at a time, not an error.
--- a/internal/controller/users.go
+++ b/internal/controller/users.go
@@ -40,6 +40,16 @@ const (
 	hashAnnotationKey = "valkey.io/internal-acl-hash"
 	aclFilename       = "users.acl"
 	passwordLength    = 26
+	// aclRevisionUser is a disabled bookkeeping user appended to the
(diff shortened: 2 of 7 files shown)


jdheyburn:
Overall LGTM, could we use an alternative to `sentinel` so that it doesn't conflict with Valkey Sentinel? Perhaps `_operator_acl_revision` instead?

internal/controller/users.go
```diff
@@ -40,6 +40,16 @@ const (
...
+	// aclRevisionSentinelUser is a disabled bookkeeping user appended to the
+	// aclfile whose only password hash is the hash of the managed ACL content
+	// above it. The ValkeyNode controller waits for this user to be present with
+	// the current hash before reporting ACLApplied, which is how it knows the
+	// running server loaded the current aclfile revision rather than a stale
+	// mounted copy. Because the hash covers the whole managed ACL, it also
+	// catches permission-only edits that leave user and password identities
+	// unchanged. It is disabled (off), so its content-hash password is never a
+	// usable credential.
+	aclRevisionSentinelUser = "_operator_acl_sentinel"
```
bjosv: Using sentinel will probably confuse people, both here and when reading the added code  (see https://valkey.io/topics/sentinel/).

An alternative name for the user could be `aclRevisionUser = "_operator_acl_revision"`, but there might be even better names.

bjosv: oh, I now see that this was already mentioned above..

melancholictheory:
Renamed to `_operator_acl_revision`, with `aclRevisionUser` for the constant, which is the name you both landed on. The prose in the code comments and the docs moved to "revision user" too, and I rebased onto main while I was here.

While renaming I left the real Valkey Sentinel references alone: the mentions in `docs/quickstart.md` and `docs/valkeynode-design.md`, the `role:sentinel` INFO fixture in `valkeynode_resources_test.go`, and the `#acl-rules-for-sentinel-and-replicas` link in `users.go` are all about Sentinel proper, which is the collision you were pointing at.

Re-ran the k3d check after the rename. The user lands in `ACL USERS` as `_operator_acl_revision`, and a permission-only edit still holds `ACLApplied` at `PendingPropagation` until the reload, then flips back to `True` with a new hash.

On the bot's suggestion to write "ACL file" instead of "aclfile" in the docs: the repo says `aclfile` in prose everywhere it comes up and never "ACL file", so changing only the lines this PR touches would leave `status-conditions.md` inconsistent with itself. Happy to switch all of them in one go if you prefer that wording.

docs/status-conditions.md
```diff
@@ -206,19 +206,17 @@ This condition is set once the node is ready and mounts an ACL Secret. An edit r
 
 | Status | Meaning |
 |---|---|
-| `True` | The desired user set and their password hashes are live on the server. |
-| `False` | The reload ran but the running ACL does not yet match the desired users and passwords (for example the mounted aclfile has not refreshed yet), or the reload failed. |
+| `True` | The desired user set, their password hashes, and the current ACL revision are live on the server. |
+| `False` | The reload ran but the running ACL does not yet match the desired revision (for example the mounted aclfile has not refreshed yet), or the reload failed. |
```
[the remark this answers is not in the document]
jdheyburn: Unnecessary

jdheyburn (approved):
I tested this locally and it works great - thanks for following up on this to finesse it!

bjosv (approved):
Great!

PR #383, opened 2026-08-17
test(e2e): Hostname migrate keeps Ready pod UID

### LINKED ISSUE #365: networking: discovery preferredEndpointType and clusterDomain

daanvinken:
e2e for #378. Does not close #365.

### Summary

Kind coverage for Hostname discovery migrate:

1. Create a 1-shard IP-announce cluster and wait until Ready.
2. Pause the operator, restore a legacy per-node `serviceName` with orphan recreate, unpause.
3. The original Ready pod UID stays. Announce stays IP.
4. Patch `preferredEndpointType: Hostname`. Hostname flags and a trailing-dot FQDN are allowed only on a later pod (new UID).
5. Assert `cluster-preferred-endpoint-type hostname` and `CLUSTER SLOTS` hostname.

Operator code is already on main via #378. This PR is the e2e only.

### Limitations

- Not the #365 multi-shard TLS Hostname client e2e (cert SANs, re-dial without skip verify).
- A 1-shard cluster writes Hostname and WorkloadRevision in one ValkeyNode update. This spec does not fake a deferred revision window.

### Testing

- `go test -tags=e2e -c ./test/e2e/`
- `go test -tags=e2e ./test/e2e/ -ginkgo.label-filter=Discovery` on Kind

### Checklist

Signed-off-by: daanvinken <daanvinken@tythus.com>


daanvinken:
These are the E2E tests that build on top of https://github.com/valkey-io/valkey-operator/pull/378

I plan to merge https://github.com/valkey-io/valkey-operator/pull/378 first, and then rebase this PR. Simply validating the E2E tests for now.

melancholictheory:
The UID and ownerReference assertions are exactly what makes this prove the property rather than look like it, thanks for picking them up.

One small thing on the hostname block. `serverCommand` reads the live pod, and a pod's container command is immutable once created, so the original pod cannot gain `--cluster-announce-hostname` without being replaced. That makes the `Fail("original pod announced hostname")` branch unreachable. The property it is aiming at is already carried by the `uid` assertion right below it, so the guard mostly reads as protecting something that cannot happen.

melancholictheory:
Which of the two should reviewers be reading? #378 and this one carry the same operator change, and your note about not being able to stack from a fork explains why, but nothing on either PR says which is the one to review. Right now someone landing on either can reasonably start reviewing the operator code, and the same comments end up split across both.

If the plan is still #378 first and a rebase here afterwards, saying so at the top of both descriptions would keep the review on one of them.

test/e2e/valkeycluster_discovery_hostname_test.go
```diff
@@ -0,0 +1,285 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	valkeyiov1alpha1 "github.com/valkey-io/valkey-operator/api/v1alpha1"
+	"github.com/valkey-io/valkey-operator/test/utils"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var _ = Describe("ValkeyCluster Hostname discovery", Ordered, Label("ValkeyCluster", "Discovery"), func() {
```
bjosv, line 37:
```suggestion
var _ = Describe("ValkeyCluster Hostname discovery", Ordered, Serial, Label("ValkeyCluster", "Discovery"), func() {
```

bjosv: nit: Since we modify the operator in this test we should probably make sure no other test is allowed to run at the same time. If parallel e2e testing is ever introduced, this would be the first thing to break (..but sure there are a lot more blocking it.).

test/e2e/valkeycluster_discovery_hostname_test.go
```diff
@@ -0,0 +1,285 @@
...
+			g.Expect(cr.Status.State).To(Equal(valkeyiov1alpha1.ClusterStateReady))
+		}, 10*time.Minute, 5*time.Second).Should(Succeed())
+
+		podUID := podUIDOf(Default, podName)
+		expectIPAnnounce(Default, stsName, podName)
+		Expect(stsServiceName(Default, stsName)).To(Equal(headless))
+		Expect(podSubdomain(Default, podName)).To(Equal(headless))
+
+		By("injecting a legacy per-node serviceName without replacing the pod")
+		withOperatorPaused(func() {
```
bjosv, line 88: nit: maybe its `withOperatorStopped`?  It was not `docker pause` as I first believed

bjosv (commented):
e2e LGTM

daanvinken:
Moved to draft, will submit E2E for review once https://github.com/valkey-io/valkey-operator/pull/378 is merged.

daanvinken:
cc @melancholictheory @bjosv for review 🙏

bjosv (dismissed):
LGTM

test/e2e/valkeycluster_discovery_hostname_test.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/valkey-io/valkey-operator/test/utils"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+var _ = Describe("ValkeyCluster Hostname discovery", Ordered, Label("ValkeyCluster", "Discovery"), func() {
+	const clusterName = "cluster-hostname-e2e"
+	const stsName = "valkey-cluster-hostname-e2e-0-0"
+	const podName = "valkey-cluster-hostname-e2e-0-0-0"
+	const headless = "valkey-cluster-hostname-e2e"
+	const announceHostnameFQDN = "$(POD_NAME).valkey-cluster-hostname-e2e.default.svc.cluster.local."
```
bjosv: Ah, we need to update this after merging #429 
```suggestion
	const announceHostnameFQDN = "$(POD_NAME).valkey-cluster-hostname-e2e.default.svc.cluster.local"
```

daanvinken: Done!

bjosv (approved):

PR #384, opened 2026-08-18
ci: remove unnecessary disk space reclaim from publish workflow

bjosv:
This PR is part of #371

### Summary

Remove the "Reclaim disk space" step from the publish workflow. It is no longer needed since the move from QEMU emulation to native per-arch runners.

Cuts the amd64 build time in half for most publish runs, i.e. removes up to 4 minutes in some runs (!!?).

### Testing

Verified on a fork with disk space logging. The runner starts with 105GB free and the build uses ~4GB, well within capacity without any reclaim.

### Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -42,12 +42,6 @@ jobs:
           # Registry refs must be lowercase; $GITHUB_REPOSITORY may not be.
           echo "name=ghcr.io/${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
 
-      - name: Reclaim disk space
-        run: |
-          docker image prune --force --all
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /usr/local/lib/android
-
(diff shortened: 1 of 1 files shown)


jdheyburn (approved):
Would this put us at risk of jobs failing because of disk full? I'm fine with removing it for now then tackling it when we come across it.

bjosv:
> Would this put us at risk of jobs failing because of disk full? I'm fine with removing it for now then tackling it when we come across it.

When reading up on it; every CI job starts on a new/clean virtual machine, and we have now split the publish-build-job onto 2 VMs. There is a minimum of [14GB](https://docs.github.com/en/actions/reference/runners/github-hosted-runners) available on each VM, and each imagebuild uses ~4GB today, so it should be safe for a long time.

The cleanup gives additional 1.8GB if we ever need that extra storage, but potentially we could get [more](https://www.geraldonit.com/mastering-disk-space-on-github-actions-runners-a-deep-dive-into-cleanup-strategies-for-x64-and-arm64-runners/).

PR #390, opened 2026-08-21
feat: add Valkey CRD for standalone instances

### LINKED ISSUE #388: [FEATURE]: add Valkey CRD for standalone instances
### Pre-flight checks

### Problem / Use case

Add Valkey CRD for standalone instances. This is fundamental block for upcoming ValkeySentinel and ValkeyCell CRDs.

### Proposed solution

Add type Valkey CRD.

### Scope of change

Large (new CRD, major behaviour change, requires design doc)

### API / behaviour impact

### Contribution

### Additional context

_No response_

deepakpunjabi:
This PR closes https://github.com/valkey-io/valkey-operator/issues/388.

### Summary

Adds the `Valkey` CRD, the first top-level API for running Valkey outside cluster mode. This PR is API types and validation only. The controller lands in a follow-up, so creating a `Valkey` today has no runtime effect. Not adding user facing docs for the same reason.

Design context is in discussion #387. This is the first slice of that plan, scoped deliberately so it avoids every decision still open in the discussion (selector versus ref linkage, per-node config, quorum-aware PDBs, monitor registration ownership).

- `spec.replicas` matches `ValkeyCluster.spec.replicas` semantics, so 0 means a lone primary and N means one primary plus N replicas. Only 0 is admitted for now.
- `spec.config` rejects `cluster-` keys, case-insensitively.
- Status reports `state`, `reason`, `message`, `primary`, `replicas`, `readyReplicas`, `observedGeneration` and conditions.
- `kubectl get valkey` prints State, Reason, Primary and Age.
- No existing behaviour changes. `ValkeyCluster` and `ValkeyNode` are untouched.

### Implementation

New files:

- `api/v1alpha1/valkey_types.go`: the `Valkey`, `ValkeySpec`, `ValkeyStatus` and `ValkeyState` types.
- `internal/controller/valkey_validation_test.go`: 20 envtest specs covering the CEL rules.
- `config/samples/v1alpha1_valkey.yaml`: minimal sample.

Generated or wired:

- `config/crd/bases/valkey.io_valkeys.yaml` and the `zz_generated.deepcopy.go` additions, both from `make manifests generate`.
- `config/crd/kustomization.yaml` gains the new CRD.

Three things worth reviewer attention:

1. **Check for `replicas` using `has()`**
Because `spec.replicas` is optional, leaving it out causes a "key not found" error if you check `self.replicas` directly. Using `has()` prevents this error so valid standalone instances aren't rejected.

**Result:**
* Missing field = allowed 
* `replicas: 0` = allowed
* `replicas: 2` = blocked

2. **Limit name length and block reserved suffixes**
Names in `metadata.name` are capped at 47 characters and cannot end in `-primary`, `-replicas`, or numbers (e.g., `-1`).

* **Length:** Child resources add text (like `valkey-<name>-replicas`), which must fit inside Kubernetes' 63-character limit.
* **Suffixes:** Prevents two instances from accidentally creating child resources with the same name.
* **Timing:** Setting these rules now prevents breaking existing deployments if validation is added later.

3. **`spec.podDisruptionBudget` is out of scope for now**
Pod Disruption Budgets (PDBs) don't make sense for a single pod. A budget either lets the pod get deleted immediately or blocks all maintenance entirely. Additionally, the only available mode (`Cluster`) doesn't apply to single pods. This field will be added later when multi-pod replication is supported.

### Limitations

- **No controller.** A `Valkey` object is accepted and validated but nothing reconciles it, so no pods are created. That is the next PR.
- **Standalone only.** `spec.replicas > 0` is rejected by a CEL rule until replication support lands. Relaxing that bound later is backwards compatible, whereas adding it later would not be.
- **No Sentinel.** No `spec.sentinel` block, since the linkage direction is still under discussion in #387.

### Testing

- `internal/controller/valkey_validation_test.go` adds 9 envtest.
- `make test` passes with no regressions.
- `make lint` passes.
- `pre-commit run --all-files` passes.

### Checklist


api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Containers holds additional containers, or overrides for the default
+	// ones, applied as a strategic merge patch.
+	// +optional
+	Containers []corev1.Container `json:"containers,omitempty"`
+
+	// Config holds additional Valkey configuration parameters.
+	// Cluster mode directives are not accepted here by design.
+	// +optional
+	Config map[string]string `json:"config,omitempty"`
```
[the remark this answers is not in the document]
deepakpunjabi: Addressed.

config/crd/bases/valkey.io_valkeys.yaml
```diff
@@ -0,0 +1,3945 @@
...
+                            required:
+                            - secretName
+                            type: object
+                        required:
+                        - server
+                        type: object
+                    required:
+                    - certificates
+                    type: object
+                type: object
```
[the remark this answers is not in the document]
deepakpunjabi: This was due to rebase adding changes to the CRD.

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+	// +optional
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// Resources overrides the resource requirements for the Valkey container.
+	// +optional
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Scheduling groups pod placement configuration for the instance's pods.
+	// +optional
+	Scheduling *SchedulingSpec `json:"scheduling,omitempty"`
```
jdheyburn, line 211: Can we copy SchedulingSpec to ValkeySchedulingSpec and use that instead please? We can just support these fields for now.

```go
// ValkeySchedulingSpec groups pod placement for a Valkey's pods.
type ValkeySchedulingSpec struct {
    Tolerations               []corev1.Toleration               `json:"tolerations,omitempty"`
    NodeSelector              map[string]string                 `json:"nodeSelector,omitempty"`
    Affinity                  *corev1.Affinity                  `json:"affinity,omitempty"`
    TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
    PriorityClassName         string                            `json:"priorityClassName,omitempty"`
}
```

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+	// The base config still emits cluster-config-file for every node.
+	// That keeps the node state file on the writable /data volume.
+	// See buildManagedConfig.
+	// +kubebuilder:validation:XValidation:rule="self.all(key, !key.lowerAscii().startsWith('cluster-'))",message="spec.config must not contain cluster- keys: a Valkey runs standalone, so cluster mode directives are not supported"
+	// +optional
+	Config map[string]string `json:"config,omitempty"`
+
+	// Networking groups how clients and peers reach the instance.
+	// +optional
+	Networking *NetworkingSpec `json:"networking,omitempty"`
```
jdheyburn, line 261: Similar to SchedulingSpec, we will want to copy this into ValkeyNetworkingSpec. Along with TLSSpec -> ValkeyTLSSpec. Since TLS will include cluster-specific detail in it soon.

```go
// ValkeyNetworkingSpec groups how clients and peers reach a Valkey.
type ValkeyNetworkingSpec struct {
    // Same markers as NetworkingSpec.ClusterDomain (default, length, pattern).
    ClusterDomain string          `json:"clusterDomain,omitempty"`
    TLS           *ValkeyTLSSpec  `json:"tls,omitempty"`
}
```

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+	// The comparison is lowercased because Valkey keys are case-insensitive.
+	// Cluster-Enabled has to be caught alongside cluster-enabled.
+	//
+	// This rejects user input only.
+	// The base config still emits cluster-config-file for every node.
+	// That keeps the node state file on the writable /data volume.
+	// See buildManagedConfig.
+	// +kubebuilder:validation:XValidation:rule="self.all(key, !key.lowerAscii().startsWith('cluster-'))",message="spec.config must not contain cluster- keys: a Valkey runs standalone, so cluster mode directives are not supported"
+	// +optional
+	Config map[string]string `json:"config,omitempty"`
```
jdheyburn, line 257: Can we add a CEL rule to reject `replicaof` and `slaveof` please?

```
// +kubebuilder:validation:XValidation:rule="self.all(key, !(key.lowerAscii() in ['replicaof', 'slaveof']))",message="spec.config must not contain replicaof or slaveof: the operator owns replication topology"
```

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+	// +optional
+	Quorum *int32 `json:"quorum,omitempty"`
+
+	// Config is per-master Sentinel tuning.
+	// Forwarded as SENTINEL SET <monitorName> <key> <value>.
+	// Values are not validated by the operator.
+	// Operator-owned keys are skipped with a ConfigurationWarning.
+	// +optional
+	Config map[string]string `json:"config,omitempty"`
+}
```
jdheyburn, line 142: Let's remove these for now and add them back after we've added ValkeySentinel CRD.

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+	// +kubebuilder:default=None
+	// +optional
+	Mode FailoverMode `json:"mode,omitempty"`
+
+	// Sentinel configures Sentinel-mode failover.
+	// Only valid when mode is Sentinel.
+	// This block does not by itself cause monitoring.
+	// A ValkeySentinel must also select this object.
+	// +optional
+	Sentinel *SentinelFailoverSpec `json:"sentinel,omitempty"`
```
jdheyburn, line 102: Same here, let's remove this for now.

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+// Sentinel is declared so the vocabulary is fixed.
+// A spec-level rule rejects it until the controllers that honour it exist.
+// Admitting a mode nothing implements would store an unactionable spec.
+// +kubebuilder:validation:Enum=None;Sentinel
+type FailoverMode string
+
+const (
+	// FailoverModeNone means nothing promotes automatically.
+	// A standalone instance, or replication with manual failover.
+	FailoverModeNone FailoverMode = "None"
```
jdheyburn, line 37: Can we use `Disabled` instead to be consistent with existing Modes we have elsewhere

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+// +kubebuilder:subresource:status
+
+// Valkey is the Schema for the valkeys API.
+//
+// The name is bounded because child resource names are derived from it.
+// Every one of them is a DNS label capped at 63 characters.
+//
+// The binding child is the Secret "internal-<name>-system-passwords".
+// Its 26 fixed characters leave 37 for the name.
+// No other derived name is tighter, so this is the only limit stated.
```
jdheyburn, line 333: The limit is really derived from the ValkeyNode StatefulSet (`valkey-<name>-<idx>`) because its `controller-revision-hash` label carries `<sts-name>-hash`.

```suggestion
// The name is bounded because child resource names are derived from it.
// Every derived name is held to a 63 character DNS label.
// That is stricter than some child types need: Secret names allow 253.
// One rule for every child keeps the limit simple.
// It also leaves room for derived names used as labels or volume names.
//
// The longest derived name is the Secret "internal-<name>-system-passwords".
// Its 26 fixed characters leave 37 for the name.
```

internal/controller/valkey_validation_test.go
```diff
@@ -0,0 +1,369 @@
...
+		},
+		// These were reserved while the design still had role Services.
+		// Those Services were dropped, so the names are legal again.
+		Entry("former primary Service suffix", "val-primary"),
+		Entry("former replicas Service suffix", "val-replicas"),
+	)
+
+	It("rejects a name too long for derived child names", func() {
+		// 38 characters, one over the limit.
+		// The system password Secret is the longest derived name.
```
jdheyburn, line 157: Same point mentioned previously.

```suggestion
		// See the Valkey type comment for how the limit is derived.
```

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+// Tightening it later would reject objects that already exist.
+// +kubebuilder:validation:XValidation:rule="self.metadata.name.size() <= 37",message="metadata.name must be at most 37 characters, because child resource names are derived from it"
+//
+// A trailing "-<number>" is reserved.
+// ValkeyNode names are derived as "<name>-<index>".
+// An instance "cache-1" would otherwise collide with node 1 of "cache".
+// +kubebuilder:validation:XValidation:rule="!self.metadata.name.matches('-[0-9]+$')",message="metadata.name must not end with '-<number>': that suffix is reserved for derived ValkeyNode names"
+// +kubebuilder:printcolumn:name="State",type="string",JSONPath=".status.state",description="Current state of the instance"
+// +kubebuilder:printcolumn:name="Reason",type="string",JSONPath=".status.reason",description="Reason for current state"
+// +kubebuilder:printcolumn:name="Primary",type="string",JSONPath=".status.primary",description="ValkeyNode currently serving as primary",priority=1
```
jdheyburn, line 345: Can we remove `,priority=1` to make it appear by default?

api/v1alpha1/valkey_types.go
```diff
@@ -0,0 +1,398 @@
...
+// No other derived name is tighter, so this is the only limit stated.
+//
+// The limit is in place from the start.
+// Tightening it later would reject objects that already exist.
+// +kubebuilder:validation:XValidation:rule="self.metadata.name.size() <= 37",message="metadata.name must be at most 37 characters, because child resource names are derived from it"
+//
+// A trailing "-<number>" is reserved.
+// ValkeyNode names are derived as "<name>-<index>".
+// An instance "cache-1" would otherwise collide with node 1 of "cache".
+// +kubebuilder:validation:XValidation:rule="!self.metadata.name.matches('-[0-9]+$')",message="metadata.name must not end with '-<number>': that suffix is reserved for derived ValkeyNode names"
```
jdheyburn, line 342: This conflicts with the plan for ValkeyCell, which manages sharded Valkey. For a ValkeyCell with shards: 2, it would create these Valkey:

```
cache-0
cache-1
```

I appreciate this is to avoid conflicts, so could we drop this rule and instead implement conflict detection on the controller side? 

The only other alternative I can think of is changing ValkeyNode to include `node` in the name:

```
cache-0-node-0
cache-0-node-1
cache-1-node-0
cache-1-node-1
```

jdheyburn (commented):
Thanks for getting started on this!

Issue #396, opened 2026-08-21
[BUG]: ValkeyCluster reports Ready/ClusterHealthy while a node is stuck on ACLApplied=ApplyFailed


bcfmtolgahan:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.
- [x] I am running a recent version of the operator (latest release or `main`) and the issue still reproduces.

### Bug Description

When a ValkeyNode is stuck on `ACLApplied=False/ApplyFailed`, the owning ValkeyCluster keeps reporting `Ready/ClusterHealthy` with every condition True. The declared users are not present in Valkey, but nothing at cluster level says so. You only find it by describing the ValkeyNode.

#257 and #140 surfaced node reconcile errors and `reconcileUsersAcl` errors in status. A persistent `ApplyFailed` on the live-apply path does not appear to be covered.

### Steps to Reproduce

1. Create a cluster with valid users and wait for Ready.
2. Add a user whose `commands.allow` entry Valkey rejects (see #395) together with a valid new user.
3. Watch the ValkeyNode condition and the ValkeyCluster status.

### Expected Behaviour

An ACL that the operator cannot apply, and keeps failing to apply, is reflected in `ValkeyCluster.status` in some form. Either a non-Ready state or a dedicated condition. Users watching `kubectl get valkeycluster` should not see a healthy cluster while their ACL is not in effect.

### Actual Behaviour

Sampled every 30s for 20 minutes, 40 samples, one single state throughout:

```
aclgood ACLApplied=False/ApplyFailed cluster=Ready/ClusterHealthy carol_exists=0
```

Full cluster status while the ACL is not applied:

```
Ready=True/ClusterHealthy  Progressing=False/ReconcileComplete
ClusterFormed=True/TopologyComplete  SlotsAssigned=True/AllSlotsAssigned
```

The node keeps retrying and the error is in the operator log and in metrics, but not in cluster status:

```
ERROR failed to apply live ACL {"ValkeyNode": {"name":"aclgood-0-0"}, "error": "ACL LOAD: ..."}

controller_runtime_reconcile_errors_total{controller="valkeynode"}   18
controller_runtime_reconcile_errors_total{controller="valkeycluster"} 0
```

### Operator version

main @ c33988f

### Kubernetes version

v1.35.3-eks

### Kubernetes distribution / environment

Amazon EKS

### Additional context

[...]

Issue #395, opened 2026-08-21
[BUG]: commands.allow/deny pattern marker is malformed, so no validation is generated


bcfmtolgahan:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.
- [x] I am running a recent version of the operator (latest release or `main`) and the issue still reproduces.

### Bug Description

The pattern markers on `commands.allow` and `commands.deny` are malformed:

```go
// valkeyacls_types.go:79 and :83
// +kubebuilder:validation:Items:Pattern=^[@a-z|]+$}
```

`Items` should be lowercase and there is a stray `}`. controller-gen drops the marker without complaining, so the generated schema has no pattern at all:

```
$ yq '...commands.properties.allow' config/crd/bases/valkey.io_valkeyclusters.yaml
{description: Allowed commands for this user, items: {type: string}, type: array}
```

Any string is admitted and lands in the ACL file. Valkey then rejects it, and what happens next depends on when the bad value arrives.

### Steps to Reproduce

```yaml
users:
  - name: mallory
    nopass: true
    commands:
      allow: ["THIS IS NOT A COMMAND !!!"]
```

Admission accepts this. Then:

1. If it is present at cluster creation, Valkey refuses to start and the pod crash-loops.
2. If it is added to a running cluster, the ACL edit is rejected atomically and every new or changed user in that edit is dropped, including valid ones.

### Expected Behaviour

Admission rejects a command entry that cannot be a Valkey command or category.

### Actual Behaviour

Case 1, invalid at creation:

```
# Aborting Valkey startup because of ACL errors: /config/users/users.acl:5:
  Error in applying operation '+THIS': Unknown command or category name in ACL.

valkey-aclbad-0-0-0   1/2   CrashLoopBackOff   9 restarts
```

Case 2, added later. `carol` is valid and was added in the same edit as `mallory`:

```
mounted /config/users/users.acl:
  3:user carol on nopass +@read ~*
  4:user mallory on nopass +THIS IS NOT A COMMAND !!!

$ valkey-cli ACL LIST | awk '{print $2}'
_exporter _operator _operator_acl_revision _replication alice bob default

carol     -> AUTH failed: WRONGPASS   (same as a user that was never declared)
alice/bob -> still work               (previously active rules kept)

ValkeyNode: ACLApplied=False/ApplyFailed
```

### Operator version

main @ c33988f

### Kubernetes version

v1.35.3-eks

### Kubernetes distribution / environment

[...]

PR #397, opened 2026-08-21
fix: default exporter.enabled to true so partial overrides keep the sidecar

### LINKED ISSUE #394: [BUG]: exporter sidecar dropped when any exporter field other than enabled is set
### Pre-flight checks

### Bug Description

`spec.exporter` has an object-level default of `{enabled: true}` (`valkeycluster_types.go:339`), but `enabled` has no default of its own (`valkeycluster_types.go:451`). An object default only applies when the whole object is absent, so setting any sibling field makes the object present and `enabled` falls back to the Go zero value `false`. The sidecar is then never rendered.

Overriding the image or resources is the usual reason to touch `spec.exporter` at all, so the field is easy to break by using it. Nothing signals the loss: no event, no condition, no operator log line.

#345 added `args`, which is one of the affected fields.

### Steps to Reproduce

Apply clusters differing only in `spec.exporter`, then list the containers in each pod:

```
spec.exporter as stored                     containers
------------------------------------------  -----------------------
{"image":"oliver006/redis_exporter:v1.80.0"} server
{"resources":{"requests":{...}}}             server
{"args":["--log-format=json"]}               server
{"securityContext":{"runAsNonRoot":true}}    server
{"enabled":true,"image":"...:v1.80.0"}       server metrics-exporter
{"enabled":true}   (field omitted entirely)  server metrics-exporter
```

### Expected Behaviour

The exporter stays enabled unless `enabled: false` is set explicitly. Setting `image`, `resources`, `args` or `securityContext` overrides that setting and leaves the sidecar running.

### Actual Behaviour

The sidecar is dropped and metrics stop. The cluster still reports healthy:

```
$ kubectl get events -n valkey-e2e --field-selector type=Warning
No resources found in valkey-e2e namespace.

conditions: Ready=True/ClusterHealthy Progressing=False/ReconcileComplete
            ClusterFormed=True/TopologyComplete SlotsAssigned=True/AllSlotsAssigned
```

```yaml
# no exporter sidecar
kind: ValkeyCluster
metadata: {name: e-image}
spec:
  shards: 1
  replicas: 0
  exporter:
---
# exporter sidecar present (control)
kind: ValkeyCluster
metadata: {name: e-omitted}
spec:
  shards: 1
  replicas: 0
```

### Operator version

main @ c33988f

### Kubernetes version

v1.35.3-eks

Amazon EKS

### Additional context

Fix is either `+kubebuilder:default=true` on `ExporterSpec.Enabled`, or making it `*bool` and treating `nil` as enabled. Happy to send a PR.

bcfmtolgahan:
This PR closes #394

### Summary

`spec.exporter` only defaulted `enabled` when the whole object was absent, so setting `image`, `resources`, `args` or `securityContext` silently dropped the metrics sidecar. The cluster controller now resolves the default itself.

- Overriding any exporter field keeps the sidecar enabled, as does an explicitly empty `exporter: {}`. Only an explicit `enabled: false` disables it.
- No change for clusters that omit `spec.exporter` or set `enabled` explicitly.
- No change for standalone `ValkeyNode`s: the sidecar still requires an explicit `enabled: true` there.

### Implementation

`ExporterSpec.Enabled` becomes `*bool`. `ValkeyClusterSpec.ExporterEnabled()` resolves a nil to enabled — the method lives on the cluster spec, not on the shared `ExporterSpec`, so it cannot be misused on a `ValkeyNode`, where nil means disabled. `buildClusterValkeyNode` propagates an explicit `true` only when enabled; disabled stays nil on the node spec, because old operators omitted `false` and an explicit `false` makes `nodeRequiresRoll` fail over upgraded clusters over a no-op spec diff (#401). A `ValkeyNode` runs the sidecar only on an explicit `true`. The three `users.go` call sites resolve through `ExporterEnabled()`.

The first cut instead put `+kubebuilder:default=true` on the field. That stamped `enabled: true` onto bare `ValkeyNode`s too, giving standalone nodes a sidecar wired to `_exporter` credentials that only the cluster controller provisions, so their pods never became ready. The `valkeynode` e2e suite caught it once the workflows ran. Resolving in the controller keeps the default a cluster-level concern.

A plain `bool` does not work for the field. Dropping `omitempty` means the typed Go client always serialises `false`; keeping it means an explicit `false` is omitted on the next write and read back as nil, i.e. enabled. The pointer avoids both.

Two existing fixtures (`newTestValkeyNode`, and the Deployment-workload node in `valkeynode_controller_test.go`) now set `Enabled: boolPtr(false)` explicitly. They were relying on the zero value meaning disabled, and their assertions are about the server container, not the exporter.

### Documentation

No change needed. `docs/valkeycluster.md` already documents `enabled: true` as the default and states that "each pod runs a `metrics-exporter` sidecar by default". The docs were right; the code was the thing out of line, and the same doc block shows an `exporter` example setting `image`, `args` and `resources`, which is precisely the shape that silently disabled the sidecar.

### Limitations

Only covers the defaulting. #396 tracks the separate question of whether a cluster should report healthy when a node cannot apply its configuration.

### Testing

`make test` passes, 136 specs.

`valkeycluster_exporter_test.go` covers, against envtest:

- `spec.exporter` omitted (via an unstructured create, since a typed struct never satisfies `omitempty`), explicitly empty, and each of `image` / `resources` / `args` / `securityContext` set alone: all resolve to enabled, propagate an explicit `true` to the node spec, and render the sidecar
- an explicit `enabled: false` is honoured, resolves to nil on the node spec, and survives an update, the regression guard for the omitempty trap above
- a node stored by an older operator (enabled absent) produces no `nodeRequiresRoll` diff, the regression guard for the upgrade failover (#401)
- a bare `ValkeyNode` is not stamped with `enabled: true`, the regression guard for the field-default trap above

`TestBuildValkeyNodePodTemplateSpec_ExporterUnsetOmitsSidecar` pins the node-level contract: no explicit `true`, no sidecar.

The full e2e suite passes on a local kind cluster.

Upgrade behaviour verified on EKS 1.35: with the previous head, upgrading the operator under a disabled-exporter cluster wrote `enabled: false` to the node specs and ran a proactive failover within seconds of leader election, with no actual pod roll. With this head the same upgrade leaves specs, pods and roles untouched, and a cluster with only `exporter.image` set gains the sidecar through one legitimate roll.

Also verified by hand on EKS 1.35 before the fix: six clusters differing only in `spec.exporter`, the four partial ones came up with a single `server` container and no warning of any kind.

### Checklist


bcfmtolgahan:
Good catch on the defaulting assertions, that was a real hole. `IsEnabled()` returns true on a nil, so every one of those specs passed with the default marker removed. Given #395 is a marker that got dropped silently, a test that cannot detect a dropped marker was the wrong test to write.

Fixed in 7d99169: the specs now assert the stored object carries a non-nil `enabled` that dereferences to true, and that an explicit `false` stays a real false rather than going nil. Confirmed the other way too, by removing `+kubebuilder:default=true` and regenerating:

```
[FAILED] the CRD default must materialise enabled on the stored object
Summarizing 5 Failures:
  [FAIL] exporter enabled defaulting [It] enables the exporter when spec.exporter is omitted
  [FAIL] exporter enabled defaulting [It] keeps the exporter enabled when only image is set
  ...
```

All five defaulting specs fail without the marker and pass with it.

On the docstring coverage check, I've left that as is. Of the functions in the diff the only new exported one is `IsEnabled`, which is documented; the rest are Ginkgo spec closures and test fixtures, and adding docstrings to those to reach a percentage would be noise.

The `golangci-lint` tool error in the run above looks like the repo's custom build not resolving in your sandbox (`plugin "logcheck" not found`, from `.custom-gcl.yml`) rather than anything in this PR. `pre-commit run --all-files`, which builds the same custom linter, passes locally.

bcfmtolgahan:
@coderabbitai review

The earlier run hit the review limit, so the last review is against 70a2ae0. The defaulting assertions were tightened in 7d99169.

internal/controller/valkeycluster_exporter_test.go
bjosv, line 1: nit: since we test more than default, maybe the name could be `valkeycluster_exporter_test.go`

bjosv (dismissed):
Nice! LGTM

bcfmtolgahan:
The e2e failure was real. The field-level default stamped enabled: true onto bare ValkeyNodes too, so a standalone node gained a sidecar wired to _exporter credentials that only the cluster controller provisions, and its pod never became ready.

Fixed in 8380580: the default is resolved in the cluster controller instead. A nil enabled still means enabled there, now also covering an explicitly empty exporter block, and nodes always receive an explicit value; a ValkeyNode runs the sidecar only on an explicit true. Regression tests added at both levels, and the failing e2e labels (valkeynode, metrics-exporter) pass on a local kind cluster. PR description updated to match.

internal/controller/valkeycluster_exporter_test.go
bcfmtolgahan: Renamed in faa4e2e. It fits even better after 8380580: the file now covers defaulting, propagation and the node-level contract.

internal/controller/valkeycluster_controller.go
```diff
@@ -917,6 +917,12 @@ func buildClusterValkeyNode(cluster *valkeyiov1alpha1.ValkeyCluster, shardIndex
 	// supplies, so the escape hatch is left verbatim.
 	nodeSelector := withZonePin(scheduling.NodeSelector, zoneForPod(effectiveZonePinning(cluster.Spec.Scheduling), shardIndex, nodeIndex))
 
+	// Resolve the cluster-level exporter default (nil means enabled) into an
+	// explicit value: a ValkeyNode runs the sidecar only on an explicit true.
+	exporter := cluster.Spec.Exporter
+	exporterEnabled := exporter.IsEnabled()
+	exporter.Enabled = &exporterEnabled
+
```
jdheyburn, line 930: I did some testing, and this exposes a bug in how we calculate whether rolls are needed once the operator is upgraded, causing failovers. Can you change it to the below:

```
      exporter := cluster.Spec.Exporter
      exporter.Enabled = nil
      if cluster.Spec.Exporter.IsEnabled() {
              enabled := true
              exporter.Enabled = &enabled
      }
```

This then needs the two assertions in `valkeycluster_exporter_test.go:153-154`:

```go
      node := buildClusterValkeyNode(stored, 0, 0)
      Expect(node.Spec.Exporter.Enabled).To(BeNil(),
              "disabled must resolve to nil on the node spec — an explicit false rolls upgraded clusters")
```


I'll work on a follow up to fix this in #401

api/v1alpha1/valkeycluster_types.go
```diff
@@ -459,6 +462,13 @@ type ExporterSpec struct {
 	Args []string `json:"args,omitempty"`
 }
 
+// IsEnabled resolves the ValkeyCluster-level default: a nil Enabled means
+// enabled, so overriding any other exporter field keeps the sidecar. Not for
+// ValkeyNode specs, where nil means disabled.
+func (e ExporterSpec) IsEnabled() bool {
+	return e.Enabled == nil || *e.Enabled
+}
```
jdheyburn: I think this has exposed an issue where we share the ExporterSpec struct with ValkeyNode, where on ValkeyCluster nil == enabled, and on ValkeyNode it is explicitly required. Can you replace this with the below:

```go
// ExporterEnabled resolves the cluster-level default: a nil
// spec.exporter.enabled means enabled, so overriding any other exporter
// field keeps the sidecar.
func (s ValkeyClusterSpec) ExporterEnabled() bool {
      return s.Exporter.Enabled == nil || *s.Exporter.Enabled
}
```

Then change these call sites:
- in users.go from `cluster.Spec.Exporter.IsEnabled()` to ` cluster.Spec.ExporterEnabled()`: lines 163, 405, 454.
- `valkeycluster_controller.go:923 (buildClusterValkeyNode): exporter.IsEnabled() → cluster.Spec.ExporterEnabled()`
- the test assertions in `valkeycluster_exporter_test.go` (lines 71,150) become `Expect(stored.Spec.ExporterEnabled()).To(BeTrue())`

That should avoid any reuse of ExporterSpec.IsEnabled outside of ValkeyCluster.

jdheyburn (commented):
Thanks for raising this! I had a couple of comments, would you be able to take a look please?

internal/controller/valkeycluster_controller.go
bcfmtolgahan: Confirmed and fixed in 24c88e6, with the two assertions as suggested plus a regression test that feeds nodeRequiresRoll a node stored by an older operator. Reproduced the failover on EKS with the previous head (proactive failover seconds after leader election, no actual pod roll); with this head the same upgrade leaves specs, pods and roles untouched.

api/v1alpha1/valkeycluster_types.go
bcfmtolgahan: Done in 24c88e6: IsEnabled is gone, the resolver is ValkeyClusterSpec.ExporterEnabled, and all call sites and test assertions use it.

jdheyburn (approved):
Thanks for raising this issue and for fixing it!

PR #398, opened 2026-08-21
fix: generate the commands.allow/deny pattern and widen it to what Valkey accepts

### LINKED ISSUE #395: [BUG]: commands.allow/deny pattern marker is malformed, so no validation is generated

bcfmtolgahan:
This PR closes #395

### Summary

The pattern markers on `commands.allow` and `commands.deny` were malformed, so controller-gen dropped them and the CRD validated nothing. This makes them generate, and widens the expression to match what Valkey actually accepts.

- `commands.allow` and `commands.deny` entries are now validated at admission.
- Entries that Valkey accepts keep working, including subcommands, and names carrying a digit, underscore or hyphen.

### Implementation

The marker read `+kubebuilder:validation:Items:Pattern=^[@a-z|]+$}` — capitalised `Items`, plus a stray brace. controller-gen ignores unknown markers without warning, so the generated schema was a bare `items: {type: string}`.

Lowercasing `items` alone is not enough. The original expression rejects 15 of the 396 commands in `COMMAND LIST` on valkey 9.0.0:

```
bitfield_ro          client|import-source  client|no-evict
client|no-touch      cluster|count-failure-reports
cluster|set-config-epoch                   cluster|slot-stats
eval_ro              evalsha_ro            fcall_ro
georadius_ro         georadiusbymember_ro  memory|malloc-stats
restore-asking       sort_ro
```

`cluster|set-config-epoch` is in the operator's own `_operator` ACL. Valkey also accepts uppercase and normalises it on the way in:

```
127.0.0.1:6379> ACL SETUSER upper on nopass +GET ~*
OK
127.0.0.1:6379> ACL LIST
user upper on nopass sanitize-payload ~* resetchannels -@all +get
```

So a lowercase-only pattern would reject specs that work today.

The expression matches the grammar rather than just the alphabet:

```
^(@[A-Za-z]+|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*(\|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*)?)$
```

A category is `@` plus letters; a command is a name of dot-separated segments, each of hyphen-separated words, optionally followed by one pipe and a subcommand of the same shape. Dots cover module commands (`json.set`, `bf.add`, `ft._list`), which are in no category except `@all` and can only be granted individually; hyphens sit only inside a segment, so `-get`, `get-` and `a--b` are rejected. Checked against valkey 9.0.0: all 396 entries of `COMMAND LIST` and all 21 `ACL CAT` categories match, uppercase forms included.

### Limitations

This is admission-level validation only. It rejects strings that cannot be a command or category; it does not check the name against the running server's command table, so a well-formed but nonexistent command (`+notacommand`) still fails later at `ACL LOAD`. Catching that needs a server round trip.

The rule is new validation on an existing `v1alpha1` field, so a stored CR carrying free text will now fail on update. Such a CR is already broken at runtime, since Valkey rejects the ACL file either way.

#396 covers the separate reporting problem: while a node cannot apply the ACL, the cluster still reports `Ready/ClusterHealthy`.

### Testing

`make test` passes.

New `valkeycluster_acl_commands_validation_test.go` asserts against envtest that the awkward-but-valid entries above are accepted, including uppercase forms, that free text, embedded spaces, a leading `+` and an empty string are rejected in both `allow` and `deny`, and that structurally invalid entries carrying only legal characters are rejected: `@`, `@@read`, `@-read`, `|`, `get|`, `|get`, `get||set`, `client|no|evict`.

Two negative controls, both confirmed:

- restoring the malformed marker fails the specs on `entry "THIS IS NOT A COMMAND !!!" must be rejected`
- restoring a character-class-only pattern fails them on `entry "@" must be rejected`

Valkey 9.0.0 rejects all eight of the structurally invalid entries above, e.g. `ERR Error in ACL SETUSER modifier '+get|': Syntax error`. Also scanned the repo for existing command values the pattern would reject; there are none.

### Documentation

`docs/valkeycluster.md` already described the three accepted forms. Added the validation rule itself to the `commands` bullet, plus the caveat that the name is not checked against the server's command table.

### Checklist

### DIFF
--- a/api/v1alpha1/valkeyacls_types.go
+++ b/api/v1alpha1/valkeyacls_types.go
@@ -76,11 +76,11 @@ type CommandsAclSpec struct {
 	// Subcommands (client|setname, config|get, etc.)
 
 	// Allowed commands for this user
-	// +kubebuilder:validation:Items:Pattern=^[@a-z|]+$}
+	// +kubebuilder:validation:items:Pattern=`^[@A-Za-z0-9|_-]+$`
 	Allow []string `json:"allow,omitempty"`
 
 	// Denied commands for this user
-	// +kubebuilder:validation:Items:Pattern=^[@a-z|]+$}
+	// +kubebuilder:validation:items:Pattern=`^[@A-Za-z0-9|_-]+$`
 	Deny []string `json:"deny,omitempty"`
 }
 
--- a/api/v1alpha1/valkeyacls_types.go
+++ b/api/v1alpha1/valkeyacls_types.go
@@ -76,11 +76,11 @@ type CommandsAclSpec struct {
 	// Subcommands (client|setname, config|get, etc.)
 
 	// Allowed commands for this user
-	// +kubebuilder:validation:items:Pattern=`^[@A-Za-z0-9|_-]+$`
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\|[A-Za-z0-9_-]+)?)$`
 	Allow []string `json:"allow,omitempty"`
 
 	// Denied commands for this user
-	// +kubebuilder:validation:items:Pattern=`^[@A-Za-z0-9|_-]+$`
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\|[A-Za-z0-9_-]+)?)$`
 	Deny []string `json:"deny,omitempty"`
 }
 
--- a/api/v1alpha1/valkeyacls_types.go
+++ b/api/v1alpha1/valkeyacls_types.go
@@ -77,11 +77,11 @@ type CommandsAclSpec struct {
 	// Subcommands (client|setname, config|get, etc.)
 
 	// Allowed commands for this user
-	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*(\|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*)?)$`
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*(\|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*)?)$`
 	Allow []string `json:"allow,omitempty"`
 
 	// Denied commands for this user
-	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*(\|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*)?)$`
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*(\|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*)?)$`
 	Deny []string `json:"deny,omitempty"`
 }
 
--- a/api/v1alpha1/valkeyacls_types.go
+++ b/api/v1alpha1/valkeyacls_types.go
@@ -73,14 +73,15 @@ type CommandsAclSpec struct {
 
 	// Command categories (@all, @read, @write, @admin, etc.)
 	// Individual commands (get, set, ping, etc.)
+	// Module commands (json.set, bf.add, etc.)
 	// Subcommands (client|setname, config|get, etc.)
 
 	// Allowed commands for this user
-	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\|[A-Za-z0-9_-]+)?)$`
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*(\|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*)?)$`
 	Allow []string `json:"allow,omitempty"`
 
 	// Denied commands for this user
-	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\|[A-Za-z0-9_-]+)?)$`
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*(\|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*)?)$`
 	Deny []string `json:"deny,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


bcfmtolgahan:
Both of you flagged the same real gap and you were right: the expression only constrained the alphabet, so `@`, `@@read`, `@-read`, `|`, `get|`, `|get`, `get||set` and `client|no|evict` still got through. That is the exact failure this validation is supposed to prevent, so the fix was incomplete.

Reproduced against valkey 9.0.0 before changing anything:

```
+@                ERR Error in ACL SETUSER modifier '+@': Unknown command or category name in ACL
+@@read           ERR ... '+@@read': Unknown command or category name in ACL
+get|             ERR ... '+get|': Syntax error
+|get             ERR ... '+|get': Unknown command or category name in ACL
+get||set         ERR ... '+get||set': Unknown command or category name in ACL
+client|no|evict  ERR ... '+client|no|evict': Unknown command or category name in ACL
```

Fixed in c98d791 by matching the grammar instead of the character set:

```
^(@[A-Za-z]+|[A-Za-z0-9_-]+(\|[A-Za-z0-9_-]+)?)$
```

Validated both directions against the same server: all 396 entries of `COMMAND LIST` and all 21 `ACL CAT` categories match, uppercase forms included, and all twelve malformed cases are rejected. `COMMAND LIST` also confirms no command carries a second pipe or a digit, so the single optional `|<subcommand>` is the right shape. Digits stay in the command alternative anyway, since rejecting a future command that has one would be worse than letting Valkey catch it.

Boundary cases added to the envtest specs, and a second negative control: with a character-class-only pattern they now fail on `entry "@" must be rejected`.

Still out of scope, and called out in the PR description: a well-formed but nonexistent command such as `+notacommand` passes admission and is only caught later at `ACL LOAD`. Catching that needs the server's command table.

bcfmtolgahan:
@coderabbitai review

The finding you and Greptile both raised is addressed in c98d791; the reviews above are against 92415de. Re-running so the verdict reflects the current head.

api/v1alpha1/valkeyacls_types.go
```diff
@@ -76,11 +76,11 @@ type CommandsAclSpec struct {
 	// Subcommands (client|setname, config|get, etc.)
 
 	// Allowed commands for this user
-	// +kubebuilder:validation:Items:Pattern=^[@a-z|]+$}
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\|[A-Za-z0-9_-]+)?)$`
```
bjosv: I believe module commands includes a `.` like [JSON.SET](https://valkey.io/topics/valkey-json/), should we allow that?

bcfmtolgahan: Yes. Module commands are in no category except @all, so granting them individually is the only option and the pattern must accept them. Fixed in a628eb0: names are dot-separated segments on both sides of the pipe; leading, trailing and doubled dots stay rejected.

api/v1alpha1/valkeyacls_types.go
```diff
@@ -73,14 +73,15 @@ type CommandsAclSpec struct {
 
 	// Command categories (@all, @read, @write, @admin, etc.)
 	// Individual commands (get, set, ping, etc.)
+	// Module commands (json.set, bf.add, etc.)
 	// Subcommands (client|setname, config|get, etc.)
 
 	// Allowed commands for this user
-	// +kubebuilder:validation:Items:Pattern=^[@a-z|]+$}
+	// +kubebuilder:validation:items:Pattern=`^(@[A-Za-z]+|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*(\|[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*)?)$`
```
jdheyburn: This regex accepts a hypend sitting at the segment edge, so any call to `[A-Za-z0-9_-]+` should be replaced with `[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*`

So the resulting pattern would be:

```
^(@[A-Za-z]+|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*(\|[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*(\.[A-Za-z0-9_]+(-[A-Za-z0-9_]+)*)*)?)$
```

Can you add these test cases too that should be rejected? Maybe add them first before the fix:

```
-get
get-
-
--
-flushall
```

jdheyburn (commented):
Thanks for raising the issue and implementing a fix! I just had a comment on the regex to make it ironclad. I hope CEL will allow it to pass.

api/v1alpha1/valkeyacls_types.go
bcfmtolgahan: Done in 3b512fb: your pattern as proposed, with the five cases added red-first plus json.-set-style combinations. Checked the full accept/reject matrix under RE2 before pushing.

jdheyburn:
memo: Github Actions has an incident, retrigger onced resolved

bjosv (approved):

Issue #375, opened 2026-08-13
[FEATURE]: Support Matrix and Versioning


stockholmux:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.

### Problem / Use case

Looking forward to GA release, the project needs to determine both how we communicate the Kubernetes and Valkey versions in the operator as well as how we intend to support older versions.

### Proposed solution

Based on discussions in the weekly operator/helm meeting, I propose the following:

**Versioning**
Only one major version is supported at any given time. Minor versions in the major version line will get patches as needed. Previous major versions will not be patched. Any major versions below the current major version number will be considered EOL immediately upon release of the next major.

**Communication**
At the time of release, the operator project will provide a support matrix that in includes both Valkey and Kubernetes versions.

The first non-heading row will include the most recent version and go in reverse version order to the oldest.
On the horizontal axis it will include Y/N columns indicating operator EOL and support for Valkey or k8s minor/major versions.

Both Valkey and K8s do not rapidly release minor versions so this chart may grow horizontally, but at a reasonable rate. Only one previous major version will be included on the chart - everything else will be considered historical and will be dropped off the chart. This will contain any horizontal growth.

Here is a mockup (using Google Sheets) but it will be rendered via tables in Markdown and/or the Valkey.io docs/website.

<img width="827" height="150" alt="Image" src="https://github.com/user-attachments/assets/f718a9c6-7150-4e62-9a3d-450184262359" />

### Scope of change

Small (small CRD field or controller behaviour tweak)

### API / behaviour impact

- [ ] This change is backwards-compatible with existing ValkeyCluster CRs.
- [ ] This change introduces a breaking API change in `v1alpha1`.

### Contribution

- [x] I would like to implement this feature myself.
- [x] I would like to help review or test this feature.

sgissi:
[...]

PR #399, opened 2026-08-24
fix: add validation for rejecting operator owned keys

### LINKED ISSUE #393: [BUG]: User config silently overriding operator-owned directives should be surfaced
### Pre-flight checks

### Bug Description

Operator overrides user specified configs via base config.
However, users never get rejection or error message while passing such config params.

A user who sets `dir`, `aclfile`, `cluster-config-file`, `cluster-enabled`, `protected-mode`, or any `tls-*` key in `spec.config` gets no signal that the value was discarded.
The bug is -  specified spec doesn't match the deployed spec.

### Steps to Reproduce

Manifests with `cluster-enabled`, `dir`, `aclfile` and/or `tls-*` will get admitted.

### Expected Behaviour

User gets error on invalid config parameter.

### Actual Behaviour

Manifest is admitted.
Deployed spec differs from admitted manifest.

```yaml

```

### Operator logs

```shell

```

### Valkey pod logs

```shell

```

### Operator version

main

### Kubernetes version

v1.35.0

kind

### Additional context

_No response_

deepakpunjabi:
This PR closes https://github.com/valkey-io/valkey-operator/issues/393.

### Summary

Previously, setting options in `ValkeyCluster.spec.config` that the operator manages internally caused a silent mismatch. Because Valkey uses the last entry in its configuration file, the operator's default settings silently overwrote user-defined settings without showing an error.

When TLS was disabled, the operator didn't override `port` or `tls-` settings. A user could accidentally change or close the port the operator needed to connect, breaking the deployment.

`spec.config` now rejects below operator-owned keys(case-insensitive): 
- `aclfile`
- `cluster-allow-replica-migration`
- `cluster-config-file`
- `cluster-enabled`
- `cluster-node-timeout`
- `cluster-replica-validity-factor`
- `dir`
- `port`
- `protected-mode`
- `shutdown-on-sigterm`
- `tls-auth-clients`
- `tls-ca-cert-file`
- `tls-cert-file`
- `tls-cluster`
- `tls-key-file`
- `tls-port`
- `tls-replication`

Cluster directives the operator does not set(`cluster-require-full-coverage`, `cluster-migration-barrier` and `cluster-allow-reads-when-down`) stay available.

Rule is applied on create and update both.

### Implementation

- `api/v1alpha1/valkeycluster_types.go` adds `ReservedConfigKeys` and a field-level CEL rule on `Config`.
- Two new test files. 
- CRD is regenerated.

- `spec.config` sets `maxProperties: 1000` to accept the CRD. Valkey has ~200 directives as of now.
- `config_reserved_keys_test.go` adds test for config drift as config is mentioned in 3 places - `getBaseConfig`, `ReservedConfigKeys` and the CEL rule.
- `TestReservedConfigKeysMatchCELRule` extracts the CEL list out of `valkeycluster_types.go` to compare it.

### Limitations

- This is a breaking change for clusters that already set one of these keys. This should be fine as users were never intended to use this keys and we are on v1alpha1.
- `ValkeyNode.spec.config` is not covered in this PR.

### Testing

`config_reserved_keys_test.go`, 3 plain Go tests without cluster:

- `TestReservedConfigKeysMatchBaseConfig` derives the real key set by calling `getBaseConfig` with TLS both off and on, and asserts it equals `ReservedConfigKeys`.
- `TestReservedConfigKeysMatchCELRule` parses the CEL literal from source and compares it to the same list.
- `TestReservedConfigKeysAreLowercase` guards the lowercased comparison, since an uppercase entry could never match.

`valkeycluster_config_validation_test.go`, 26 envtest specs against a Kubernetes API server:

- 17 specs, one per reserved key, each asserting rejection. Generated from `ReservedConfigKeys` so the coverage cannot fall behind the list.
- 3 specs for case-insensitivity: `Cluster-Enabled`, `PROTECTED-MODE`, `TLS-Port`.
- 3 specs for ordinary tunables, the cluster directives the operator does not own, and an absent `config`.
- Rejection when a reserved key is mixed in with valid ones.
- Rejection when a reserved key is added by update rather than create.
- The message names the offending keys, since there is no webhook to elaborate.

- `make test`, `make lint`, `pre-commit run --all-files` passes.

### Checklist

### DIFF
--- a/api/v1alpha1/valkeycluster_types.go
+++ b/api/v1alpha1/valkeycluster_types.go
@@ -266,6 +266,35 @@ type ZonePinning struct {
 	Zones []string `json:"zones"`
 }
 
+// ReservedConfigKeys are the valkey.conf directives the operator sets itself.
+// A user value for one of these in spec.config is refused rather than silently dropped.
+//
+// The list has to match the keys the operator's base config emits, with TLS both on and off.
+// TestReservedConfigKeysMatchBaseConfig enforces that.
+// It also has to match the CEL rule on ValkeyClusterSpec.Config, which cannot reference a Go value.
+// Adding a key therefore means editing both, and the tests say so when you miss one.
+//
+// Lowercase entries only, because the CEL rule lowercases before comparing.
+var ReservedConfigKeys = []string{
+	"aclfile",
+	"cluster-allow-replica-migration",
+	"cluster-config-file",
+	"cluster-enabled",
+	"cluster-node-timeout",
+	"cluster-replica-validity-factor",
+	"dir",
+	"port",
+	"protected-mode",
+	"shutdown-on-sigterm",
+	"tls-auth-clients",
+	"tls-ca-cert-file",
+	"tls-cert-file",
+	"tls-cluster",
+	"tls-key-file",
+	"tls-port",
+	"tls-replication",
+}
+
 // ValkeyClusterSpec defines the desired state of ValkeyCluster.
 // +kubebuilder:validation:XValidation:rule="!(has(self.persistence) && self.workloadType == 'Deployment')",message="persistence requires workloadType StatefulSet"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.persistence) || has(self.persistence)",message="persistence cannot be removed once set"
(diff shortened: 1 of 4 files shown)


melancholictheory:
Ran this on a k3d cluster against a live API server rather than only through the suite, since a CEL rule is only as good as what the apiserver actually admits. Rejection works for `aclfile`, and the `lowerAscii()` handling holds for `ACLFILE` and for `Cluster-Node-Timeout`. Keys the operator does not set, `cluster-require-full-coverage` and `cluster-migration-barrier`, are still accepted, which is the carve-out the comment promises.

The part worth adding to the description is the upgrade path, because it is better than a reader would assume and nothing here says so. I created a cluster carrying `dir` under the current CRD, then installed this one on top:

- patching an unrelated field, `replicas`, succeeds and the apiserver attaches a warning quoting the rule. That is validation ratcheting: the unchanged offending field does not block the write, so nobody's existing cluster becomes unmanageable the moment the CRD lands.
- patching `spec.config` while keeping the reserved key is refused, which is the point.
- removing the key succeeds, so the migration is one edit.

Existing clusters therefore keep reconciling and get nudged on every unrelated edit rather than being wedged. Worth stating, since "operator-owned keys are now rejected" reads much scarier than that.

One observation that is not about this PR. `shutdown-on-sigterm` is on the list, and `getBaseConfig` emits it unconditionally with a comment saying it needs Valkey 9.0+. Valkey refuses to start on a directive it does not know, so the operator already cannot run against 8.x whatever the reserved list says. That is a data point for the support matrix in #375, and a second candidate for the gated set in #307, where `tls-auto-reload-interval` is currently the only entry.

On that: `tls-auto-reload-interval` is deliberately absent from `ReservedConfigKeys`, so #307 can still gate a user-set value rather than the API refusing it outright. That reads correct to me, and it is worth keeping in mind before anyone adds it here later.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -266,6 +266,35 @@ type ZonePinning struct {
...
+	"aclfile",
+	"cluster-allow-replica-migration",
+	"cluster-config-file",
+	"cluster-enabled",
+	"cluster-node-timeout",
+	"cluster-replica-validity-factor",
+	"dir",
+	"port",
+	"protected-mode",
+	"shutdown-on-sigterm",
```
jdheyburn, line 288: We could allow users to override this default. But I would say let's wait for the issue to come in first requesting the override so that we can understand the use case.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -359,7 +388,23 @@ type ValkeyClusterSpec struct {
...
+	// It is set high deliberately, because raising a bound later is backwards compatible while lowering one locks out anyone already above it.
+	// Valkey has roughly 200 directives in total, so 1000 cannot realistically be reached.
+	// +kubebuilder:validation:MaxProperties=1000
+	// +kubebuilder:validation:XValidation:rule="self.all(key, !(key.lowerAscii() in ['aclfile','cluster-allow-replica-migration','cluster-config-file','cluster-enabled','cluster-node-timeout','cluster-replica-validity-factor','dir','port','protected-mode','shutdown-on-sigterm','tls-auth-clients','tls-ca-cert-file','tls-cert-file','tls-cluster','tls-key-file','tls-port','tls-replication']))",message="spec.config must not set operator-owned keys (aclfile, cluster-allow-replica-migration, cluster-config-file, cluster-enabled, cluster-node-timeout, cluster-replica-validity-factor, dir, port, protected-mode, shutdown-on-sigterm, tls-auth-clients, tls-ca-cert-file, tls-cert-file, tls-cluster, tls-key-file, tls-port, tls-replication): the operator sets these itself and a user value would be ignored or would break its connection to the nodes"
```
jdheyburn, line 407: Instead of listing these out twice, could we link to documentation where folks can go to find the blocked configs?

You would also need to add the documentation, which was going to be another comment I had :) Perhaps in docs/valkeycluster.md under config.

api/v1alpha1/valkeycluster_types.go
```diff
@@ -266,6 +266,35 @@ type ZonePinning struct {
...
+	"protected-mode",
+	"shutdown-on-sigterm",
+	"tls-auth-clients",
+	"tls-ca-cert-file",
+	"tls-cert-file",
+	"tls-cluster",
+	"tls-key-file",
+	"tls-port",
+	"tls-replication",
+}
```
jdheyburn, line 296: Can you also add these, which are managed but applied directly on the pod.

```
cluster-announce-ip
primaryuser
primaryauth
```

PR #402, opened 2026-08-26
fix: use workloadRevision for determining proactive failover

### LINKED ISSUE #401: Roll detection compares specs, not rendered templates — spec-shape changes cause spurious proactive failovers on operator upgrade
### Summary

`nodeRequiresRoll` (`internal/controller/failover.go`) predicts "will this spec update roll pods?" with `equality.Semantic.DeepEqual` over the whole `ValkeyNode` spec (minus `Config`). A pod roll is actually determined by the rendered pod template — the spec is only an input to it. Any spec-shape change that leaves the template identical is therefore misread as a roll, and the cost of a false positive here is a real, client-visible proactive `CLUSTER FAILOVER` per shard primary.

### Concrete instance

Found while reviewing #397, and reproduced on a kind cluster. That PR changes `ExporterSpec.Enabled` from `bool` to `*bool` and stamps an explicit value onto every node spec. Nodes written by the current operator stored `enabled: false` as an absent field (`omitempty`), which decodes to nil under `*bool`; the desired spec carries `&false`. The templates are identical — nil and false both mean no sidecar on a `ValkeyNode` — but `DeepEqual` sees nil ≠ `&false`, `needsProactiveFailoverForRoll` sees a non-`WorkloadRevision` diff, and on the first reconcile after the operator upgrade every exporter-disabled cluster runs a failover per primary. Observed: `FailoverInitiated`/`FailoverCompleted` events within seconds of the upgraded operator taking leadership, while pod UIDs and restart counts were untouched.

#397 can sidestep its instance (stamp the pointer only when true), but the class remains: every future spec evolution — a new field with a controller-resolved default, a pointer conversion, an encoding change — re-creates this hazard on upgrade.

### History

This class has been patched around before:

- The `liveTemplateHash` backfill escape inside `needsProactiveFailoverForRoll` exists to suppress one such false positive (empty `WorkloadRevision` being backfilled).
- #381 (serverConfigHash removal) and #317 (no-op reconciles from defaulted fields) were adjacent fights with spec-encoding drift being mistaken for real change.

### Proposal

Decide roll-vs-no-roll by comparing rendered template hashes rather than specs: render the desired pod template, hash it with the existing `podTemplateRollHash`, and compare against the live workload's template hash. Sketch:

```go
// A spec update rolls pods only if it changes the rendered template,
// not merely the spec encoding.
desired := buildValkeyNodePodTemplateSpec(desiredNode, labels)
return liveTemplateHash != "" && liveTemplateHash != podTemplateRollHash(desired)
```

This makes the false-positive class structurally impossible and should subsume the `liveTemplateHash` backfill special case rather than adding to it.

### Edge cases to work through

- Live template hash unavailable (workload not yet created, first reconcile after CRD adoption).
- Spec changes that disrupt pods without changing the template, if any (PVC resize/reclaim, `workloadType` StatefulSet↔Deployment) — confirm they are handled by their own paths or fold them in deliberately.
- Interaction with `WorkloadRevision` authorization: today the spec-diff check and the revision check are separate branches; a template-hash comparison may collapse them.

### Acceptance

- Upgrading the operator across a spec-shape change (e.g. the #397 `bool`→`*bool` conversion, with or without its stamping fix) produces zero `FailoverInitiated` events and zero pod restarts for clusters whose rendered templates are unchanged.
- Genuine template changes (image bump, sidecar added/removed, resources changed) still fail over proactively before the roll.

jdheyburn:
This PR closes #401

### Summary

needsProactiveFailoverForRoll compared specs with DeepEqual, so any changes to spec rendering an identical pod template (e.g. the #397 bool→*bool conversion) was misread as a roll and triggered a client-visible CLUSTER FAILOVER per shard primary on operator upgrade. Roll-vs-no-roll is now decided by comparing the live workload's template hash against the desired WorkloadRevision (both already podTemplateRollHash), making the false-positive class structurally impossible and subsuming the WorkloadRevision backfill special case. nodeRequiresRoll is deleted.

### Implementation

- A StatefulSet↔Deployment swap is guarded explicitly (both kinds render the same template); "" normalizes to StatefulSet so the CRD-default backfill isn't mistaken for a swap.
- Unknown live template hash means the update creates a workload rather than rolling one, and no failover.
- The WorkloadRevision authorization flow in the ValkeyNode controller is unchanged

### Limitations

- Unknown live template skips failover. When the StatefulSet/Deployment object is absent, the predicate returns false on the grounds that the update creates rather than rolls. The corner case: a deleted workload whose pod survives gets adopted on recreate and could roll under a new template without proactive failover. Accepted as rarer than the false-positive class this removes.
- Disruptions must reach the template (or be guarded). The decision now sees only the rendered template plus the explicit WorkloadType guard. Any future spec field that replaces pods without changing the template needs its own guard, as WorkloadType did.
- The WorkloadType-swap guard is defensive, not a supported path. The node controller has no stale-workload cleanup on a type switch, so the guard fails over conservatively but the swap itself remains rough; out of scope here.
- Validation simulated the shape change. The kind A/B test induced the class via a stale workloadRevision patch rather than the literal #397 bool→*bool upgrade; the unit tests cover the class, and #397 rebased onto this will exercise the real instance.

### Testing

Kind A/B test on issue-401: stale-revision mutation under the old operator → 2 spurious failovers with zero pod changes; identical mutation under this branch → zero failover events, revisions silently rewritten; control (real template change) → proactive failover per shard before each primary roll.

### Checklist


jdheyburn:
@coderabbitai @greptile-apps review

bjosv (dismissed):
Well spotted

bjosv (approved):
Worked fine when testing locally

melancholictheory:
This landed while I was testing it, so treat the below as after the fact rather than a review. I built 17e0acb, which is what merged, and ran it on k3d instead of only reading the diff, since the change moves the failover decision onto a comparison that now runs on every reconcile rather than in one narrow branch. Cluster of 3 shards with 1 replica each.

Steady state, roughly 100 seconds of reconciles with nothing changed: no failover events. That is the part I most wanted to see, because if the live hash and `desired.Spec.WorkloadRevision` disagreed systematically (API server defaulting, for instance), the new predicate would fire on every pass rather than never.

Operator restart: no failover events, and pod UIDs and restart counts were untouched. Worth being precise about what that does and does not cover, though. Restarting the same build exercises a reconcile after a leader change against specs the same version wrote. It is not the #397 case, where the stored encoding was written by an older operator, so it is weaker evidence than a genuine cross-version upgrade would be.

Real template change (bumped `resources.requests`): `FailoverInitiated` fired, the roll completed, cluster back to Ready, and the sequencing in the log is the one you would want:

```
primary has no synced replicas, deferring roll   name=rolltest-2-0 shardNodes=1
proactive failover before rolling primary        name=rolltest-2-0 syncedReplicas=1
initiating proactive failover                    target=10.42.1.119
proactive failover completed                     newPrimary=10.42.1.119
```

One failover per shard, no repeats while the roll was pending, which was my other worry: the predicate stays true from the moment the revision is authorized until the workload actually updates, and `gateRollingWorkloadUpdate` can hold that window open for a while.

On the edge cases #401 listed as needing work, the audit comes out clean and I think it is worth saying so explicitly rather than leaving it open. Going through `ValkeyNodeSpec`, the only field that does not reach the rendered pod template is `Persistence`, and it cannot roll pods: there is no `volumeClaimTemplates` anywhere in the tree, the operator manages the PVC as its own object and mounts it by a stable claim name from `valkeyNodePVCName`, so a size or reclaim-policy edit changes the PVC and leaves the template identical. `WorkloadType` is the other non-template case and the new explicit check covers it. Everything else lands in the template, `Config` included, through the derived annotation. So dropping the spec comparison does not silently lose a disruptive case.

One follow-up sized thing, since it is no longer a review comment. `effectiveWorkloadType` and `livePodTemplateHash` now spell out the same "empty means StatefulSet" rule in two places, the latter as `case Deployment` with everything else falling to `default`. Worth collapsing onto the helper at some point so the two cannot drift if that CRD default ever changes.

PR #409, opened 2026-08-27
feat: add networking.tls.serverName for operator TLS verify

### LINKED ISSUE #405: [FEATURE]: spec.networking.tls.serverName for operator hostname verify
### Pre-flight checks

### Problem / Use case

We're on 0.5.0. The operator always sets ServerName to: `valkey-<name>.<ns>.svc.cluster.local`. No override, no InsecureSkipVerify on that path.
- https://github.com/valkey-io/valkey-operator/blob/v0.5.0/internal/controller/valkeycluster_controller.go#L944
- https://github.com/valkey-io/valkey-operator/blob/v0.5.0/internal/controller/valkeynode_controller.go#L878-L881
- https://github.com/valkey-io/valkey-operator/blob/v0.5.0/internal/controller/utils.go#L337-L341

Our cert policy requires the kube cluster in the SAN (a cert from kube-dev-cluster shouldn't work on kube-prod-cluster). So we can't put `valkey-foo.valkey.svc.cluster.local` on the cert — that name looks the same on every cluster.

#378 doesn't help. That's CLUSTER SLOTS announce, not this pin.

### Proposed solution

`spec.networking.tls.serverName`:

- unset: keep `valkey-<name>.<ns>.svc.cluster.local`
- set: use that string, still dial the pod IP

e.g. we'd set `valkey-foo.valkey.kube-dev-cluster.svc.cluster.local`

CNPG does this with cnpg.io/pluginServerName
https://github.com/cloudnative-pg/cloudnative-pg/blob/main/pkg/utils/labels_annotations.go
https://github.com/cloudnative-pg/cloudnative-pg/pull/9222

### Scope of change

### API / behaviour impact

### Contribution

### Additional context

#318 #360 #365 #297

#378 / #365 is hostname announce for apps. #360 is splitting TLS secrets. Neither changes `getTLSConfig` ServerName. #297 thread said the operator does not use InsecureSkipVerify.

spotu-dev:
This PR closes #405

### Summary

The operator's TLS clients always pin `valkey-<name>.<namespace>.svc.cluster.local` when they dial a node by pod IP. That fails if the certificate does not carry that SAN.

This adds `spec.networking.tls.serverName`. When set, both operator clients use it. When unset, the operator still pins the same default FQDN.

- New optional field `spec.networking.tls.serverName` on ValkeyCluster.
- Same field on `ValkeyNode.spec.tls.serverName`; the cluster controller copies the resolved name down.
- Default remains `valkey-<name>.<namespace>.svc.cluster.local`.

**Upgrade (TLS + exporter):** the operator now sets `REDIS_EXPORTER_TLS_SERVER_NAME` to that FQDN on every TLS cluster, even if you did not set `serverName`. The exporter still dials `localhost`; this only changes certificate verify/SNI (it used to verify `localhost`). That env lives in the pod template, so **every existing TLS cluster with the metrics exporter rolls once** on operator upgrade. After #402, a primary in that roll also does `CLUSTER FAILOVER`. Please mention this in the next release notes. Clusters without TLS, or with `exporter.enabled: false`, do not roll from this.

### Implementation

`getTLSConfig` already takes a `serverName`. The two callers that used to hardcode the headless FQDN now go through `tlsServerName`:

- `getValkeyClusterState` (cluster scrape)
- `buildNodeClientOption` (node client)

Putting the field only on the node would leave the cluster scrape still pinning the old name.

### Limitations

This only changes what the operator verifies. It does not change `CLUSTER SLOTS` announce (`cluster-announce-hostname`). That belongs on #367 / #365.

#378 will conflict on the two ServerName call sites. After that merges, unset should use `headlessServiceFQDN` (`clusterDomain` + trailing-dot default). This PR keeps today's `main` default.

### Testing

- `TestTLSServerName`: override, default FQDN, empty when there is no cluster name
- `TestNodeTLSFromCluster` / `TestBuildClusterValkeyNodeTLSServerName`: copy-down

```
go test ./internal/controller/ -run 'TestTLSServerName|TestNodeTLSFromCluster|TestBuildClusterValkeyNode'
```

### Checklist

Assisted by Grok 4.6. I reviewed the change and take responsibility for it.


spotu-dev:
@sandeepkunusoth thanks for approving the workflows.

Verify generated code failed while installing controller-gen (sum.golang.org INTERNAL_ERROR), not a dirty generate. I can't re-run from the fork.

https://github.com/valkey-io/valkey-operator/actions/runs/33086143637

Could you re-run failed jobs on that Lint run please?

sandeepkunusoth (dismissed):

spotu-dev:
@sandeepkunusoth 
Thanks for the review. All checks are green and you approved. I don't have merge rights on valkey-io/valkey-operator. Could you please merge #409?

internal/controller/valkeynode_controller.go
```diff
@@ -881,10 +881,7 @@ func (r *ValkeyNodeReconciler) buildNodeClientOption(ctx context.Context, node *
 	var tlsConfig *tls.Config
 	if node.Spec.TLS != nil && node.Spec.TLS.Certificates.Server.SecretName != "" {
 		secretName := node.Spec.TLS.Certificates.Server.SecretName
-		serverName := ""
-		if clusterName, ok := node.Labels[LabelCluster]; ok {
-			serverName = fmt.Sprintf("%s.%s.svc.cluster.local", headlessServiceName(clusterName), node.Namespace)
-		}
+		serverName := tlsServerName(node.Spec.TLS.ServerName, node.Labels[LabelCluster], node.Namespace)
```
jdheyburn: You can probably take `node.Spec.TLS.ServerName` verbatim here since Cluster would have already resolved this for you.

api/v1alpha1/valkeynode_types.go
```diff
@@ -146,6 +146,13 @@ type ValkeyNodeSpec struct {
 // API is the resolved view the node controller renders into valkey.conf and
 // volume mounts.
 type NodeTLSSpec struct {
+	// ServerName is the hostname used for TLS verification when connecting
+	// to the pod IP. The cluster controller copies this from
+	// spec.networking.tls.serverName for cluster-owned nodes.
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
```
jdheyburn, line 155: Would we want to validate on DNS-1123 too?

```go
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MaxLength=253
// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="must be a valid DNS-1123 subdomain (lowercase alphanumerics, '-' and '.', starting and ending with an alphanumeric)"
```

jdheyburn (commented):
Thanks for raising this! Besides the comments I had, if the exporter is enabled it will still be validating against `localhost`.

In metrics_exporter.go, redis-exporter has support for that - something like this:

```go
env = append(env, corev1.EnvVar{
    Name:  "REDIS_EXPORTER_TLS_SERVER_NAME",
    Value: tlsSpec.ServerName,
})
```

https://github.com/valkey-io/valkey-operator/blob/9f541b24fe1d09e51547306ded562ea27f6714ba/internal/controller/metrics_exporter.go#L57-L60

melancholictheory (commented):
The shape is right, and putting the field on both levels rather than the node alone is what keeps `getValkeyClusterState` from carrying on with the old pin. Three things from going through it.

**The exporter env var rolls every TLS cluster.** I measured this rather than eyeballing it, because `generateMetricsExporterContainerDef` writes into a container that is part of the pod template, and `normalizePodTemplate` strips only `ObjectMeta`, so container env is hashed. Same `ValkeyNode`, TLS on and exporter enabled, through `podTemplateRollHash`:

```
upstream/main   b8a062ffd92b4b9950eb29b6670ce7142839f21f23752f75cd2dee1c091383f5
this branch     b1d66b680f2f6ab036fa99bf6583e883207f9b6a56aa8ca17c11a030214da2ab
```

So every existing TLS cluster with an exporter gets a pod roll on operator upgrade, and now that #402 has merged, each shard primary also gets a proactive `CLUSTER FAILOVER` on the way through. Nothing here is wrong, but it is the kind of thing that is better decided than discovered on somebody's first upgrade.

There is a version of this that avoids it entirely. Keep `ServerName` on the node as the override only, leave it empty when the user has not set one, resolve the default at the point of use with `tlsServerName(node.Spec.TLS.ServerName, node.Labels[LabelCluster], node.Namespace)`, and append the exporter variable only when it is non-empty. Clusters that do not use the feature then render a byte-identical template and never roll; clusters that do opt in roll once, which is what they asked for.

That trade depends on something I could not tell from the diff. The exporter dials `localhost`, so today it verifies against `localhost`, and after this change it verifies against the headless FQDN. If that is a fix in its own right, it deserves to be called out as a behaviour change, because it is the part that forces the roll for everyone. If it is incidental to the main goal, gating it on an explicit `ServerName` costs nothing.

**The CEL rule rejects the form your own Limitations section predicts.** `dns1123Subdomain` requires the value to end with an alphanumeric, and #378 says announce and TLS FQDNs append a trailing dot. Checked against a live API server with this PR's CRD applied:

```
serverName: valkey-foo.valkey.kube-dev-cluster.svc.cluster.local     accepted
serverName: valkey-foo.valkey.svc.cluster.local.                     rejected
  must be a valid DNS-1123 subdomain (lowercase alphanumerics, '-' and '.',
  starting and ending with an alphanumeric)
```

A trailing dot is a normal way to write an FQDN and Go trims it before matching, so a user copying the name straight off their certificate can hit this. Once #378 makes the generated default carry a trailing dot, the operator will be producing a form that a user cannot type into the same field.

**Small one.** When `ServerName` resolves empty, which is a standalone `ValkeyNode` with TLS and no override, `REDIS_EXPORTER_TLS_SERVER_NAME` is still appended with an empty value rather than left out. Probably harmless, and it goes away on its own if the variable becomes conditional.

For what it is worth I checked `getBaseConfig` too, since `buildServerConfig` changed shape: it does not read `ServerName`, so `valkey.conf` is untouched and the roll above comes purely from the exporter container.

internal/controller/valkeynode_controller.go
spotu-dev: Cluster now resolves once in nodeTLSFromCluster(cluster) and writes that onto ValkeyNode.spec.tls.serverName.

api/v1alpha1/valkeynode_types.go
spotu-dev: Added. MinLength=1 + DNS-1123 CEL on both ValkeyCluster.spec.networking.tls.serverName and ValkeyNode.spec.tls.serverName.

bjosv:
> The exporter env var rolls every TLS cluster.

It seems like this fixed a bug since redis_exporter now gets the commonly used servername, even when not changed? To me that warrants a roll at this stage. We need to make sure to mention it in the release notes if agreed.

Regarding the failing DCO job:
There is a mismatch in the last commit, author email (twitter.com) vs. sign-off (x.ai), which our DCO job doesn't accept.
If you can sign-off the last commit then maybe you can amend it and force push?

spotu-dev:
Thanks for the patience on this. 

@melancholictheory answering the three points in order.

1. Exporter env rolls every TLS+exporter cluster
Same view as @bjosv — this is a real redis_exporter verify-name bug (it was checking `localhost`), so the roll is warranted. Sidecar still dials `localhost`; we only set `REDIS_EXPORTER_TLS_SERVER_NAME`. Called that out in the PR Features section for the next release notes.

2. CEL `dns1123Subdomain` rejects a trailing-dot FQDN
Announce / DNS ([RFC 1034](https://www.rfc-editor.org/rfc/rfc1034), [RFC 1035](https://www.rfc-editor.org/rfc/rfc1035)) may use an absolute FQDN with a trailing dot. TLS SNI ([RFC 6066 §3](https://www.rfc-editor.org/rfc/rfc6066#section-3)) and cert SAN `dNSName` must not. `serverName` on this PR is the verify/SNI name, so the DNS-1123 CEL asked for is the right check here. I don’t think we should relax it here. The follow-up should be on [#378](https://github.com/valkey-io/valkey-operator/issues/378): that work builds the announce FQDN and should strip the dot when it feeds the same string into TLS ServerName. I can create a draft on that.

3. Empty `REDIS_EXPORTER_TLS_SERVER_NAME` on a standalone ValkeyNode
Done in [`57c747b`](https://github.com/valkey-io/valkey-operator/pull/409/commits/57c747b). We only set the env when `spec.tls.serverName` is non-empty (same field the node client uses). A hand-applied node with TLS and no `serverName` no longer gets a blank var. Cluster nodes are unchanged — the controller still stamps a FQDN, so they still get the env.

@bjosv DCO is clean now — author, committer, and Signed-off-by are all `spotu@x.ai`.

New commit re-queued fork Actions, can someone approve and run please?

melancholictheory:
You are right on 2, and I put the fix in the wrong place. RFC 6066 section 3 says the SNI hostname is carried "without a trailing dot", so a field that is the SNI name should reject one. Nothing to relax here.

Worth making the follow-up sharper than I framed it, though, because it is not only that a user cannot type the absolute form.

In #378, `headlessServiceFQDN` returns the trailing-dot name and its own comment says it feeds TLS ServerName and Hostname announce. Its test pins `"valkey-c.default.svc.cluster.local."`, and it is used as `serverName` in both the cluster-state path and the node client path.

This PR stamps the resolved default onto `ValkeyNode.spec.tls.serverName` through `nodeTLSFromCluster`, and `NodeTLSSpec.ServerName` carries the same DNS-1123 rule. So on the merge order #409 then #378, the operator computes a trailing-dot default and writes it to a field whose CEL rejects it. That is the operator failing to write its own node spec, not a user hitting a validation message, which is a louder failure and worth catching in the draft you offered.

The split you describe is the right one: keep the dot for announce, strip it where the same string becomes ServerName.

3 checks out in `57c747b`, the env is now behind `if tlsSpec.ServerName != ""`. Nothing to add on 1.

spotu-dev:
Thanks !!  I'll strip the trialling dot as a follow up in the #378. If there's nothing more on this PR, One of @melancholictheory @bjosv Can I get a 🚢  please ?

sandeepkunusoth (approved):

spotu-dev:
@sandeepkunusoth thanks for the re-approve. I don't have merge rights, could you merge?

bjosv:
Thanks for your work!

Issue #403, opened 2026-08-26
[BUG]: ValkeyNode with unparseable topology labels is never cleaned up and wedges the cluster in Reconciling


jdheyburn:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.
- [x] I am running a recent version of the operator (latest release or `main`) and the issue still reproduces.

### Bug Description

A ValkeyNode that carries a cluster's `valkey.io/cluster` label but has an unparseable `valkey.io/shard-index` (or `valkey.io/node-index`) label is never cleaned up, and its presence wedges the owning ValkeyCluster in `Reconciling` indefinitely. The cluster only recovers when the node is deleted by hand.

Two pieces of code interact:

- The cluster controller lists its nodes by the `valkey.io/cluster` label alone, so the malformed node is included in every reconcile.
- `deleteExcessValkeyNodes` calls `strconv.Atoi` on the topology labels and `continue`s on error, so the node is never considered excess and never deleted.

The scale-in path does count it as an extra node, so every reconcile requeues with `Ready=False/Reconciling "Scaling in cluster"` / `Progressing=RebalancingSlots` and the healthy path is never reached again.

Found while validating #400 (prompted by the Greptile comment there). The false `Degraded/ACLApplyFailed` that comment predicts does **not** reproduce — the node controller rewrites the condition to `Applied` within seconds, and the wedge described here keeps the cluster off the healthy path where the aggregation runs. The wedge itself is pre-existing: none of the implicated paths (`deleteExcessValkeyNodes`, `handleScaleIn`, the node listing) are touched by #400.

### Steps to Reproduce

1. Create a 1-shard, 0-replica cluster and wait for `Ready/ClusterHealthy`.
2. Create a ValkeyNode labelled for the cluster but with a malformed shard index:

   ```yaml
   apiVersion: valkey.io/v1alpha1
   kind: ValkeyNode
   metadata:
     name: aclfail-bogus
     namespace: default
     labels:
       valkey.io/cluster: aclfail
       valkey.io/node-index: "99"
       valkey.io/shard-index: not-an-integer
   spec:
     workloadType: StatefulSet
   ```

3. Watch the ValkeyCluster status.

### Expected Behaviour

Either the operator deletes the out-of-topology node (it is labelled as owned by the cluster and is not part of the desired topology), or it excludes it from reconciliation and reports something actionable. In no case should the cluster stay in `Reconciling` forever with no event or condition naming the offending node.

[...]

PR #415, opened 2026-08-30
(feat) allow custom serviceAccountName on valkey cluster pods

### LINKED ISSUE #414: [FEATURE]: Allow setting serviceAccountName on ValkeyCluster pods
### Pre-flight checks

### Problem / Use case

Pods created for a ValkeyCluster run as the namespace `default` ServiceAccount, and there is no way to change that from the CR.

`ValkeyClusterSpec` has no `serviceAccountName` or `serviceAccount` field, and neither does `ValkeyNodeSpec`. I checked `api/v1alpha1/valkeycluster_types.go` on `main` as well as v0.5.0. The `containers` strategic merge patch does not reach it either, since `serviceAccountName` sits on the pod spec rather than on a container.

Setting it on the generated StatefulSet does not survive:

```
$ kubectl -n valkey patch sts <name> --type merge \
    -p '{"spec":{"template":{"spec":{"serviceAccountName":"valkey"}}}}'
statefulset.apps/<name> patched

$ kubectl -n valkey get sts <name> -o jsonpath='{.spec.template.spec.serviceAccountName}'
valkey

# about 20 seconds later
$ kubectl -n valkey get sts <name> -o jsonpath='{.spec.template.spec.serviceAccountName}'
(empty)
```

The ValkeyNode controller rewrites the template, so it is not settable on the generated object either.

Two reasons this matters:

A pod on `default` inherits whatever any RoleBinding grants that ServiceAccount in the namespace, now or later. It is a permission surface nobody thinks to look at, because nothing about the workload mentions it.

It also blocks a common cluster baseline. Where a rule forbids pods from using the default ServiceAccount, every ValkeyCluster pod and every StatefulSet the operator creates is flagged, with no way to comply. The only option left is a policy exemption, which is a worse answer than a field.

Part of the exposure has a workaround. Setting `automountServiceAccountToken: false` on the namespace's `default` ServiceAccount stops the API token from being mounted, and that works precisely because it is a field of the ServiceAccount rather than of the pod. The identity itself has no workaround.

### Proposed solution

Add an optional `serviceAccountName` to `ValkeyClusterSpec`, carried through to `ValkeyNodeSpec` and onto the pod template of the generated StatefulSet. Empty by default, so existing clusters keep the behaviour they have today.

```yaml
kind: ValkeyCluster
metadata:
  name: example
spec:
  shards: 3
  replicas: 1
  serviceAccountName: valkey
```

The CR already accepts pod level settings through `podSecurityContext`, so this follows a shape that is already there rather than introducing a new one. Creating the ServiceAccount stays the user's job; the operator would only need to reference it.

### Scope of change

### API / behaviour impact

### Contribution

- [ ] I would like to implement this feature myself.

### Additional context

Running valkey-operator v0.5.0 (`ghcr.io/valkey-io/valkey-operator:v0.5.0`), chart `valkey-operator` 0.5.0, on Kubernetes 1.36. A 3 shard cluster with one replica each, so six StatefulSets and six pods, all on `default`.

Closest existing issue I found is #191, about permissions on Talos, which is closed and covers different ground.

sandeepkunusoth:
This PR closes https://github.com/valkey-io/valkey-operator/issues/414

Adds support for configuring the Kubernetes `ServiceAccount` used by ValkeyCluster pods through `spec.serviceAccountName`.

## Motivation

ValkeyCluster pods currently use the namespace's `default` ServiceAccount, and there is no way to configure a different ServiceAccount through the ValkeyCluster CR.

This can be a problem for environments with security policies that prohibit the use of the `default` ServiceAccount or require workloads to run under a dedicated identity.

## Changes

- Added optional `serviceAccountName` to `ValkeyClusterSpec`.
- Added optional `serviceAccountName` to `ValkeyNodeSpec`.
- Propagated the value from `ValkeyCluster` to generated `ValkeyNode` resources.
- Applied `serviceAccountName` to the generated StatefulSet pod template.
- Updated generated CRDs.
- Added unit tests covering ServiceAccount propagation and StatefulSet rendering.
- Updated documentation with a usage example.

Existing clusters remain unchanged when `serviceAccountName` is not specified.

### Checklist


melancholictheory (commented):
The shape holds up where it matters. `ServiceAccountName` goes through `buildValkeyNodePodTemplateSpec`, which both the StatefulSet and the Deployment builders call, so neither workload type is left out. And the `!= ""` guard does what the issue asked for: I hashed the rendered template on this branch and on `main` with no `serviceAccountName` set and got the same value, so nobody rolls on the upgrade that ships this. The assertion message on the empty case says exactly that, which is nice to see written down.

Two things worth changing.

The pattern is stricter than Kubernetes and disagrees with its own length bound. `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$` is a DNS-1123 label, which forbids dots and caps at 63 anyway, while `MaxLength=253` is the subdomain bound. Kubernetes validates ServiceAccount names with `NameIsDNSSubdomain`, so dots are legal. Both ends check out:

```
$ kubectl create sa my.sa --dry-run=server
serviceaccount/my.sa

spec.serviceAccountName: Invalid value: "my.sa": should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```

So a ServiceAccount the cluster will happily create cannot be named here.

A ServiceAccount that does not exist is invisible on the ValkeyCluster. The docs are right that pod creation fails at admission, but it fails quietly from the CR's point of view: no pod is created, the StatefulSet sits at `replicas=0`, and the only signal is an event on the workload.

```
Warning  FailedCreate  statefulset/...  pods "..." is forbidden: error looking up
service account default/does-not-exist: serviceaccount "does-not-exist" not found
```

A typo therefore reads as a cluster that just never progresses. Given `ConfigurationWarning` already exists for accepted-but-risky input, this looks like it belongs there, and it is the same shape as #403.

Minor: both new tests are named `..._ContainerSecurityContext_Passthrough` and `..._ContainerSecurityContext_EmptyIsNoop` but test `ServiceAccountName`.

internal/controller/valkeycluster_controller.go
```diff
@@ -874,6 +876,24 @@ func nodeTLSFromCluster(cluster *valkeyiov1alpha1.ValkeyCluster) *valkeyiov1alph
 	}
 }
 
+// serviceAccountConfigWarnings returns a ConfigurationWarning if the
+// cluster's explicitly requested ServiceAccount does not exist.
+func (r *ValkeyClusterReconciler) serviceAccountConfigWarnings(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) []configWarning {
+	if sa := cluster.Spec.ServiceAccountName; sa != "" {
+		saObj := &corev1.ServiceAccount{}
+		if err := r.Get(ctx, client.ObjectKey{Namespace: cluster.Namespace, Name: sa}, saObj); err != nil {
```
jdheyburn, line 884: `serviceAccountConfigWarnings` reads the ServiceAccount through the cached client, which starts a cluster-wide ServiceAccount informer. The cache in `cmd/main.go` is label-filtered to operator-managed objects on purpose, and ServiceAccounts are user-created so they can't be filtered that way. The informer is also why this needs `list`/`watch` for what is a single `get`, and if those verbs are ever missing (e.g. a Helm ClusterRole that hasn't been synced) the `Get` blocks waiting for cache sync and, with one worker, every ValkeyCluster stops reconciling. I reproduced that locally.

The simplest fix is to bypass the cache for ServiceAccounts at the manager, in cmd/main.go:

```go
   Client: client.Options{
       Cache: &client.CacheOptions{
           DisableFor: []client.Object{&corev1.ServiceAccount{}},
       },
   },
```

Your r.Get then goes straight to the API server as written, and the RBAC marker can go back to `verbs=get`.

internal/controller/valkeycluster_controller.go
```diff
@@ -874,6 +876,24 @@ func nodeTLSFromCluster(cluster *valkeyiov1alpha1.ValkeyCluster) *valkeyiov1alph
...
+			if apierrors.IsNotFound(err) {
+				msg := fmt.Sprintf("ServiceAccount %q does not exist; Pods will fail to create until it is created", sa)
+				return []configWarning{{
+					reason:  valkeyiov1alpha1.ReasonServiceAccountNotFound,
+					message: msg,
+				}}
+			}
```
jdheyburn, line 891: Do we want to log errors other than not found instead of silently dropping them? Any transient API error would clear `ConfigurationWarning` for that reconcile, then the next one would re-emit the Warning with a reset `LastTransitionTime`. Something like this maybe?

```go
// serviceAccountConfigWarnings reports when the cluster's requested
// ServiceAccount cannot be confirmed to exist. The lookup is advisory:
// pod admission is what actually enforces it.
func (r *ValkeyClusterReconciler) serviceAccountConfigWarnings(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) []configWarning {
      sa := cluster.Spec.ServiceAccountName
      if sa == "" {
              return nil
      }
      err := r.Get(ctx, client.ObjectKey{Namespace: cluster.Namespace, Name: sa}, &corev1.ServiceAccount{})
      switch {
      case err == nil:
              return nil
      case apierrors.IsNotFound(err):
              return []configWarning{{
                      reason:  valkeyiov1alpha1.ReasonServiceAccountNotFound,
                      message: fmt.Sprintf("ServiceAccount %q does not exist; Pods will fail to create until it is created", sa),
              }}
      default:
              logf.FromContext(ctx).Error(err, "could not verify ServiceAccount", "serviceAccount", sa)
              return []configWarning{{
                      reason:  valkeyiov1alpha1.ReasonServiceAccountLookupFailed,
                      message: fmt.Sprintf("could not verify ServiceAccount %q exists: %v", sa, err),
              }}
      }
}
```

Then ensure you create a `ReasonServiceAccountLookupFailed = "ServiceAccountLookupFailed"` constant and update it under `ConfigurationWarning` in the docs.

docs/valkeycluster.md
```diff
@@ -77,6 +77,17 @@ containers:
...
+### Service account
+
+```yaml
+serviceAccountName: valkey-sa
+```
+
+`serviceAccountName` sets the Kubernetes ServiceAccount used by each ValkeyNode
+pod. If unset, the pod uses the namespace's default ServiceAccount.
+
+The referenced ServiceAccount **must already exist in the same namespace** as the ValkeyCluster, The operator only assigns the name to each pod template; it does not create, modify, or validate RBAC for that ServiceAccount. If the name does not exist, pod creation fails at admission.
```
jdheyburn, line 89: typo: change the first comma for a period. "as the ValkeyCluster. The operator"

docs/valkeycluster.md
```diff
@@ -77,6 +77,17 @@ containers:
...
 `containers` patches the pod's container list using strategic merge patch. Containers named `server` or `metrics-exporter` are merged by name; anything else is appended as a sidecar.
 
+### Service account
+
+```yaml
+serviceAccountName: valkey-sa
+```
+
+`serviceAccountName` sets the Kubernetes ServiceAccount used by each ValkeyNode
+pod. If unset, the pod uses the namespace's default ServiceAccount.
```
jdheyburn, line 87: Perhaps mention that changing this causes every pod to roll.

jdheyburn (commented):
Thanks for raising!

1. Could you add in the PR description that ClusterRole in valkey-helm needs `serviceaccounts` too?
2. Do we want to stick to podSpec equal fields `serviceAccountName`, or perhaps expose it such as:

```yaml
serviceAccount:
  name: myname
```

docs/valkeycluster.md
bjosv, line 89: Maybe we can give some more info like:
```suggestion
The referenced ServiceAccount **must already exist in the same namespace** as the ValkeyCluster, The operator only assigns the name to each pod template; it does not create, modify, or validate RBAC for that ServiceAccount. If the name does not exist, the workload is created but pod admission rejects every pod and the cluster never becomes ready. The operator reports this as a `ConfigurationWarning` condition with reason `ServiceAccountNotFound`.
```

melancholictheory (commented):
Re-review of 7ea9575. The three points from the first pass are closed: the pattern is the DNS-1123 subdomain one and agrees with the 253 bound, a missing ServiceAccount now surfaces as `ConfigurationWarning` with `ServiceAccountNotFound` and has an envtest case, and the tests are named for what they test.

On the lookup I would go one step further than the `DisableFor` fix suggested above. The reconciler already carries an uncached `APIReader` (`mgr.GetAPIReader()`, used for the TLS Secret, user passwords and the ACL Secret for the same reason), so `r.APIReader.Get(...)` in `serviceAccountConfigWarnings` gives the same result with no informer and the RBAC marker back to `verbs=get`, and without touching `cmd/main.go`, which #447 also changes. The envtest case already sets `APIReader: k8sClient`, so it keeps passing as is.

With that, what stands between this and an approval from my side is the rest of jdheyburn's list: logging non-NotFound errors under their own reason, the rolls note and the typo in the docs, and the chart ClusterRole needing `serviceaccounts` get once this lands (I have valkey-helm#251 open against that file and can fold the verb in there). On the shape question, `serviceAccountName` as the PodSpec field name is what I would keep. A `serviceAccount` block makes sense once there is a second field to put in it.

PR #421, opened 2026-09-02
fix: strip trailing-dot announce FQDN when used as TLS SNI

spotu-dev:
Follow-up after #409 and #378.

`headlessServiceFQDN` is the absolute announce name (`….svc.cluster.local.`).
#378 already strips that dot in `tlsServerName` before stamping
`ValkeyNode.spec.tls.serverName`. This PR makes the same rule the last
step before the TLS handshake:

- `sniHostname` implements RFC 6066 §3 (`HostName` is ASCII without a trailing dot).
- `getTLSConfig` always applies it, so a dotted name never goes out as SNI
  even if a caller passes `headlessServiceFQDN` or a stamped field with a dot.
- Announce / `CLUSTER SLOTS` is unchanged (still the trailing-dot form).
- Test pins announce vs SNI as the same string ± the final `.`.

No CEL change. `serverName` stays DNS-1123.

### Testing

```
go test ./internal/controller/ -run 'TestTLSServerName|TestSNIHostname|TestNodeTLSFromCluster|TestHeadlessServiceFQDN'
```


spotu-dev:
Closing: not a live bug. #378 already strips the trailing dot in tlsServerName before stamp/dial; CEL rejects a dotted user serverName; node-to-node does not name-check. getTLSConfig TrimSuffix would be a no-op (Go hostnameInSNI also strips).

Issue #423, opened 2026-09-03
[ISSUE] Guard unknown enum config values on older Valkey images


sandeepkunusoth:
## Problem

Version-gating configuration directives is not sufficient to prevent all
`FATAL CONFIG FILE ERROR` failures.

A configuration directive may already exist in an older Valkey version, while
newer Valkey versions can add additional valid values for that directive.

For example:

```yaml
spec:
  image: valkey/valkey:<older-version>
  config:
    <existing-config>: <value-introduced-in-newer-version>
```

The older Valkey server knows the configuration directive, but does not know
the newer value.

When the operator renders this configuration into valkey.conf, Valkey can
fail during startup with above error.

## Current behavior

The version-gating framework introduced in #307 handles configuration
directives that have been explicitly registered with a minimum supported
Valkey version.

It does not currently provide version-aware validation for individual enum
values.

Unsupported directive

tls-auto-reload-interval

The directive itself does not exist in the older Valkey version.

Unsupported enum value

<known-directive> <value-introduced-in-newer-version>

The directive exists, but the value is not recognized by the older
Valkey version.

The second case is not addressed by directive-level gating alone.

### Option 1: Extend the existing version-gating framework
Extend the existing configuration-gating mechanism so that it can represent
both:

the minimum Valkey version for a configuration directive
the minimum Valkey version for individual enum values

```yaml
configuration directive
        |
        +-- minimum version
        |
        +-- enum values
              |
              +-- value A -> minimum version
              +-- value B -> minimum version
              +-- value C -> minimum version
```

### Option 2: Investigate whether Valkey could expose configuration metadata through an API or command

There can be other options we can explore.

melancholictheory:
This class is not hypothetical, and there is already an instance of it inside the operator rather than in `spec.config`.

`shutdown-on-sigterm` exists in both 8.1 and 9.0, but the value set grew:

```c
// 8.1.0 src/config.c
configEnum shutdown_on_sig_enum[] = {
    {"default", 0}, {"save", ...}, {"nosave", ...}, {"now", ...}, {"force", ...}, {NULL, 0}};

// 9.0.0 src/config.c
configEnum shutdown_on_sig_enum[] = {
    {"default", ...}, {"save", ...}, {"nosave", ...}, {"now", ...}, {"force", ...},
    {"safe", SHUTDOWN_SAFE}, {"failover", SHUTDOWN_FAILOVER}, {NULL, 0}};
```

[...]

PR #427, opened 2026-09-04
feat: generate versionGatedConfig from Valkey release tags

bjosv:
### Summary

The version-gated config map (added in #307) was maintained by hand, which can fall behind as new Valkey versions add directives. Deriving it from the Valkey sources keeps it complete and current.

Add `hack/gen_config_version_metadata.py`, which scans `src/config.c` across Valkey release tags and emits `internal/valkey/config_version_metadata.go` (directive -> minimum Valkey version). For each Valkey minor it reads one release: the final X.Y.0 if released, otherwise the newest release candidate (so users testing an rc image still get gating); once the final ships it takes over and its versions replace the rc ones.

Patch releases are skipped so a directive is attributed to the minor that introduced it, not to an older patch that might have backported it. HIDDEN_CONFIG directives are also skipped: they are internal, absent from CONFIG GET, and not a supported user-facing setting.

The script formats its output with gofmt and offers --check to verify the committed file is current. Remove the hand-written map from config.go and document regeneration in the developer guide.

### Other

- The script can also be improved to get enum-values and create the structure needed for #423 (PoC:ed).
- Non-`IMMUTABLE_CONFIG` configs can also be listed for `liveConfigAllowlist` but we probably want to hand-pick what configs to include in this list.

### Testing

Tested locally on Valkey unstable with the soon tagged 9.2.0-rc1, which will result in following additions (but sorted):
```
+       "bgsave-default-method":                  semver.MustParse("9.2.0-rc1"),
+       "cluster-replica-priority":               semver.MustParse("9.2.0-rc1"),
+       "forkless-infrastructure-enabled":        semver.MustParse("9.2.0-rc1"),
+       "hotkeys-sampling-percentage":            semver.MustParse("9.2.0-rc1"),
+       "hotkeys-top-k":                          semver.MustParse("9.2.0-rc1"),
+       "hotkeys-window-seconds":                 semver.MustParse("9.2.0-rc1"),
+       "repl-throttling-enabled":                semver.MustParse("9.2.0-rc1"),
```

### Checklist


internal/valkey/config_version_metadata.go
```diff
@@ -0,0 +1,28 @@
...
+// versionGatedConfig maps user-facing config directives to the minimum
+// Valkey version that understands them. Only directives introduced after
+// the 8.1.0 baseline are listed; anything not present is assumed
+// supported by every Valkey version the operator supports.
+var versionGatedConfig = map[string]*semver.Version{
+	"cluster-announce-client-port":           semver.MustParse("9.0.0"),
+	"cluster-announce-client-tls-port":       semver.MustParse("9.0.0"),
+	"cluster-config-save-behavior":           semver.MustParse("9.1.0"),
+	"cluster-databases":                      semver.MustParse("9.0.0"),
+	"cluster-slot-migration-log-max-len":     semver.MustParse("9.0.0"),
```
sandeepkunusoth, line 21: LGTM. can we move this auto-generated file to internal/valkey/ folder? or should we add it as part of config/ folder as its auto generated?

bjosv: Ah good, yes it fits better under internal/valkey/, config/ is more for yamls and kustomization.
I'll update and do some renaming since it becomes generic config information, lets see if you agree with the names.

hack/gen_version_gated_config.py
```diff
@@ -0,0 +1,407 @@
...
+            return 2
+        try:
+            with open(args.out, "r", encoding="utf-8") as f:
+                existing = f.read()
+        except FileNotFoundError:
+            print(f"error: --check: {args.out} does not exist", file=sys.stderr)
+            return 1
+        if existing != output:
+            regen = (
+                f"hack/gen_version_gated_config.py --valkey-repo {args.valkey_repo} "
```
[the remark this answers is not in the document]
bjosv: We could add it to a weekly CI in the future, but adding it to current CI will trigger failures unrelated to the PR

melancholictheory (commented):
Checked the generator by rebuilding its answer independently rather than reading the script, and the output holds up. I pulled the directive list out of `src/config.c` at the 8.1.0, 9.0.0 and 9.1.0 tags and diffed the minors. Six directives appear in that ground truth and not in the generated map:

```
cluster-message-gossip-perc, io-threads-always-active, lua-enable-insecure-api,
min-io-threads-avoid-copy-reply, min-string-size-avoid-copy-reply,
min-string-size-avoid-copy-reply-threaded
```

All six carry `HIDDEN_CONFIG`, so that is the documented exclusion doing its job, not a gap. No false negatives anywhere else, which is the failure mode that would matter, since a directive missing from the map passes the gate and crash-loops the pod.

One thing does regress, though.

`tls-auto-reload-interval` was moved to `9.1.0-rc1` in #307 after we found that release candidates were being flattened to GA. The generated map puts it back to `9.1.0`, and the effect is visible through the operator's own comparison:

```
generated min = 9.1.0
  valkey/valkey:9.1.0-rc1    meets(9.1.0)=false   meets(9.1.0-rc1)=true
  valkey/valkey:9.1.0-rc2    meets(9.1.0)=false   meets(9.1.0-rc1)=true
  valkey/valkey:9.1.0        meets(9.1.0)=true
```

So an image on `9.1.0-rc1` now loses a directive its build supports, silently, with a `ConfigurationWarning` explaining that the version is too old.

That works against the stated goal rather than with it. The rc fallback exists "so users testing an rc image still get gating", but "once the final ships it takes over and its versions replace the rc ones" breaks exactly those users the moment the minor lands. Someone testing an unreleased minor is served correctly; someone testing a released one is not, which reads like a side effect of the rule rather than a decision.

Attributing each directive to the earliest release that contains it, release candidates included, fixes it and keeps everything else the same: `9.1.0-rc1` as the minimum covers the rcs and GA alike, which is what the hand-written entry did.

Smaller note: the generated header says the baseline is 8.1.0 while the operator's stated baseline is 9.0 (#263), so the 9.0.0 rows can never gate anything on a supported image. Harmless as margin, worth being deliberate about rather than incidental.

bjosv:
> `tls-auto-reload-interval` was moved to `9.1.0-rc1` in #307 after we found that release candidates were being flattened to GA.  ...

The idea was to enabling updating both DefaultVersion and this config meta data at the same time using a dependabot/CI job. The job will run this script to get the latest release+config metadata and prepare a PR with the collected updates. 

Release candidates are transient by construction. so the idea was that the committed map never carries an 'rc' for a released minor. At a Valkey GA we would flip the 'rc' and update the DefaultVersion at the same time.

melancholictheory:
Fair. With 9.1.0 GA out, `9.1.0` is the right output under that rule, so "reverts #307" was the wrong framing on my side. A job that moves `DefaultVersion` and the metadata together is a cleaner invariant than hand-maintaining rc entries anyway.

The rc gap survives the rule change, though, and it is not really about the map. I ran `MeetsMinVersion` against current main:

```
min = 9.1.0
  valkey/valkey:9.0.6      false
  valkey/valkey:9.1.0-rc1  false
  valkey/valkey:9.1.0-rc2  false
  valkey/valkey:9.1.0      true
  valkey/valkey:9.2.0-rc1  true
```

A prerelease of a later release passes the gate. A prerelease of the same release does not. That is semver prerelease ordering arriving through `!version.LessThan(min)`, and it holds whatever the generator emits.

Which means the rc question can be settled without touching the generator's policy at all. If the comparison drops the prerelease from the image version before comparing, a GA-attributed map behaves the same as an rc-attributed one on every image I tried:

```
min = 9.1.0        current   core-compare
  9.0.6              false      false
  9.1.0-rc1          false      true
  9.1.0-rc2          false      true
  9.1.0              true       true
  9.1.2              true       true
```

The committed map then never needs an rc, which is what you want, and an rc image still gets gated correctly.

Worth seeing what this looks like from the other side once the map says `9.1.0`: the directive is dropped from valkey.conf and the condition reads `spec.config.tls-auto-reload-interval requires Valkey 9.1.0+, detected 9.1.0-rc1 from spec.image`. If your build does have the directive, that is a hard message to act on.

None of this is this PR's job. Happy to bring it separately if the core comparison looks right to you.

sandeepkunusoth:
> spec.config.tls-auto-reload-interval requires Valkey 9.1.0+, detected 9.1.0-rc1 from spec.image

wont we get this warning only when user explicitly pass spec.image as 9.1.0-rc1 in valkey CR spec? i think that fine.

melancholictheory:
Right, and the warning text is the least of it. Dropping that part.

The piece underneath is not about whether rc images deserve support. It is that the comparison already supports them unevenly. The map this PR generates has two distinct minimums, so `MeetsMinVersion` answers like this:

```
image        min=9.0.0   min=9.1.0
  9.0.0-rc1    false       false
  9.0.0        true        false
  9.0.6        true        false
  9.1.0-rc1    true        false
  9.1.0        true        true
  9.2.0-rc1    true        true
```

A prerelease is refused only when it is a prerelease of the minimum itself. Every other prerelease passes. So one `9.1.0-rc1` cluster is given the eight `9.0.0` directives and refused the three `9.1.0` ones, out of the same comparison.

That is semver ordering showing through rather than a decision about rc support, and with this PR it covers 11 directives instead of the one hand-written entry. Still nothing this PR has to carry.

sandeepkunusoth (approved):

Issue #408, opened 2026-08-27
[BUG]: a node whose pod never becomes Ready is not recovered by correcting the spec


melancholictheory:
### Pre-flight checks

- [x] I have searched existing issues and discussions and could not find a duplicate.

### Problem

When a `ValkeyNode`'s pod cannot become Ready, correcting the `ValkeyCluster` spec does not recover it. The StatefulSet template updates, the pod is never replaced, and the cluster sits in `Reconciling` indefinitely. Only deleting the pod by hand gets it back.

The cause is that the node StatefulSet is created with `PodManagementPolicy: OrderedReady` (`valkeynode_resources.go:597`). The StatefulSet controller does no update work until the existing pod is Running and Ready, so the pod that needs replacing is the one blocking its own replacement. Nothing in the operator deletes a pod in that state.

### Steps to reproduce

Verified on k3d, 3 shards with 1 replica each, operator built from `main`.

1. Set `spec.image` to `valkey/valkey:8.1.1`, below the 9.0 baseline that #263 states.
2. The replica rolls first and its pod fails to start:

```
*** FATAL CONFIG FILE ERROR (Version 8.1.1) ***
Reading the configuration file, at line 10
>>> 'shutdown-on-sigterm failover'
argument(s) must be one of the following: default, save, nosave, now, force
```

That part is expected. `shutdown-on-sigterm` requires 9.0+ and the operator writes it unconditionally, which is correct given the 9.0 baseline.

3. Revert `spec.image` to `valkey/valkey:9.0.0`.

The `ValkeyNode` spec updates and so does the StatefulSet template:

```
STS template image:  valkey/valkey:9.0.0
STS revisions:       current=1-67c4795f68  update=1-67c4795f68
STS status:          replicas=1 ready=  updated=
pod revision-hash:   1-5c5d7ff747          <- still the 8.1.1 revision
pod image:           valkey/valkey:8.1.1
```

The node controller logs the same line on every pass, because from its point of view the work is done:

```
StatefulSet already matches desired (no pod template roll)  name=valkey-rolltest-0-1
ValkeyNode not ready, requeuing                             name=rolltest-0-1
```

4. Observed for about seven minutes with no change. `kubectl delete pod` recovered the cluster to Ready in roughly 30 seconds.

### Scope

[...]

PR #431, opened 2026-09-07
fix: refresh topology status while a node roll is stuck

daanvinken:
When `reconcileValkeyNodes` requeues (`UpdatingNodes`), the controller only set Ready and Progressing. `updateStatus` skipped `readyShards` because Valkey cluster state was nil. `ClusterFormed` stayed True with the previous `observedGeneration` while a shard had no running pod.

On that path the controller now lists pods (no Valkey clients) and:

- sets `status.readyShards` to the number of shards that have at least one Ready Pod
- sets `ClusterFormed` and `SlotsAssigned` False on the current generation if a shard has no Ready Pod

A shard still counts if a replica Pod is Ready and node-index 0 is not (live primary after failover). This does not implement primary failback (#311).

Related: #408

## Testing

`go test ./internal/controller/` covering `UpdatingNodes topology status` and `TestCountShardsWithReadyPod`.

### DIFF
--- a/docs/status-conditions.md
+++ b/docs/status-conditions.md
@@ -19,7 +19,7 @@ These top-level fields in `.status` provide a high-level, human-readable summary
 
 - **`shards`**: The number of shards currently detected in the Valkey cluster.
 
-- **`readyShards`**: The number of shards that are fully healthy, meaning they have a primary and the desired number of replicas. When Valkey cluster state is not available (for example during `UpdatingNodes`), this counts shards that have a Ready primary Pod (node-index 0).
+- **`readyShards`**: The number of shards that are fully healthy, meaning they have a primary and the desired number of replicas. When Valkey cluster state is not available (for example during `UpdatingNodes`), this counts shards that have at least one Ready Pod.
 
 ---
 
@@ -105,13 +105,13 @@ Indicates whether all nodes have joined the cluster and meet the desired shard/r
 | Status | Meaning |
 |---|---|
 | `True` | All shards and replicas are present and joined. |
-| `False` | Pending nodes, missing shards, missing replicas, or a required primary Pod is not Ready. |
+| `False` | Pending nodes, missing shards, missing replicas, or a shard has no Ready Pod. |
 
 Typical reasons:
 - `ClusterFormed` – Cluster is formed
 - `MissingShards` – waiting for shards
 - `MissingReplicas` – waiting for replicas
-- `UpdatingNodes` – a required primary Pod is not Ready during a ValkeyNode roll
+- `UpdatingNodes` – a shard has no Ready Pod during a ValkeyNode roll
 
 #### `SlotsAssigned`
 Indicates whether all **16384** hash slots are assigned to primaries.
@@ -124,7 +124,7 @@ Indicates whether all **16384** hash slots are assigned to primaries.
 Common reasons:
 - `AllSlotsAssigned` – all 16384 slots are assigned
 - `SlotsUnassigned` – waiting for slots to be assigned
-- `UpdatingNodes` – a required primary Pod is not Ready; slot ownership is not trusted for this generation
+- `UpdatingNodes` – a shard has no Ready Pod; slot ownership is not trusted for this generation
 
 #### `ConfigurationWarning`
 Indicates the operator accepted a spec value it considers risky, rather than rejecting or overriding it.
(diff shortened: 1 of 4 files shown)


daanvinken:
Rebased onto current main.

@jdheyburn could you take a look?

docs/status-conditions.md
```diff
@@ -19,7 +19,7 @@ These top-level fields in `.status` provide a high-level, human-readable summary
 
 - **`shards`**: The number of shards currently detected in the Valkey cluster.
 
-- **`readyShards`**: The number of shards that are fully healthy, meaning they have a primary and the desired number of replicas.
+- **`readyShards`**: The number of shards that are fully healthy, meaning they have a primary and the desired number of replicas. When Valkey cluster state is not available (for example during `UpdatingNodes`), this counts shards that have at least one Ready Pod.
```
jdheyburn, line 22: I would add "that have at least one Ready Pod that is not being deleted."

internal/controller/valkeycluster_controller.go
```diff
@@ -510,6 +511,69 @@ func podSchedulingIssueForPod(pod *corev1.Pod) *podSchedulingIssue {
...
+		setCondition(cluster, valkeyiov1alpha1.ConditionClusterFormed, valkeyiov1alpha1.ReasonUpdatingNodes, "A shard has no Ready Pod", metav1.ConditionFalse)
+		setCondition(cluster, valkeyiov1alpha1.ConditionSlotsAssigned, valkeyiov1alpha1.ReasonUpdatingNodes, "A shard has no Ready Pod", metav1.ConditionFalse)
+		return
+	}
+	if c := meta.FindStatusCondition(cluster.Status.Conditions, valkeyiov1alpha1.ConditionClusterFormed); c != nil && c.Status == metav1.ConditionTrue {
+		setCondition(cluster, valkeyiov1alpha1.ConditionClusterFormed, c.Reason, c.Message, metav1.ConditionTrue)
+	}
+	if c := meta.FindStatusCondition(cluster.Status.Conditions, valkeyiov1alpha1.ConditionSlotsAssigned); c != nil && c.Status == metav1.ConditionTrue {
+		setCondition(cluster, valkeyiov1alpha1.ConditionSlotsAssigned, c.Reason, c.Message, metav1.ConditionTrue)
+	}
```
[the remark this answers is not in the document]
jdheyburn: @daanvinken I was able to replicate this. With `replicas: 1` and the shard 0 replica stuck on a bad image tag, I get `ClusterFormed=True/TopologyComplete "All nodes joined cluster"` at the new generation while that pod has no running container.

We should preserve the previous generation by deleting the two if blocks at lines 570-575 and let the function end after the ready < spec.shards branch. The state path re-evaluates both conditions once the roll completes anyway.

internal/controller/valkeycluster_controller.go
```diff
@@ -511,6 +512,69 @@ func podSchedulingIssueForPod(pod *corev1.Pod) *podSchedulingIssue {
...
+func (r *ValkeyClusterReconciler) refreshTopologyStatusFromPods(ctx context.Context, cluster *valkeyiov1alpha1.ValkeyCluster) {
+	log := logf.FromContext(ctx)
+	pods := &corev1.PodList{}
+	if err := r.List(ctx, pods, client.InNamespace(cluster.Namespace), client.MatchingLabels(map[string]string{LabelCluster: cluster.Name})); err != nil {
+		log.Error(err, "failed to list Valkey pods for topology status")
+		return
+	}
+	ready := countShardsWithReadyPod(pods.Items, cluster.Spec.Shards)
+	cluster.Status.ReadyShards = ready
+	if ready < cluster.Spec.Shards {
```
jdheyburn, line 565: I found that this branch also runs when scaling out and bootstrapping, as `reconcileValkeyNodes` requeues for any nodes not-ready, and not just rolls.

When I bumped spec.shards from 3 to 4, I see this condition below for the whole time until the new shard is up:

```
ClusterFormed=False/UpdatingNodes  "A shard has no Ready Pod"
SlotsAssigned=False/UpdatingNodes  "A shard has no Ready Pod"
```

When I retried it on main, I got the below, which is more correct as all existing primaries own all the slots throughout.

```
ClusterFormed=False/MissingShards
SlotsAssigned=True/AllSlotsAssigned
```

For the fix, could we consider only shards that already existed - maybe something like `min(spec.shards, status.shards)`, and don't overwrite a condition that's already False with a more specific reason? I'd be interested in seeing how this branch would then handle scale ins too.

Perhaps we could leave `SlotsAssigned` out of this path altogether, since a restarting pod doesn't change who owns slots, so the stale SlotsAssigned=True from the last scrape is more accurate.

You can use this to repo the issue:

```
kubectl patch valkeycluster rollstatus --type merge -p '{"spec":{"shards":4}}'
for i in $(seq 1 45); do sleep 1
  kubectl get valkeycluster rollstatus -o jsonpath='t+'"$i"'s state={.status.state}/{.status.reason} ClusterFormed={.status.conditions[?(@.type=="ClusterFormed")].status}/{.status.conditions[?(@.type=="ClusterFormed")].reason} SlotsAssigned={.status.conditions[?(@.type=="SlotsAssigned")].status}/{.status.conditions[?(@.type=="SlotsAssigned")].reason}{"\n"}'
done | uniq -f1
```

jdheyburn (commented):
Thanks for raising! Left some comments on my findings when testing locally.

PR #432, opened 2026-09-07
fix: recover live-template STS Create after serviceName migrate race (#424)

### LINKED ISSUE #424: fix: recover live-template STS Create after serviceName migrate race
Follow-up from #378.

`orphanAndRecreateStatefulSet` deletes the STS (orphan) and Creates a live-template STS with the new `serviceName`. If Create returns `AlreadyExists`, we requeue. If the next Get is NotFound, the pod is still there, and WorkloadRevision does not match, `refuseDesiredSTSCreate` requeues and never Creates. The pod has no StatefulSet until WorkloadRevision moves.

That refuse path must stay: we must not Create the full desired template and skip WorkloadRevision.

## Todo

- After `AlreadyExists`, Get (uncached). If the name is free, Create the live-template STS already built in that function
- Do not Create the full desired spec on that path
- Unit test: `AlreadyExists`, then STS absent, pod present, stale WorkloadRevision, Create of live-template STS succeeds

## Acceptance

- The pod gets a StatefulSet with the new `serviceName` and the live pod template
- Desired template (Hostname flags, config hash, and so on) still waits on WorkloadRevision

Related: #378

daanvinken:
This PR closes #424

### Summary

After a `serviceName` orphan recreate, Create can return AlreadyExists while delete is still in flight. The next Get is then NotFound. `refuseDesiredSTSCreate` requeues and never Creates. The pod has no StatefulSet.

After AlreadyExists, Get uncached. If the name is free, Create the live-template StatefulSet already built. Do not Create the full desired spec.

The pod gets a StatefulSet with the new `serviceName` and the live pod template. Desired template (Hostname flags, config hash) still waits on WorkloadRevision.

### Implementation

`orphanAndRecreateStatefulSet` sets the owner reference on the live-template object first, then orphan-deletes, then Creates. Owner-ref setup is in-memory. If it fails, the live StatefulSet is still there and the next reconcile retries.

Create no longer requeues immediately on AlreadyExists. It Gets with `APIReader`. If NotFound, it Creates the live-template object already built. If Get finds the new `serviceName`, it returns that object. If Get still has the old `serviceName`, it requeues.

`refuseDesiredSTSCreate` is unchanged.

### Limitations

Next reconcile: a cached Get NotFound still hits refuse and requeues. This PR does not build a live-template from the pod.

### Testing

Unit test with a fake client: first Create returns AlreadyExists, Get returns NotFound, second Create succeeds. Cluster-owned node, pod present, stale WorkloadRevision. The stored StatefulSet has the new `serviceName` and the live pod template, not the desired template.

### Checklist


internal/controller/valkeynode_controller.go
```diff
@@ -512,10 +512,31 @@ func (r *ValkeyNodeReconciler) orphanAndRecreateStatefulSet(
 		return nil, err
```
bjosv: If the SetControllerReference fails we would get into the same problematic loop as when the Create failed before this PR.
Should we r.Delete(..DeletePropagationOrphan..) after SetControllerReference instead?

This is more hardening than a failure-handling since SetControllerReference might not normally fail.

daanvinken: Good catch. `SetControllerReference` only mutates the in-memory object, so it does not need the live StatefulSet.

Owner ref is now set first, then orphan-delete, then Create. If owner-ref setup fails, the old StatefulSet is still there and the next reconcile retries.

daanvinken: Yeah good call, that order should also be flipped to:
1. `SetControllerReference`
2. Delete (orphan)
3. Create

bjosv (approved):
Thanks!

Issue #256, opened 2026-06-17
[bug] Handle quorum loss with persistence enabled (stuck/unschedulable pods)


bjosv:
Not sure this is top priority, but keeping the info about this issue identified when working on PR #244.

Currently, `promoteOrphanedReplicas` skips `TAKEOVER` when persistence is configured, assuming the crashed pod will return with the same node ID and data. This is correct for transient failures (OOMKill, node reboot), but leaves the cluster stuck in degraded state when the primary cannot come back:

- Pod unschedulable (node drained, resource pressure, taint added)
- PVC bound to a failed node (local storage, zone loss)
- PVC manually deleted
- Node permanently removed from the cluster

**Expected behavior:**
After a configurable timeout, or based on pod status checks, the operator should detect that the primary is not coming back and promote a replica via TAKEOVER (accepting potential data loss for the unreplicated writes).

**Alternatives:**

1. Timeout-based: If a primary remains in fail state for longer than a threshold (e.g., spec.failoverTimeout, default 5m), proceed with TAKEOVER even with persistence enabled.
2. Pod-status-based: Check if the primary's pod is in an unrecoverable state (Unschedulable, PVC Lost, node deleted) and trigger TAKEOVER immediately.
3. Annotation escape hatch: Allow manual intervention via valkey.io/force-takeover: "true" on the ValkeyCluster resource.
4. ...

**Risks:**
TAKEOVER trades consensus for liveness. If the "dead" primary is actually partitioned (still serving clients on the other side of a network split), promoting a replica creates two owners for the same slots with competing epochs. With persistence, the old primary could resume serving stale data when the partition heals.

To mitigate this, the operator should:
1. Confirm via the Kubernetes API that the old primary is actually gone (pod deleted, node gone), not just unreachable over the cluster bus.
2. Fence the old primary (delete its pod) before issuing TAKEOVER, preventing it from coming back as a competing primary.

**Context:**

- Discussion in PR244 https://github.com/valkey-io/valkey-operator/pull/244#discussion_r3394795886
- Issue 240: https://github.com/valkey-io/valkey-operator/issues/240#issuecomment-4668331686

melancholictheory:
The gate is carrying two different concerns and they come apart usefully.

[...]

PR #433, opened 2026-09-10
refactor: parse CLUSTER NODES into a struct instead of seven ad-hoc field walks

### LINKED ISSUE #406: refactor: parse CLUSTER NODES into a struct instead of seven ad-hoc field walks
Follow-up from https://github.com/valkey-io/valkey-operator/pull/320#discussion_r3805022717.

CLUSTER NODES output is currently parsed in seven places, each doing its own
`strings.Fields(line)` and indexing by position:

`internal/valkey/clusterstate.go`
- `GetSlots`
- `IsNodeFailed`
- `PrimaryIdFromSelf`
- `FindStaleAddressPeers`
- `GetFailingNodes`
- `getNodeState` (flags)

`internal/controller/valkeynode_controller.go`
- `parseClusterNodesRole`

They agree on the `len(fields) < 8` guard and the flag split at `fields[2]`, then
diverge. Three of them re-state the line layout in a comment.

## Why it's worth doing

A concrete case from #320. `parseClusterNodesRole` treats `len(fields) > 8` as
"owns at least one slot", but the trailing fields can also be migration markers
like `[5461-<-id]`. `parseSlotsRanges` skips those; `GetSlots` returns them
verbatim. Both behaviours are intentional and they happen to agree on the role
they produce, but establishing that took reading three functions across two
packages.

Two reviewers and an automated reviewer stopped at that same line in one review
round, and one of them proposed a change that would have made the two paths
disagree and put the role poller into a trigger loop during resharding. The
positional parsing is what makes that easy to get wrong.

## Proposal

A `ClusterNode` struct in `internal/valkey` with named fields: id, host, port,
flags, primary id, config epoch, link state, owned slot ranges, and
importing/migrating markers held **separately** from owned slots. Plus
`ParseClusterNodes(string) []ClusterNode` and a helper for the `myself` line.

Callers move to reading fields. `internal/valkey` imports nothing internal and
`internal/controller` already imports it, so it needs no new package and creates
no cycle.

## Notes

- Owned slots and in-flight markers must stay separate fields. Conflating them
  is the ambiguity above.
- Behaviour should be preserved exactly. #320 documents why a slot-less master
  resolves to replica, and why a marker-only master resolves to primary; both
  need to survive the refactor with the tests to prove it.
- `internal/controller/valkeynode_acl.go` also does a `strings.Fields` walk, but
  on ACL output, not CLUSTER NODES. Out of scope.

bjosv:
This PR closes #406.

### Summary

`CLUSTER NODES` output was parsed in six places, each doing its own `strings.Fields(line)` and indexing by position. They agreed on the `len(fields) < 8` guard and the flag split at `fields[2]`, then diverged, and three of them restated the line layout in a comment.

Three commits, each reviewable on its own:

1. **`refactor: parse CLUSTER NODES into a ClusterNode struct`** adds `ClusterNode` with named fields and `ParseClusterNodes` as the only code that reads field positions. `GetSlots`, `IsNodeFailed`, `PrimaryIdFromSelf`, `FindStaleAddressPeers`, `GetFailingNodes` and the flag extraction in `getNodeState` now read fields instead of indices.
2. **`refactor: parse CLUSTER NODES once per scrape`** replaces the raw `ClusterNodes` string on `NodeState` with the parsed table, produced once by the scrape. Every accessor previously re-parsed it, and `IsNodeFailed` and `FindStaleAddressPeers` do so inside nested loops over each node.
3. **`refactor: return ClusterNode from GetFailingNodes`** stops fabricating `NodeState` values with only two fields set.

`NodeState.ClusterNodes` (raw string) is gone, replaced by the parsed table.
The output is parsed once by the scrape instead of on every accessor call, and the raw string is no longer retained per node.

One behaviour change, in `forgetStaleNodes`: the address match now requires a non-empty address. A `noaddr` entry has no address, and neither does a ValkeyNode whose pod has no IP yet, so the two used to match and suppress `CLUSTER FORGET` for every `noaddr` entry. They now fall through to the node ID checks, as intended. Covered by `TestHasNodeWithPodIP`.

Otherwise no change to what the operator does to a cluster: the pending-node classification, failure detection and stale-address healing all behave as before, and are covered by tests.

### Implementation

Design notes are in [`docs/clusterstate-design.md`](docs/clusterstate-design.md).

### Limitations

None

### Checklist


melancholictheory (commented):
Checked the behaviour-preservation claim on the part most likely to break quietly. `IsNodeFailed` already matched `fail` and `fail?` before this PR, so `ClusterNode.IsFailing` covering both is a rename, not a change. `make test` is green on 2b67520, and coverage on `internal/valkey` goes from 61.4% to 69.1%.

The unification does pass over two sites, and both of them read flags from the one place that cannot hold the answer:

- `GetSyncedReplicas` (`internal/valkey/clusterstate.go:237`)
- `countReadyShards` (`internal/controller/valkeycluster_controller.go:1531`)

Both spell it `slices.Contains(node.Flags, "fail") || slices.Contains(node.Flags, "pfail")`, and each half fails for its own reason.

`pfail` is not a token Valkey emits. `clusterNodeFlagsTable` in `cluster_legacy.c` holds `myself, master, slave, fail?, fail, handshake, noaddr, nofailover`, and `fail?` is what `CLUSTER_NODE_PFAIL` renders as. That clause is always false.

The `fail` clause is unreachable for a different reason, one this PR makes easier to see at `clusterstate.go:576`: `node.Flags = myself.Flags`. Those are the flags on the node's own line, and a node does not flag itself failing. `clusterCron` skips it when marking PFAIL:

```c
if (node->flags & (CLUSTER_NODE_MYSELF | CLUSTER_NODE_NOADDR | CLUSTER_NODE_HANDSHAKE)) continue;
```

Ran the PR's own parser over a table where a peer is down, rather than trusting the reading:

```
NodeState.Flags = [myself master]
  contains "fail"  = false
  contains "pfail" = false
per-entry flags and IsFailing:
  a1  flags=[myself master]  IsFailing=false
  b2  flags=[master fail?]   IsFailing=true
  c3  flags=[slave]          IsFailing=false
```

The answer is in the table this PR builds, and `IsFailing` reports it correctly. The two callers just never ask.

As for what it costs, an unreachable node never reaches `shard.Nodes` at all: `getNodeState` returns nil and the caller gates on that, so `len(shard.Nodes) < requiredNodes` catches the ordinary outage. The gap is the node the operator can still reach while its peers cannot. `countReadyShards` counts that shard ready, and if the partitioned node is the primary then `IsReplicationInSync` returns true unconditionally for primaries, so nothing downstream stops it either. In `GetSyncedReplicas` the following `master_link_status != "up"` check covers most of the same ground.

`state.IsNodeFailed(node.Id)` is already the right call for `countReadyShards`, which has `state` in hand. `GetSyncedReplicas` hangs off `ShardState` with no route to the full state, so that one is less mechanical.

Changing behaviour may well belong outside a refactor, and leaving both checks alone is defensible. The comment above `GetSyncedReplicas` saying "Nodes with fail/pfail flags are excluded" is harder to leave, since that is the sentence the next reader will believe. Same two sites as #256.

bjosv:
@melancholictheory nice finding, I  created issue #441 so we can address this pre-existing issue in another PR

internal/controller/valkeycluster_controller.go
```diff
@@ -1406,8 +1406,10 @@ func (r *ValkeyClusterReconciler) forgetStaleNodes(ctx context.Context, cluster
 	for _, shard := range state.Shards {
 		for _, node := range shard.Nodes {
 			for _, failing := range node.GetFailingNodes() {
+				// Two empty addresses do not identify a node: such an entry
+				// requires the node ID checks below.
 				idx := slices.IndexFunc(nodes.Items, func(n valkeyiov1alpha1.ValkeyNode) bool {
-					return n.Status.PodIP == failing.Address
+					return n.Status.PodIP != "" && n.Status.PodIP == failing.Host
```
jdheyburn: nit: Since this changes behaviour, could you update the "Features / Behaviour Changes" section in the PR please?

Could you include test cases for this as well too? Some suggested cases:

  - empty host, with one ValkeyNode whose PodIP is "": should return false (the new behaviour)
  - matching IP: true
  - non-matching IP: false

internal/valkey/clusterstate.go
jdheyburn: I think the comment could be updated since we now structure it:

```suggestion
// parseClusterNodesLine splits well-formed migration markers off before
// calling this, so a "[" field here is a malformed marker. Skip it rather
// than fail the whole line.
```

jdheyburn (commented):
Thanks for raising!

`parseClusterNodesRole` is now available on main, I think you already rebased here. Would you be able to include that in scope for this change too?

internal/valkey/clusterstate.go
bjosv: Thanks, updated

internal/controller/valkeycluster_controller.go
bjosv: We need to refactor `forgetStaleNodes` for unittestability, but I extracted this part as the helper-function `hasNodeWithPodIP` and tested it as suggested.

bjosv:
> `parseClusterNodesRole` is now available on main, I think you already rebased here. Would you be able to include that in scope for this change too?

Good call, I updated the branch to include `parseClusterNodesRole` and add the needed changes and test (also fixed top post and comments)

jdheyburn (approved):

jdheyburn:
@bjosv LGTM thank you! After you've rebased main you can merge it in

Issue #407, opened 2026-08-27
chore: define default deployment scope and tune operator resource limits


bjosv:
The operator's resource limits in `config/manager/manager.yaml` are the untouched kubebuilder scaffold defaults, with a `TODO(user)` comment. They are not based on measured profiles:

```yaml
resources:
  limits:
    cpu: 500m
    memory: 128Mi
  requests:
    cpu: 10m
    memory: 64Mi
```

Profiling against v0.5.0 shows the memory limit may be too tight depending on the deployment scope, and the CPU request is too low for active reconciliation. The current 128 Mi limit `OOMKills` during scaling beyond ~20 shards.

### What needs to happen

1. **Define the default deployment scope** — the maximum cluster size (shards, replicas, number of clusters) the operator must handle out of the box without manual resource tuning. This is the baseline all default limits are sized against.

2. **Tune resource limits and requests** to match that scope, based on the profiling data.

3. **Document guidance** for users who exceed the default scope (e.g. "if scaling beyond N shards, raise the operator memory limit to X").

### Profiling summary (v0.5.0)

The binding constraint is **transient allocation during slot rebalancing**, not steady-state pod count.

**Steady state (no scaling)**

| workload | pods | peak RSS | outcome |
|---|---|---|---|
| 40 clusters, 3 shards each | 240 | 83 MB | healthy |
| 100 clusters, 3 shards each | 498 | 124 MB | healthy |

**Scaling (active slot rebalancing)**

| workload | pods | peak RSS | outcome |
|---|---|---|---|
| 1 cluster, 1 to 3 shards | 6 | 60 MB | healthy |
| 1 cluster, 1 to 23 shards | 46 | 138 MB | **OOMKilled** |
| 4 clusters, 1 to 19 shards | 152 | 120 MB | **OOMKilled** at shard 19 |
| 10 clusters, 1 to 16 shards | 320 | 158 MB | **OOMKilled** at shard 16, host CPU saturated |

At the possible default scope (10 clusters × 5 shards = 100 pods), peak RSS was 90 MB with 128 Mi limit — comfortable headroom.

Key observations:
- Fixed baseline is ~60 MB (Go runtime, informer cache). Nearly half the 128 Mi budget is consumed before any workload.
- Holding pods is cheap (~0.075 MB per pod). Rebalancing is expensive: heap hit 109 MB at just 42 pods during slot migration.
[...]

PR #436, opened 2026-09-11
perf: tune valkey-go connections for control-plane use

bjosv:
This PR is part of #407

### Summary

valkey-go sizes each connection for a data-plane client: 0.5 MiB read and write buffers and a 1024-entry command ring, the latter eagerly allocating a mutex, two sync.Cond and a channel per entry. That is about `1.33 MiB` total per connection.

The operator opens a connection per node on every scrape, sends a handful of small control commands, and closes it. In a large cluster those buffers dominated its allocation.

pprof `alloc_space` **before** this PR, single-cluster scale-up:

```
26.02GB 85.18%  github.com/valkey-io/valkey-go.NewClient
9.71GB 31.79%  bufio.NewReaderSize
9.70GB 31.75%  bufio.NewWriterSize
4.06GB 13.28%  valkey-go.newRing
2.43GB  7.95%  sync.NewCond
```

That churn drove the GC target to 133 MB against the 128Mi container limit, which is what OOM killed the operator past ~20 shards.

This PR tunes valkey-go connections for control-plane use.

The following valkey-go options are now used:

```
                       default    new
  ReadBufferEachConn   512 KiB   16 KiB  CLUSTER NODES output grows with node count, at roughly
                                         100 bytes per line, so a 100-node cluster returns about 10 KiB
  WriteBufferEachConn  512 KiB    8 KiB  the largest request is CLUSTER MIGRATESLOTS, at roughly
                                         90 bytes per slot range
  RingScaleEachConn       2^10      2^4  the widest batch is the 5-command DoMulti in getNodeState,
                                         and ForceSingleClient means no other user shares the connection
```

Exceeding a buffer costs an extra flush rather than failing, and a full ring applies back-pressure rather than erroring, so these are floors on syscall count and not correctness limits.

### Testing

Running a scale-up, same as in #407, it was possible to get 1 Valkey cluster with 81 shards instead of 22 before hitting OOM with default resource limits.

#### Single-cluster scale-up, 128Mi limit unchanged:

| | before | after |
|---|---|---|
| total allocation | 30.6 GB | 5.4 GB |
| `valkey-go.NewClient` share | 85% | 13% |
| RSS at 38 pods | 135 MB | 66 MB |
| peak GC target | 133 MB | 113 MB (at 160 pods) |
| OOMKilled at | 22 shards, 44 pods | 81 shards, 162 pods |

### Checklist


bjosv:
@rueian is this tuning correct? Maybe there is other/better options?

sandeepkunusoth (dismissed):

rueian:
Hi @bjosv,

LGTM. A few suggestions and corrections:

* 4K for both ReadBufferEachConn and WriteBufferEachConn might be enough if supporting a higher number of connections is more important than per connection perfomance in your case.
* per Do/DoMulti only uses 1 slot of the RingScaleEachConn. RingScaleEachConn is for pipelining concurrent operations to one connection. If you don't have concurrent operations to one connection, you can give it a lower value.
* ForceSingleClient doesn't mean no other user shares the connection, or at least not directly. It just means the client is using the SingleClient implementation.
* In the control-plane use, you may also want to set PipelineMultiplex to -1, to ensure each SingleClient only uses 1 connection. The default value of PipelineMultiplex for a SingleClient is 2, which means 4 connections.

bjosv:
Thanks @rueian much appreciated!
I added your suggested PipelineMultiplex and updated the comments.

We might need to lower the buffersizes in the future after we have decided what resource limits we aim for, but was great to know that we can. Thanks again.

sandeepkunusoth (approved):

PR #439, opened 2026-09-14
fix: let a standalone ValkeyNode start

### LINKED ISSUE #438: [BUG]: a standalone ValkeyNode never starts, it asks for a Secret that cannot exist
### Pre-flight checks

### Bug Description

A `ValkeyNode` created on its own, without a `ValkeyCluster`, never starts. The pod stays in
`CreateContainerConfigError` forever because the container asks for a Secret that only the cluster
controller creates.

`buildContainersDef` always wires `PRIMARY_AUTH` to `getSystemPasswordSecretName(node.Labels[LabelCluster])`.
A standalone node has no cluster label, so that resolves to `internal--system-passwords` (empty name,
two dashes), which does not exist.

The rest of the code already expects standalone nodes and provides for them in three places:

- `operatorUserPasswordSecret` returns nil on an empty cluster name, with the comment *"this can be empty
  only when valkeynode is created independently without ValkeyCluster CR"*
- the exporter sidecar is only added on an explicit `enabled: true`, *"so a standalone node must not gain
  one by default"*
- `ensureConfigMap` builds the node its own ConfigMap when `ServerConfigMapName` is empty, rendered by
  `generateValkeyNodeConfig`, which leaves `cluster-enabled` off

`PRIMARY_AUTH` is the one site that was missed. Guarding it the same way is enough to make the node
start: it comes up `1/1 Running` with `cluster_enabled:0` and the CR reports
`Ready=True/PodRunning ACLApplied=True/Applied`.

### Steps to Reproduce

```yaml
kind: ValkeyNode
metadata:
  name: mynode
spec:
  image: valkey/valkey:9.0.0
```

```
kubectl apply -f node.yaml
kubectl get pods
kubectl describe pod valkey-mynode-0
```

### Expected Behaviour

The pod starts and serves a single non-cluster Valkey.

### Actual Behaviour

The pod never starts:

```
Warning  Failed  kubelet  Error: secret "internal--system-passwords" not found
```

and stays in `CreateContainerConfigError` indefinitely.

See the manifest above. No `ValkeyCluster` exists in the namespace.

### Operator logs

Nothing. The operator creates the StatefulSet successfully; the failure is at the kubelet.

### Valkey pod logs

None, the container never starts.

### Operator version

Reproduced with the manager built from `main` (`3ec092d`). Also present in v0.5.0.

### Kubernetes version

v1.35.1

Other

### Additional context

Found while checking whether a single non-cluster node was possible today, ahead of the standalone
`Valkey` CRD in #388 / #390. Whatever controller ends up backing that CRD will likely create
`ValkeyNode`s with no cluster label, so this is on that path too.

Happy to send the fix (guard the flags and the env the same way `operatorUserPasswordSecret` already
guards its Secret).

Alfex4936:
This PR closes #438

### Summary

A `ValkeyNode` created without a `ValkeyCluster` never starts. Its container asks for
`internal--system-passwords`, a Secret that cannot exist, and the pod sits in
`CreateContainerConfigError`.

I ran into this while checking whether a single non-cluster node was possible today. I applied a
bare `ValkeyNode`, the StatefulSet came up fine, and then the pod just never went Ready:

```
Error: secret "internal--system-passwords" not found
```

The double dash is the giveaway. `buildContainersDef` wires `PRIMARY_AUTH` to
`getSystemPasswordSecretName(node.Labels[LabelCluster])`, and with no cluster label that name is built
from an empty string.

What made me think this was a bug and not intended is that the code around it already handles this
case. `operatorUserPasswordSecret` returns nil when the cluster name is empty, with the comment
"this can be empty only when valkeynode is created independently without ValkeyCluster CR". The
exporter sidecar is gated on an explicit `enabled: true` for the same reason, "so a standalone node
must not gain one by default". And `ensureConfigMap` builds the node its own ConfigMap when
`ServerConfigMapName` is empty, rendered by `generateValkeyNodeConfig`, which deliberately leaves
`cluster-enabled` off. So a cluster-less node is provided for in three places. `PRIMARY_AUTH` looks
like the one site that was missed.

For a node with no `valkey.io/cluster` label, `--primaryuser`, `--primaryauth` and the `PRIMARY_AUTH`
env are no longer set. Such a node has no primary to replicate from, so the flags had nothing to do
anyway.

Cluster-owned nodes are unchanged.

### Implementation

Pulled the replication flags into `replicationArgsAndEnv`, a sibling of the existing
`valkeyAnnounceArgsAndEnv`, returning `(nil, nil)` when the cluster label is absent. That keeps
`buildContainersDef` readable, since the args and env were being appended inline.

### Limitations

This only makes a bare `ValkeyNode` start. It is still a single non-cluster server with no Service,
no replica and no failover, which is what the `Valkey` CRD in #388 / #390 is for. I expect whatever
controller backs that CRD will create `ValkeyNode`s with no cluster label, so it lands on this path
too.

### Testing

`make test` and `make lint` pass.

`TestBuildValkeyNodePodTemplateSpec` was asserting the broken shape: its fixture sets no labels, so it
was a standalone node expecting `--primaryauth`. Updated it to the standalone expectation, and added
`TestBuildValkeyNodePodTemplateSpec_ClusterNodeHasReplicationAuth` for the other branch, so both sides
of the guard are covered.

Also ran both sides for real, on Kubernetes 1.35, with the manager built from `main` (`3ec092d`) and
its cache restricted to one namespace with `--watch-namespace`.

Before, applying the bare `ValkeyNode` from #438:

```
NAME              READY   STATUS                       RESTARTS   AGE
valkey-mynode-0   0/1     CreateContainerConfigError   0          25s

Warning  Failed  pod/valkey-mynode-0  Error: secret "internal--system-passwords" not found
```

After, with this patch (the existing pod has to be deleted once, since the StatefulSet rollout is
blocked behind a pod that will never go Ready):

```
NAME              READY   STATUS    RESTARTS   AGE
valkey-mynode-0   1/1     Running   0          13s

$ kubectl get valkeynode mynode -o jsonpath='{range .status.conditions[*]}{.type}={.status}/{.reason} {end}'
Ready=True/PodRunning ACLApplied=True/Applied
```

The node reports `cluster_enabled:0`, and the operations cluster mode rejects all work:

| Operation | ValkeyCluster, any shard count | standalone ValkeyNode |
|---|---|---|
| `MGET` across slots | `CROSSSLOT` | `1 2` |
| `MSET` across slots | `CROSSSLOT` | `OK` |
| multi-key `DEL` | `CROSSSLOT` | `2` |
| `SUNION` across slots | `CROSSSLOT` | `a b` |
| `MULTI`/`EXEC` spanning keys | `CROSSSLOT` | `OK OK` |
| `SELECT 1` | `ERR DB index is out of range` | `OK` |

### Checklist


deepakpunjabi:
@Alfex4936 `ValkeyNode` is meant for internal use and should not ideally be used by end users. Would your use-case be served by upcoming`Valkey` CR instead?

Alfex4936:
Yeah, `Valkey` CR is what I want. #387's motivation section describes my case pretty much exactly (single keyspace, multi-key ops and transactions, `SELECT`). Not asking for `ValkeyNode` to be user facing.

One question though, since it decides whether this PR is useful to you at all: will a `ValkeyNode` created by a `Valkey` carry `valkey.io/cluster`?

#387 says data pods reuse `ValkeyNode` unchanged, and the topology sketch has `Valkey "auth-cache"` owning `ValkeyNode auth-cache-0..2` plus `Secret internal-auth-cache-system-passwords` with `_replication` in it. But `buildContainersDef` can only get that Secret's name from `node.Labels["valkey.io/cluster"]`. If the `Valkey` controller stamps that label then this is a no-op for you and I'll close it. If it doesn't, the name has to come from somewhere owner agnostic, and `replicationArgsAndEnv` is where that would go.

Also, separate from all that: `TestBuildValkeyNodePodTemplateSpec`'s fixture sets no labels, so it's a standalone node, and it currently asserts `--primaryauth` with a Secret name built from `""`. Probably worth fixing either way.

Happy to cut this down to just the test change, or close it, whichever you prefer.

jdheyburn:
Hey @Alfex4936 thanks for raising this. We don't support users managing a `ValkeyNode` themselves - its intention is that it is always operator managed. We frequently change the API CR of ValkeyNode and having to consider users for that would cause a lot of toil.

Going forward the strategy is to use `Valkey` CR which will be introduced by Deepak. It is unlikely that it would carry the `valkey.io/cluster` label, but we would make the necessary changes to `ValkeyNode` to remove that dependency. However I would still avoid using `ValkeyNode` directly and instead wait for `Valkey`.

Alfex4936:
Got it, that answers my question. Closing this one.

One thing for whoever does the `ValkeyNode` change later: `TestBuildValkeyNodePodTemplateSpec`'s fixture sets no labels, so it is already a standalone node, and it currently asserts `--primaryauth` with a Secret name built from `""`. So the test locks in exactly the behaviour you would be removing, and it'll fail the moment you drop the dependency. The diff here is one way to do it if it's any use.

Leaving #438 open with the repro, close it if you'd rather track it with the `Valkey` work.

Thanks both for the quick replies.

PR #442, opened 2026-09-17
fix: ask the peers, not the node, whether a cluster member is failing

### LINKED ISSUE #441: [BUG]: Node-failure checks in `GetSyncedReplicas` and `countReadyShards` never match
### Pre-flight checks

### Bug Description

Two sites check node health with:

```go
if slices.Contains(node.Flags, "fail") || slices.Contains(node.Flags, "pfail") {
```

- `GetSyncedReplicas` (`internal/valkey/clusterstate.go`)
- `countReadyShards` (`internal/controller/valkeycluster_controller.go`)

Neither clause ever excludes a node:

- `"pfail"` is not a token Valkey emits (it renders `PFAIL` as `fail?`).
- `"fail"` only appears on a peer's CLUSTER NODES entry, never on a node's own `myself` line, which is exactly what `node.Flags` holds. A running node does not flag itself failing.

The flag check is dead, so:

- `countReadyShards` can count a shard ready while a still-present node in the shard is marked failing by peers (an unreachable node usually drops out of `shard.Nodes` and is caught by the node-count check instead).
- `GetSyncedReplicas` relies on the following `master_link_status != "up"` check, which covers most but not all cases.

It was noticed while reviewing the #433 parsing refactor, which preserved the existing `node.Flags` semantics.

## Possible fix

- `countReadyShards`: already has `state`; use `state.IsNodeFailed(node.Id)`.
- `GetSyncedReplicas`: pass `*ClusterState` in (its only caller, `findFailoverShard`, has it), switch to `IsNodeFailed`, fix the doc comment, and add a test for a replica a peer reports as `fail?`.
- Update (or add) tests

### Steps to Reproduce

-

### Expected Behaviour

-

### Actual Behaviour

-

```yaml

```

### Operator logs

```shell

```

### Valkey pod logs

```shell

```

### Operator version

main

### Kubernetes version

-

kind

### Additional context

_No response_

melancholictheory:
This PR closes #441.

### Summary

`GetSyncedReplicas` and `countReadyShards` both guard on

```go
if slices.Contains(node.Flags, "fail") || slices.Contains(node.Flags, "pfail") {
```

and neither half can ever be true. `pfail` is not a token Valkey emits; `CLUSTER_NODE_PFAIL` renders as `fail?`. And `node.Flags` is the node's own `myself` line, which never carries a failure flag, because a node does not mark itself failing (`clusterCron` skips `myself` when setting PFAIL). A failure flag on a node is set by its peers, not by the node.

Both sites now ask the peers through `ClusterState.IsNodeFailed`, which already existed and already matches `fail` and `fail?`.

`countReadyShards` no longer counts a shard ready while a node in it, still reachable by the operator, is reported `fail?` or `fail` by its peers. Before, that shard passed, and if the flagged node was the primary nothing downstream stopped it either, since `IsReplicationInSync` returns true for any primary.

`GetSyncedReplicas` no longer offers such a replica as a failover target. In practice the neighbouring `master_link_status != "up"` check caught most of these already, so the visible change there is small.

An unreachable node is unaffected: it never enters `shard.Nodes`, and the node-count check handles it as before.

### Implementation

`GetSyncedReplicas` takes `*ClusterState`. `findFailoverShard`, its only production caller, already holds it. `countReadyShards` already had `state` in scope.

### Testing

`make test` and `pre-commit run --all-files`.

The old check looked covered. `TestShardState_GetSyncedReplicas` built replicas with `Flags: {"slave", "fail"}` and `{"slave", "pfail"}`, states that cannot occur, and asserted they were excluded. The test handed the check the exact strings it looked for. That fixture is replaced by a replica whose own scrape is clean and whose primary reports it as `fail?`, the same shape `TestClusterState_IsNodeFailed` uses.

`countReadyShards` had no test; `TestCountReadyShards` covers a healthy shard and a shard with a present node the primary reports `fail?`. `TestFindFailoverShard` gains a case where the only replica is peer-reported failing.

All three new cases fail with the old checks reinstated behind the new signatures and pass with the fix.

### Checklist

Signed-off-by: melancholictheory <selimvhorst@gmail.com>


internal/valkey/clusterstate_test.go
```diff
@@ -184,27 +184,29 @@ func TestShardState_GetSyncedReplicas(t *testing.T) {
...
-		Info:    map[string]string{"role": "slave", "master_link_status": "up"},
-	}
-	pfailReplica := &NodeState{
-		Id:      "replica-4-id",
-		Address: "10.0.0.5",
-		Flags:   []string{"slave", "pfail"},
+		Flags:   []string{"slave"},
 		Info:    map[string]string{"role": "slave", "master_link_status": "up"},
 	}
+	primary.ClusterNodes = "primary-id 10.0.0.1:6379@16379 myself,master - 0 0 1 connected 0-5461\n" +
```
bjosv: #433 is merged now, and we should be able to use `primary.nodes = ParseClusterNodes(..` instead

internal/controller/valkeycluster_controller.go
jdheyburn, line 1550: Would any failing node cause readyCount to equal 0, such as a pod that were cut off momentarily and flagged every peer with `fail?` in its table?

Would it be better to determine node failure by looking at what other nodes think of it, and only if a majority of peer nodes agree it is failed? Something like below:

```go
// IsNodeFailedByMajority reports whether more than half of the other live
// nodes that know nodeId mark it "fail" or "fail?". A lone "fail?" is one
// node's opinion, and a node cut off from the bus marks every peer that way,
// so a single report is not enough here. IsNodeFailed keeps the any-viewer
// rule for the takeover path.
func (s *ClusterState) IsNodeFailedByMajority(nodeId string) bool {
      var viewers, failing int
      for _, shard := range s.Shards {
              for _, node := range shard.Nodes {
                      if node.Id == nodeId {
                              continue // its own entry never carries a failure flag
                      }
                      for _, entry := range node.nodes {
                              if entry.Id != nodeId {
                                      continue
                              }
                              viewers++
                              if entry.IsFailing() {
                                      failing++
                              }
                              break
                      }
              }
      }
      return viewers > 0 && failing*2 > viewers
}
```

melancholictheory: Agreed, and the second call site is where it costs more than a wrong number. One partitioned node marks every peer `fail?` in its own table, so under the any-viewer rule `GetSyncedReplicas` came back empty for every shard, `findFailoverShard` returned nil, and the roll went ahead on every primary without a proactive failover. `readyShards` at 0 was the visible half.

Took `IsNodeFailedByMajority` as you wrote it, in 35413c4, at both sites. `IsNodeFailed` stays any-viewer on the takeover path. Two partition tests, one per site, with three nodes where one node's table flags everyone: the majority rule keeps three ready shards and both failover targets, and both tests fail with the any-viewer rule put back. A node the other two agree on still drops out. Two-node clusters mark each other failed under a partition, which is true, and a node nobody knows yet has no viewers and is not failed, same as before.

internal/valkey/clusterstate_test.go
melancholictheory: Done in 86356ca, along with the two controller-package fixtures, which go through the new `SetClusterNodes` since they cannot reach the unexported field.

internal/valkey/clusterstate.go
```diff
@@ -356,6 +367,34 @@ func (s *ClusterState) HasFailoverQuorum() bool {
...
+		for _, node := range shard.Nodes {
+			if node.Id == nodeId {
+				continue // its own entry never carries a failure flag
+			}
+			for _, entry := range node.nodes {
+				if entry.Id != nodeId {
+					continue
+				}
+				viewers++
+				if entry.IsFailing() {
```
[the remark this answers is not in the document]
melancholictheory: Right, and it is Valkey's own distinction. `fail` is set only after a majority of primaries reported the node unreachable (`markNodeAsFailingIfNeeded`, quorum `size/2 + 1`) and is then broadcast with `clusterSendFail`, so one table carrying it already stands for a cluster-wide decision. `fail?` is a single node's opinion, and the one a cut-off node can put on everyone by itself.

f8f9231 makes a confirmed `fail` authoritative from any viewer and keeps the majority rule for `fail?` only. Added the one-confirmed-fail-of-three case to `TestClusterState_IsNodeFailedByMajority`; it fails without the early return.

internal/controller/valkeycluster_controller.go
melancholictheory: One refinement on top of your rule, in f8f9231: a confirmed `fail` counts from a single viewer, since Valkey only sets it after a primary majority and broadcasts it. `fail?` still needs the peer majority, which is the flag the partition case produces.

internal/valkey/clusterstate.go
```diff
@@ -356,6 +367,41 @@ func (s *ClusterState) HasFailoverQuorum() bool {
 // IsNodeFailed returns true if any live node reports the given node ID as
 // "fail" or "fail?" in CLUSTER NODES. Includes "fail?" (pfail) because when
 // majority of primaries are down, pfail can never promote to fail.
```
jdheyburn: Can you move this to above the function it's describing please?

internal/valkey/clusterstate.go
```diff
@@ -356,6 +367,41 @@ func (s *ClusterState) HasFailoverQuorum() bool {
...
+			for _, entry := range node.nodes {
+				if entry.Id != nodeId {
+					continue
+				}
+				if entry.HasFlag("fail") {
+					return true
+				}
+				viewers++
+				if entry.HasFlag("fail?") {
+					suspecting++
```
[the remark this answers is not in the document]
melancholictheory: Right, and it is the same rule Valkey applies. `clusterProcessGossipSection` records a failure report only when `clusterNodeIsVotingPrimary(sender)`, a primary that owns slots, so a replica's `fail?` never enters the vote that promotes it to `fail`. Counting replicas as viewers put them into a vote they do not take part in.

462faeb takes the majority over voting primaries only, mirrored as `isVotingPrimary` (primary with slots, from the node's own `myself` entry). A confirmed `fail` stays authoritative from any viewer, since it already carries that vote. Added the replica-only case to `TestClusterState_IsNodeFailedByMajority`: two replicas suspecting, one primary not, is false; it fails with the filter removed.

internal/valkey/clusterstate.go
melancholictheory: Moved in 462faeb. Adding the new function had left the `IsNodeFailed` comment stranded above the wrong one.

internal/valkey/clusterstate.go
```diff
@@ -257,6 +261,13 @@ func (s *ShardState) GetSyncedReplicas() []*NodeState {
 	return replicas
 }
 
+// SetClusterNodes replaces the node's peer table with one parsed from raw
+// CLUSTER NODES output. The scrape fills the table for live nodes; tests in
+// other packages use this to build a node with a particular view of its peers.
+func (n *NodeState) SetClusterNodes(raw string) {
```
bjosv, line 267: I don't think we have a convention for naming test helper-functions, but there are some projects that uses `ForTesting` suffixes

internal/valkey/clusterstate.go
```diff
@@ -353,6 +364,53 @@ func (s *ClusterState) HasFailoverQuorum() bool {
...
+				}
+				viewers++
+				if entry.HasFlag("fail?") {
+					suspecting++
+				}
+				break
+			}
+		}
+	}
+	return viewers > 0 && suspecting*2 > viewers
```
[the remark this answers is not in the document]
melancholictheory: Agreed, and it is Valkey's rule once more. The quorum there is taken over `cluster->size`, every primary that owns slots, and `fail?` is promoted only on reports from at least `size/2 + 1` of them (`markNodeAsFailingIfNeeded`). A primary that has not learned the node yet is still part of `size`, it just has not reported. Leaving it out of the denominator was the defect you reproduced.

Checking that against the source showed the denominator short in a second way: `cluster->size` keeps primaries the scrape could not reach, and counting voting primaries in the state dropped those as well. So 347e078 takes `size` from `cluster_size` in CLUSTER INFO, the largest value any scraped node reports, which is what `HasFailoverQuorum` already did a few lines up. Both now share a `clusterSize` helper. `reports` still counts `fail?` from voting primaries that have an entry.

`TestClusterState_IsNodeFailedByMajority` gets two cases: one of three primaries knows and suspects the target while the other two have no entry (false), and a `cluster_size` of five with three primaries reached and two of them suspecting (false, Valkey needs three). Both fail with the previous denominator put back.

internal/valkey/clusterstate.go
melancholictheory: Renamed to `SetClusterNodesForTesting` in 347e078. The comment now says tests in other packages are its only intended caller.

bjosv (approved):
LGTM

jdheyburn (approved):
Thank you!

PR #450, opened 2026-09-22
feat: make cluster-node-timeout user configurable

### LINKED ISSUE #434: [FEATURE]: Make cluster-node-timeout user-configurable
### Pre-flight checks

### Problem / Use case

The operator hardcodes `cluster-node-timeout` to 2s in `getBaseConfig`. Base config is written last in `valkey.conf`, so the last value wins and users cannot override it via `spec.config`.

At the same time, Valkey server default is 15s. 2s is much more aggressive.
This can cause concerns when:
1. Network blips longer than 2s trigger failover.
2. Larger clusters and larger datasets are more failure-prone at short timeouts. Fork-driven CoW latency on a big dataset can (potentially) stall a primary past 2s and cascade.

The right value depends on network stability

### Proposed solution

Remove `cluster-node-timeout` from `getBaseConfig` and let it come from `spec.config`.
Operator applies default when unset.

### Scope of change

### API / behaviour impact

### Contribution

### Additional context

Open questions:

1. Should we raise default `cluster-node-timeout` value to 5s (still under valkey default 15s) for accommodating typical networking? If the default stays 2s, existing clusters see no change. If the default changes, existing clusters roll once on operator upgrade.
2. Should the CRD enforce a minimum? eg. values under 1s not allowed.
3. Should we make `cluster-node-timeout` live-settable so a change applies without rolling pods?

deepakpunjabi:
This PR closes #434.

### Summary

`cluster-node-timeout` was hardcoded to 2000 in `getBaseConfig`. User could not set the config. This drops the directive from the base config. Valkey's own default of 15s now applies when the user sets nothing, and `spec.config` owns the value.

- `cluster-node-timeout` is settable via `spec.config`.
- The operator no longer sets a default. Unset means Valkey's default of 15s.
- Clusters created on an older operator roll once on upgrade and move from 2s to 15s. Set the value explicitly in spec.config to keep the old behaviour. **Need to add this in release notes for new operator version.**

### Implementation

- `internal/controller/config.go`: removed the `cluster-node-timeout` entry from `getBaseConfig`.
- `internal/controller/config_rollhash_test.go`: removed the key from `legacyRollConfigRender`, the frozen roll-hash pin.
- I have not matched 80 chars per line in comments as there is no style guide in golang or valkey-operator around it. Let me know if this needs to be enforced so I can update the PR.
- No CRD field was added, so make manifests produces no diff.

### Limitations

- No validation on the value. `spec.config` is a `map[string]string`, so a minimum (for example rejecting under 1s) would mean a per-directive CEL rule on a free-form map. Valkey rejects genuinely invalid values at startup. The docs name a sane floor instead. Config-aware validation is already tracked as a future plan in docs/valkeycluster.md.
- Not live-settable. Changing it rolls the pods. `cluster-node-timeout` is settable via `CONFIG SET` in Valkey, so adding it to `liveConfigAllowlist` is feasible. Left out here to keep this PR to one change, per the suggestion on the issue.
- One-time roll on operator upgrade for existing clusters. Unavoidable if the hardcoded value goes away.

### Testing

- `make test` passes.
- Added a spec in `internal/controller/config_test.go`.

### Checklist


internal/controller/config_rollhash_test.go
```diff
@@ -38,6 +38,9 @@ import (
 // Do NOT update this copy to track production code. Change it only when a pod
 // roll on operator upgrade is intended and understood.
 //
+// cluster-node-timeout was dropped from the base config so users can set it via spec.config (#434).
+// Existing clusters roll once on upgrade and pick up the Valkey default of 15s.
+//
```
jdheyburn: I think we can omit this comment.

docs/valkeycluster.md
```diff
@@ -39,6 +39,19 @@ maxmemory         # There are no safeguards, ensure you do not exceed your conta
...
+`cluster-node-timeout` controls how long a node may be unreachable before the cluster marks it failed and starts a failover. The operator does not set it, so the Valkey default of 15 seconds applies. Set it in `spec.config` to tune failover for your network.
+
+```yaml
+config:
+  cluster-node-timeout: "5000"
+```
+
+Lower values fail over faster but make the cluster more sensitive to network blips and to latency spikes on a busy primary. Values under 1 second are rarely a good idea outside a low-latency single-zone network. There is no validation on it by operator as it would require the type to be config aware instead of plain string. Changing it rolls the pods.
+
+Operator versions before #434 have `cluster-node-timeout` hardcoded to 2000. Clusters created on an older operator roll once on upgrade and move to the Valkey default. Set the value explicitly in `spec.config` to keep the old behaviour.
```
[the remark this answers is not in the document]
deepakpunjabi: So we don't know in which operator version this will be merged. We can update the comment when target release is finalized.

jdheyburn (commented):
I think we should have a follow up issue for setting a `cluster-node-timeout: 2000` on the e2e tests where we need a fast failover .

deepakpunjabi:
> I think we should have a follow up issue for setting a cluster-node-timeout: 2000 on the e2e tests where we need a fast failover .

Ack, will do.

internal/controller/config_rollhash_test.go
deepakpunjabi: Done

jdheyburn (approved):

PR #452, opened 2026-09-22
fix: skip only the shard whose primary is unidentifiable

### LINKED ISSUE #451: [BUG]: One shard with an unidentifiable primary stalls the whole reconcile
### Pre-flight checks

### Bug Description

`reconcileValkeyNodes` means to skip a shard's roll when it cannot identify that
shard's primary, but does so with `return` from inside the loop over shards:

```go
for shardIndex := range int(cluster.Spec.Shards) {
    if clusterState != nil && shardExistsInTopology(...) &&
        primaryNodeIndexForShard(...) < 0 {
        return true, nil
    }
```

That aborts the remaining shards' rolls and the rest of `Reconcile`:
`promoteOrphanedReplicas`, `healStaleAddressPeers`, `forgetStaleNodes`,
MEET/ADDSLOTSRANGE/REPLICATE, `handleScaleIn`, the rebalancer.

Those phases are what repair the offending shard, so its own remedy sits behind
the deferral its condition triggers: it defers because the primary is
unidentifiable, and the primary stays unidentifiable because it deferred. Repeats
every 2s forever at `Ready=False UpdatingNodes`, while `status.readyShards` still
reports full health.

The cluster does not recover on its own.

### Steps to Reproduce

Chaos defaults, `Persistence: false`. Fails on iteration 1 in about a third of runs, ~8 minutes:

```bash
CHAOS_SCENARIOS=delete-primary-pod CHAOS_SHARDS=3 make test-chaos
```

### Expected Behaviour

A skip should be scoped to the shard it concerns. One unidentifiable primary
should not stop other shards rolling, nor block the phases that resolve it.

### Actual Behaviour

The cluster does not recover on its own.

```yaml

```

### Operator logs

```shell

```

### Valkey pod logs

```shell

```

### Operator version

`main` from `118c675` (2026-09-14). `v0.6.0` is not affected.

### Kubernetes version

-

kind

### Additional context

_No response_

bjosv:
This PR closes #451.

### Summary

`reconcileValkeyNodes` meant to skip one shard's roll when it could not identify that shard's primary, but did so with `return` from inside the loop over shards. That aborted the remaining shards' rolls and the rest of `Reconcile`:
`promoteOrphanedReplicas`, `healStaleAddressPeers`, `forgetStaleNodes`, MEET/ADDSLOTSRANGE/REPLICATE, `handleScaleIn`, the rebalancer.

Those phases are what repair the offending shard, so its own remedy sat behind the deferral its condition triggered: it deferred because the primary was unidentifiable, and the primary stayed unidentifiable because it deferred.
Repeating every 2s at `Ready=False UpdatingNodes`, with no recovery without manual intervention. In the worst observed case every shard lost its primary, leaving no hash slots served at all.

The skip is now scoped to the shard it concerns. Other shards keep rolling and the later phases run, so each reason a primary is unidentifiable reaches the mechanism that resolves it: the rebalancer, the stale-address handling, or `handleScaleIn`.

`Ready` is still withheld with a 2s requeue while any shard is skipped.

The issue is seen after #320, but the defect predate #320 and are byte-identical at `9564ac8`. The guard's `clusterState != nil` clause was acting as an undocumented "a roll is pending" gate, because state was only scraped in that case. #320 widened the scrape condition, deleted `anyNodeRequiresFailoverAwareRoll`, then hoisted the snapshot into `Reconcile` to fix a double-scrape raised in review. 

### Implementation

- The guard `continue`s instead of `return true, nil`.
- After the loop the function reports `errShardRollSkipped`, a sentinel rather than a failure. The caller matches it with `errors.Is`, continues through the remaining phases, and withholds `Ready` at the end.

### Testing

Chaos suite before: **24 of 24 runs failed**, median at iteration 12, minimum 1.
After: a run reached **iteration 166** over 2h05m with 153 iterations passing and no deadlock, ending on an unrelated harness problem (the ValkeyCluster CR was gone while the recovery poll was still running). `scale-replicas` passed at iterations 160 and 161, having previously been fatal. 30 data-integrity checks passed.

### Checklist


internal/controller/valkeycluster_controller.go
```diff
@@ -89,7 +89,9 @@ type ValkeyClusterReconciler struct {
 //     (upsertConfigMap).
 //   - Ensure one ValkeyNode per (shard, node) pair exists, creating missing
 //     nodes and propagating spec changes one at a time in shard order with
-//     replicas updated before the primary (reconcileValkeyNodes).
+//     replicas updated before the primary (reconcileValkeyNodes). A shard whose
+//     primary cannot be identified has its roll skipped; the remaining shards and
+//     the phases below still run.
```
jdheyburn: > the remaining shards and the phases below still run.

This is only true once no other shard is mid-roll. Perhaps this wording is better:

> A shard whose primary cannot be identified has its roll skipped; other shards still roll, and once none is mid-roll the phases below run to repair it.

jdheyburn (commented):
Are you also able to enhance this test case that CodeRabbit suggested?

https://github.com/valkey-io/valkey-operator/pull/452#pullrequestreview-5290760631

PR #453, opened 2026-09-23
test(e2e): cover replica scale-down with a relocated primary

bjosv:
This PR reproduce #451 and is a test follow-up to the correction in #452

### Summary

Adds an e2e spec for scaling replicas down while a shard's primary sits above the new node-index bound. Completing that scale-down means removing the node currently serving the shard, so the spec also asserts the seeded keys survive.

Depends on #452. Without it the spec fails after 120s with `Reconciling`/`UpdatingNodes` and three ValkeyNodes where two are desired, which is the stall that PR fixes.

Test only.

### Implementation

The spec seeds 500 keys, fails over to node-index 2, then patches `spec.replicas` to 1 so that index falls outside the desired topology, and asserts the cluster returns to `Ready` with two nodes and all keys intact.

### Limitations

Legacy tests can use the added test utils. Not done in this PR. 

### Testing

```
TEST_LABELS="replica-scale-down" make test-e2e
```
- Without #452: fails at the "return to Ready with two nodes" assertion after 120s, reporting `cluster state: Reconciling, reason: UpdatingNodes`.
- With #452: passes.

### Checklist


jdheyburn:
This covers the second repo from #451, could it also cover the first one (quorum loss via 3 shards, 1 replica, no persistence)?

Before killing the primaries of shards 0 and 1:
- set `cluster-replica-no-failover yes` on their replicas
- set `shutdown-on-sigterm "nosave now"` on the primaries

Without that, the SIGTERM hand-off promotes the replicas and the spec passes on main.

Then you can force-delete both primaries and assert `Ready` plus `cluster_slots_ok:16384`. On my kind cluster this stalls main (`cluster_slots_ok:5462`, no TAKEOVER) and recovers with #452 (TAKEOVER, then Ready ~40s after the swap).

